PR #48, opened 2018-07-31
Change the behavior of config options

raoyitao:
* There is a parent-child relation in the configuration options of
  entities. When an option is not present in a child, then look up
  in the parent. Some options have default values. If a user doesn't
  specify an option in the entity constructor then the default value
  will be used. Currently if an option is not explicitly specified by
  user then the default value defined in parent will be used. We want
  to change this behavior that its default value takes precedence, but
  provide a parameter which sometimes can be used to makes clear that
  default value in parent is prefered.

### DIFF
--- a/testplan/common/config/base.py
+++ b/testplan/common/config/base.py
@@ -23,23 +23,32 @@ class DefaultValueWrapper(object):
     Utility class for distinguishing if a value is passed as schema default.
     """
 
-    def __init__(self, value):
+    def __init__(self, value, low_precedence=None):
         self.value = value
+        self.low_precedence = low_precedence
 
     def __repr__(self):
-        return '{}(value={})'.format(self.__class__.__name__, repr(self.value))
+        return '{}(value={}{})'.format(
+            self.__class__.__name__, repr(self.value),
+            ', prefer parent\'s default option' \
+                if self.low_precedence else ''
+        )
 
 
-def ConfigOption(key, default=ABSENT):
+def ConfigOption(key, default=ABSENT, low_precedence=None):
     """
     Wrapper around Optional, subclassing is not an option
     as Schema library does internal type checks as `type(obj) is Optional`.
+
+    With `low_precedence` set to True, the default value defined in parent
+    class has higher priority to be retrieved, unless explicitly set the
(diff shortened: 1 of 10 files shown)


testplan/common/config/base.py
```diff
@@ -74,12 +83,20 @@ def get_key_str(option):
...
+        if raw_key not in target_key_mapping:
+            target[source_key] = source[source_key]
+        else:
+            target_key = target_key_mapping[raw_key]
+            if isinstance(source_key.default, DefaultValueWrapper) and \
+                    source_key.default.low_precedence is not None and \
+                    isinstance(target_key.default, DefaultValueWrapper) and \
+                    target_key.default.low_precedence is None:
+                target_key.default.low_precedence = \
+                        source_key.default.low_precedence
```
raoyitao: I think it's better write like this to be safe:
    value = target.pop(target_key)
    target_key.default.low_precedence = source_key.default.low_precedence 
    target[target_key] = value

testplan/common/config/base.py
```diff
@@ -23,23 +23,32 @@ class DefaultValueWrapper(object):
...
-        return '{}(value={})'.format(self.__class__.__name__, repr(self.value))
+        return '{}(value={}{})'.format(
+            self.__class__.__name__, repr(self.value),
+            ', prefer parent\'s default option' \
+                if self.low_precedence else ''
+        )
 
 
-def ConfigOption(key, default=ABSENT):
+def ConfigOption(key, default=ABSENT, low_precedence=None):
```
Pyifan: Would it be better for low_precedence to have a default value as True? Then we toggle the behavior with low_precedence = True/False, rather than None/False which is not very conventional?

testplan/common/config/base.py
Pyifan: This piece of logic deserves some comment to explain its behavior: if inherited class re-define an option, and its low_precedence is default to None, this code will fetch the low_precedence flag from the option of the same name in parent class. (personally i think this is too much automation... but you have a point that when user set low_presedence=True he will want this to propagate down, let's see what John thinks)

PR #53, opened 2018-09-03
Add diff assertion for comparing textual content

raoyitao:
* Line diff assertion. Checks if 2 blocks of textual content have
differences. If difference found, generates a list of strings showing
the delta. You can set comparison option like "ignore_space_change",
"ignore_whitespaces" or "ignore_blank_lines", you can also specify the
output of delta in unified or context mode.

### DIFF
--- a/doc/en/assertions.rst
+++ b/doc/en/assertions.rst
@@ -504,6 +504,83 @@ of exception has been raised without matching the given ``pattern`` or ``func``.
           ...
 
 
+:py:meth:`result.diff <testplan.testing.multitest.result.Result.diff>`
+----------------------------------------------------------------------------
+
+Line diff assertion. Checks if textual content ``first`` and ``second`` have difference with given options.
+If difference found, generates a list of strings showing the delta.
+
+    .. code-block:: python
+
+      @testcase
+      def sample_testcase(self, env, result):
+          first, second = '', ''
+          with open('1.txt', 'r') as f1:
+              first = f1.read()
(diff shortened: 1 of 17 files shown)


doc/en/assertions.rst
```diff
@@ -504,6 +504,84 @@ of exception has been raised without matching the given ``pattern`` or ``func``.
...
+----------------------------------------------------------------------------
+
+Line diff assertion. Checks if textual content ``first`` and ``second`` have difference with given options.
+If difference found, generates a list of strings showing the delta.
+
+    .. code-block:: python
+
+      @testcase
+      def sample_testcase(self, env, result):
+          f1 = open('1.txt', 'r')
```
Pyifan: Would it be a bit more fancy looking if we use context mgr here?

raoyitao:
Travis CI failed.
It reported that could not fetch python 3.4:
--------------------------------------------------------------------------------
406 Setting APT mirror in /etc/apt/sources.list: http://archive.ubuntu.com/ubuntu/
407 3.4 is not installed; attempting download
408 Downloading archive: https://s3.amazonaws.com/travis-python-archives/binaries/ubuntu/14.04/x86_64/python-3.4.tar.bz2
409 $ curl -sSf -o python-3.4.tar.bz2 ${archive_url}
410 curl: (56) SSL read: error:00000000:lib(0):func(0):reason(0), errno 104
411 Unable to download 3.4 archive. The archive may not exist. Please consider a different version.

PR #63, opened 2018-10-22
Always specify testplan package for remote host

ryan-collingham:
Plus some other tweaks and refactoring of how local and remote workspaces, working directories and dependencies are managed for the RemotePool. Doesn't require testplan to be within the current workspace - the workspace should correspond to what we are testing.


ryan-collingham:
Splitting this large PR into smaller ones, so we can work out what is breaking Travis.

PR #66, opened 2018-10-24
Set the worker heartbeat for the child loop workers

ryan-collingham:
For workers spawned by child processes, set the worker heartbeat to a non-zero value to enable checking of worker health. That way the testcase will quickly fail if the child thread dies instead of hanging.


ryan-collingham:
worker_heartbeat should only be set on process pools and not thread pools, as the thread workers do not send heartbeats.

PR #90, opened 2018-12-13
Fetch error log of failed driver

raoyitao:
* When error occurs during environment startup or shutdown, user may
  want to know what happened about the Multitest drivers. Currently in
  our report only stack trace information is retained, so it will be
  helpful for user if we can fetch error log from the failed drivers
  and attach it to report. Especially this feature provides convenience
  when the test is executed in remote.

### DIFF
--- a/test/functional/testplan/testing/multitest/test_multitest_drivers.py
+++ b/test/functional/testplan/testing/multitest/test_multitest_drivers.py
@@ -1,14 +1,16 @@
 """TODO."""
 
 import os
+import re
 
 from testplan.testing.multitest import MultiTest, testsuite, testcase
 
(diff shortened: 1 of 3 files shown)


testplan/common/entity/base.py
```diff
@@ -148,10 +148,10 @@ def start(self):
 
         # Wait resources status to be STARTED.
         for resource in self._resources.values():
-            if resource.cfg.async_start is False:
-                continue
             if resource in self.start_exceptions:
                 break
+            if resource.cfg.async_start is False:
+                continue
             else:
```
Pyifan: I don't quite get the rational of this change.

PR #92, opened 2018-12-19
Driver support for unit test

raoyitao:
* Only MultiTest has drivers which can be started and stopped along
  with test execution. We want to enable this feature for unittest.
  Just move related code to the parent class of MultiTest, and then
  ProcessRunnerTest can utilize the drivers, thus the derived class
  of ProcessRunnerTest, such as GTest, is able to aquire information
  of drivers from environment variables of unittest process.


ryan-collingham:
This commit is already in https://github.com/Morgan-Stanley/testplan/pull/95/ ?

PR #98, opened 2019-01-07
Add timeout support for testplan

raoyitao:
* Some tasks may spend too much time to execute or have a dead loop,
  testplan provides a timeout option to end the execution.

* Make sure that after timeout, user can get a pdf report showing
  the results of finished tasks, while unfinished tasked can still be
  listed with an error message.

* If timeout, the status of testplan report is ERROR with an error
  message displayed in pdf report, even all finished tasks are passed.

* This feature should be work for all tasks that executed in local
  runner or pools.

### DIFF
--- a/testplan/runnable.py
+++ b/testplan/runnable.py
@@ -8,7 +8,7 @@ import webbrowser
 
 from collections import OrderedDict
 
-from schema import Schema, Or, Use
+from schema import Schema, Or, And, Use
 
 from testplan import defaults
 from testplan.common.config import ConfigOption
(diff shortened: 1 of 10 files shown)


testplan/runnable.py
```diff
@@ -401,6 +405,16 @@ def _wait_ongoing(self):
                 resource.abort()
 
         while self.active:
+            if self.cfg.timeout and \
+                    time.time() - self._start_time > self.cfg.timeout:
+                self.result.test_report.logger.error(
+                    'Timeout: Aborting execution after {} seconds'.format(
+                        self.cfg.timeout))
+                for dep in self.abort_dependencies():
```
JohnChiotis: Why dont we simply call self.abort() method that is inherited by entity base class?

raoyitao: we cannot abort TestRunner, or it won't create test result & generate pdf report. It only aborts executors and the reports of finished multitest can still be fetched.

PR #99, opened 2019-01-09
Add resource monitor for testplan

yuxuan-ms:
Add resource monitor that will export a JSON file includes:

* test and Testplan internals (events) start and stop times (seconds)
* resource utilisation per host

To run Testplan with Resource Monitor just run your Testplan script with the --resource-monitor command line argument.

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -458,6 +458,8 @@ class Entity(object):
         Creates runpath related directories.
         """
         self._runpath = self.generate_runpath()
+        # Update self.cfg.runpath if use the default runpath
+        self.cfg.runpath = self._runpath
         self._scratch = os.path.join(self._runpath, 'scratch')
         if self.runpath is None:
             raise RuntimeError('{} runpath cannot be None'.format(
@@ -561,6 +563,7 @@ class Runnable(Entity):
         self._environment = self.__class__.ENVIRONMENT(parent=self)
         self._result = self.__class__.RESULT()
         self._steps = deque()
+        self.resource_monitor = None
 
     @property
     def result(self):
--- a/testplan/common/utils/monitor/__init__.py
+++ b/testplan/common/utils/monitor/__init__.py
@@ -1,0 +1,269 @@
+"""Resource monitor utilities."""
+
+import os
+import uuid
+import time
+import json
+import socket
+import logging
+import threading
+from testplan.common.utils.path import makedirs
+from .collectors import get_collector
+
+
+class EventMonitor(object):
+    """
+    Event monitor will record event start and stop times(seconds)
+    """
+    def __init__(self, parent=None):
+        self.logger = logging.getLogger('Monitor')
+        self.parent = parent
+        self.hostname = socket.getfqdn()
+        self.events_metadata = {}
+        self.events_data = {}
+        self.lock = threading.Lock()
+        self.log_directory = None
+        self.uid = str(uuid.uuid4())
+
+    def attach(self, event_monitor):
+        """
+        Merge the record event in two EventMonitor
+
+        :param event_monitor:  The EventMonitor will be merged.
+        :type event_monitor: ``EventMonitor``
+        :return:
+        """
+        self.events_metadata.update(event_monitor.events_metadata)
+        self.events_data.update(event_monitor.events_data)
+
+    def setup_logger(self, log_path=None, log_level=None):
+        """
+        Sets up a unique logger for Monitor each time it is called. Should only be called once or logs will be directed
+        to new logger on subsequent call.
+
+        :param log_path: Absolute path to file to save logs to.
+        :type log_path: ``str``
+        :param log_level: Level of logging to save. Currently either ``logging.DEBUG`` if "debug" given or
+                          ``logging.INFO`` if anything else is given.
+        :type log_level: ``str``
+
+        :return: ``None``
+        :rtype: ``NoneType``
+        """
+        self.logger = logging.getLogger(str(uuid.uuid4()))
+        if log_level == 'debug':
+            log_level = logging.DEBUG
+        else:
+            log_level = logging.INFO
+        if log_path:
+            directory = log_path.rsplit(os.sep, 1)[0]
+            self.log_directory = directory
+            makedirs(directory)
+            file_handler = logging.FileHandler(filename=log_path)
+        else:
+            file_handler = logging.NullHandler()  # pylint: disable=bad-option-value
+        formatter = logging.Formatter('%(asctime)s-Monitor-%(levelname)-8s:  %(message)s')
+        file_handler.setFormatter(formatter)
+        file_handler.setLevel(log_level)
+        self.logger.setLevel(log_level)
+        self.logger.addHandler(file_handler)
+
+    def _record_event(self, event_uuid, event):
+        if event_uuid not in self.events_data:
+            self.events_data[event_uuid] = []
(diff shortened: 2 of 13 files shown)


testplan/common/utils/monitor/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+
+    def _record_event(self, event_uuid, event):
+        if event_uuid not in self.events_data:
+            self.events_data[event_uuid] = []
+        self.events_data[event_uuid].append({
+            'event': event,
+            'time': time.time(),
+        })
+
+    def started(self, event_type, metadata=None):
```
Pyifan: Consider rename started -> start_event? otherwise it is confusing when resource_monitor has a start() method that does totally different thing.

same for stopped

testplan/testing/multitest/base.py
```diff
@@ -333,6 +335,10 @@ def run_tests(self):
         ctx = [(self.test_context[idx][0], self.test_context[idx][1][:])
                for idx in range(len(self.test_context))]
 
+        if self.cfg.resource_monitor:
```
Pyifan: Can we handle other tests like GTest too?

Or if we want finer granularity e.g suite/case level, would current design be easy to adapt?
Maybe do we need to record parent-child relationship via json?

testplan/common/entity/base.py
```diff
@@ -458,6 +458,8 @@ def make_runpath_dirs(self):
         Creates runpath related directories.
         """
         self._runpath = self.generate_runpath()
+        # Update self.cfg.runpath if use the default runpath
+        self.cfg.runpath = self._runpath
```
JohnChiotis: Configuration input should not be altered

testplan/common/utils/monitor/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+from testplan.common.utils.path import makedirs
+from .collectors import get_collector
+
+
+class EventMonitor(object):
+    """
+    Event monitor will record event start and stop times(seconds)
+    """
+    def __init__(self, parent=None):
+        self.logger = logging.getLogger('Monitor')
```
JohnChiotis: I think in this implementation we may have to use the global TESTPLAN_LOGGER object that repo offers and just register a 'monitor' file handler?
from testplan.logger import TESTPLAN_LOGGER

testplan/common/utils/monitor/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+        :type log_path: ``str``
+        :param log_level: Level of logging to save. Currently either ``logging.DEBUG`` if "debug" given or
+                          ``logging.INFO`` if anything else is given.
+        :type log_level: ``str``
+
+        :return: ``None``
+        :rtype: ``NoneType``
+        """
+        self.logger = logging.getLogger(str(uuid.uuid4()))
+        if log_level == 'debug':
```
JohnChiotis: Will have to change based on the global logger.

testplan/common/utils/monitor/__init__.py
JohnChiotis: event_started()

testplan/common/utils/monitor/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+            elif isinstance(metadata, dict):
+                _metadata = metadata.copy()
+            else:
+                raise TypeError('Event metadata must be a dictionary - not {}'.format(type(metadata)))
+            _metadata['type'] = event_type
+            self.events_metadata[event_uuid] = _metadata
+            self._record_event(event_uuid, 'start')
+        return event_uuid
+
+    def stopped(self, event_uuid, metadata=None):
```
JohnChiotis: event_stopped()

testplan/common/utils/monitor/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+        Start the thread of collect the metrics.
+
+        :return: ``None``
+        :rtype: ``NoneType``
+        """
+        self._poll = threading.Thread(target=self._monitoring)
+        self._poll.daemon = True
+        self._poll_start = True
+        self._poll.start()
+        self.logger.info('Resource polling started')
```
JohnChiotis: dot . end of sentence

testplan/common/utils/monitor/collectors.py
```diff
@@ -0,0 +1,121 @@
...
+    def __init__(self, director):
+        super(LinuxCollector, self).__init__(director)
+        try:
+            cpu_info = subprocess.check_output(['lscpu'])
+            self.metadata['extra']['cpu_info'] = cpu_info.decode()
+        except Exception:
+            pass
+
+
+class LinuxContainerCollector(LinuxCollector):
```
JohnChiotis: Is this relevant to OSS world? Are the file paths standard? should they be configurable?

testplan/common/utils/monitor/collectors.py
```diff
@@ -0,0 +1,121 @@
...
+                except (IOError, AttributeError, ValueError):
+                    pass
+
+
+class WindowsCollector(Collector):
+    def __init__(self, director):
+        super(WindowsCollector, self).__init__(director)
+
+
+def get_collector(director=os.getcwd()):
```
JohnChiotis: we use an internal utility instead of os.getcwd()
https://github.com/Morgan-Stanley/testplan/blob/master/testplan/common/utils/path.py#L36

testplan/common/utils/monitor/collectors.py
```diff
@@ -0,0 +1,121 @@
...
+    if platform.system() == 'Windows':
+        return WindowsCollector(director)
+    elif platform.system() == 'Linux':
+        try:
+            with open('/proc/1/cgroup') as cgroup_file:
```
JohnChiotis: Same for this path

testplan/testing/multitest/base.py
JohnChiotis: Should we have the monitor logic in the Test() base class of all MultiTest, GTest etc?

testplan/common/utils/monitor/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+        """
+        self.logger = logging.getLogger(str(uuid.uuid4()))
+        if log_level == 'debug':
+            log_level = logging.DEBUG
+        else:
+            log_level = logging.INFO
+        if log_path:
+            directory = log_path.rsplit(os.sep, 1)[0]
+            self.log_directory = directory
+            makedirs(directory)
```
JohnChiotis: Check for race conditions

testplan/common/utils/monitor/__init__.py
```diff
@@ -0,0 +1,269 @@
...
+            file_handler = logging.FileHandler(filename=log_path)
+        else:
+            file_handler = logging.NullHandler()  # pylint: disable=bad-option-value
+        formatter = logging.Formatter('%(asctime)s-Monitor-%(levelname)-8s:  %(message)s')
+        file_handler.setFormatter(formatter)
+        file_handler.setLevel(log_level)
+        self.logger.setLevel(log_level)
```
JohnChiotis: if self logger is global, that needs to go

ryan-collingham:
This has been open for a long time - what's happening with the resource monitor?

yuxuan-ms:
> This has been open for a long time - what's happening with the resource monitor?

Let's close the pull request. After complete the report UI, I will create a new pull request.

PR #101, opened 2019-01-15
Fix Python 3 AttributeError for dict itervalues

apretori-tic:
Thanks for making Testplan open source!

We found a problem using Python3 in our testcase:
```
  File "testplan/testplan/common/utils/sockets/fix/server.py", line 371, in active_connections(<testplan.common.utils.sockets.fix.server.Server object at 0x7f1408fc5c18>)
    for detail in self._conndetails_by_fd.itervalues()
AttributeError: 'dict' object has no attribute 'itervalues'
```

This patch addresses the above error.
I am not sure if you would be happy with it within `active_connections` function itself or whether there is a common place where this kind of thing should be placed.

### DIFF
--- a/testplan/common/utils/sockets/fix/server.py
+++ b/testplan/common/utils/sockets/fix/server.py
@@ -367,8 +367,21 @@ class Server(object):
           target)
         :rtype: ``list`` of ``tuple`` of ``str`` and ``str``
         """
+
+        # from PEP-0469 (common subset of Python 2 and 3)
+        try:
+            dict.iteritems
+        except AttributeError:
+            # Python 3
+            def itervalues(d):
+                return iter(d.values())
+        else:
+            # Python 2
+            def itervalues(d):
+                return d.itervalues()
+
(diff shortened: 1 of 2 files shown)


testplan/common/utils/sockets/fix/server.py
```diff
@@ -367,8 +367,21 @@ def active_connections(self):
...
+            def itervalues(d):
+                return iter(d.values())
+        else:
+            # Python 2
+            def itervalues(d):
+                return d.itervalues()
+
         return [detail.name
-                for detail in self._conndetails_by_fd.itervalues()
+                for detail in itervalues(self._conndetails_by_fd)
```
JohnChiotis: I think we can skip the new code addition that defines an itervalues method and we can simple change itervalues() to values() that works in both versions.
Is that ok?

apretori-tic: Ah yes perfect! Thank you.

PR #107, opened 2019-01-24
Local web app + server.

lambchr:
* Created a simple local web application using Flask.
* Created placeholder HTML files for the Monitor & Testplan UIs.

### DIFF
--- a/test/unit/testplan/web_ui/test_web_app.py
+++ b/test/unit/testplan/web_ui/test_web_app.py
@@ -1,0 +1,202 @@
+import os
+import uuid
+import shutil
+import tempfile
+
+import pytest
+
+from testplan.web_ui.web_app import app as tp_web_app
+
+STATIC_REPORTS = {
+    'testing': {
+        'uid': str(uuid.uuid4()),
+        'contents': str(uuid.uuid4()),
+    },
+    'monitor': {
+        'uid': str(uuid.uuid4()),
+        'contents': str(uuid.uuid4()),
+    },
+}
+
+DATA_REPORTS = {
+    'testplan': {
+        'uid': str(uuid.uuid4()),
+        'contents': str(uuid.uuid4()),
+    },
+    'monitor': {
+        'uid': str(uuid.uuid4()),
+        'contents': str(uuid.uuid4()),
+    },
+}
+
+
+def _create_tmp_file(tmp_file, contents):
+    """
+    Create the given file, and it's base directory if necessary.
+    Fill the file with the given contents.
+
+    :param tmp_file: Name of the file to be created (including path).
+    :type tmp_file: ``str``
+    :param contents: Contents of the file to be created.
+    :type contents: ``str``
+    """
+    base_dir = os.path.dirname(tmp_file)
+    if not os.path.exists(base_dir):
+        os.makedirs(base_dir)
+
+    with open(tmp_file, 'w') as tmp_file:
+        tmp_file.write(str(contents))
+
+
+def _create_tmp_static_files(base_dir):
+    """
+    Create fake static files to be sent by the Testplan web app.
+        <STATIC_PATH>/testplan/index.html
+        <STATIC_PATH>/testplan/static.file
+        <STATIC_PATH>/monitor/index.html
+        <STATIC_PATH>/monitor/static.file
+
+    :param base_dir: The Testplan web app's STATIC_PATH directory.
+    :type base_dir: ``str``
+    """
+    for report_type, report in list(STATIC_REPORTS.items()):
+        index_file = os.path.join(base_dir, report_type, 'index.html')
+        _create_tmp_file(tmp_file=index_file, contents=report['contents'])
+        static_file = os.path.join(base_dir, report_type, 'static.file')
+        _create_tmp_file(tmp_file=static_file, contents=report['contents'])
+
+
+def _create_tmp_data_files(base_dir):
+    """
+    Create fake data files to be sent by the Testplan web app.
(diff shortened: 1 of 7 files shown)


test/unit/testplan/web_ui/test_web_app.py
```diff
@@ -0,0 +1,202 @@
...
+        shutil.rmtree(self.static_dir)
+
+    def test_testplan_index(self):
+        """Does /testplan/<uid> return the correct index.html page."""
+        # Use correct UID, expect 200 response.
+        response = self.client.get('/testplan/{}'.format(STATIC_REPORTS['testing']['uid']))
+        assert response.status_code == 200
+        assert str(STATIC_REPORTS['testing']['contents']) in str(response.data)
+
+        # Use incorrect UID, still expect 200 response.
```
ryan-collingham: Do we not want a 404 response for an incorrect UID?

test/unit/testplan/web_ui/test_web_app.py
```diff
@@ -0,0 +1,202 @@
...
+        assert str(STATIC_REPORTS['testing']['contents']) in str(response.data)
+
+    def test_testplan_static(self):
+        """Does /testplan/static/static.file respond with the static.file file."""
+        # Use correct UID, expect 200 response
+        response = self.client.get('/testplan/static/static.file'.format(STATIC_REPORTS['testing']['uid']))
+        assert response.status_code == 200
+        assert str(STATIC_REPORTS['testing']['contents']) in str(response.data)
+
+        # This endpoint should return the index.html page no matter what UID is sent
```
ryan-collingham: Comment doesn't match the code?

test/unit/testplan/web_ui/test_web_app.py
```diff
@@ -0,0 +1,202 @@
...
+        assert str(STATIC_REPORTS['monitor']['contents']) in str(response.data)
+
+    def test_monitor_static(self):
+        """Does /testplan/static/static.file respond with the static.file file."""
+        # Use correct UID, expect 200 response
+        response = self.client.get('/monitor/static/static.file'.format(STATIC_REPORTS['monitor']['uid']))
+        assert response.status_code == 200
+        assert str(STATIC_REPORTS['monitor']['contents']) in str(response.data)
+
+        # This endpoint should return the index.html page no matter what UID is sent
```
ryan-collingham: again, I don't think this comment matches the code

testplan/web_ui/web_app.py
```diff
@@ -0,0 +1,105 @@
...
+    """Web App command line arguments."""
+    parser = argparse.ArgumentParser()
+    parser.add_argument('--static-path', nargs='?', default=None, const=pwd())
+    parser.add_argument('--data-path', nargs='?', default=None, const=pwd())
+    return parser.parse_args()
+
+@_api.route('/testplan/<string:report_uid>')
+class Testplan(Resource):
+    def get(self, report_uid):
+        """Get a Testplan report (HTML) given it's uid."""
```
ryan-collingham: Do we want to validate that the UID exists here (and return 404 if not), or will the web app handle that?

testplan/web_ui/web_app.py
```diff
@@ -0,0 +1,105 @@
...
+            'reports',
+            'testplan_{uid}'.format(uid=report_uid)
+        )
+        return send_from_directory(directory=directory, filename='report.json')
+
+@_api.route('/testplan/<string:report_uid>/assertions/<string:assertions_uid>')
+class TestplanAssertions(Resource):
+    def get(self, report_uid, assertions_uid):
+        """Get an Assertion report (JSON) for a specific Testplan report given their uids."""
+        raise NotImplemented()  # pylint: disable=notimplemented-raised
```
ryan-collingham: It's not great to shadow the NotImplemented builtin, would slightly prefer `from werkzeug import exceptions` and use as `exceptions.NotImplemented`.

lambchr: Fair point, I'll change it

test/unit/testplan/web_ui/test_web_app.py
lambchr: Thanks I'll change it

test/unit/testplan/web_ui/test_web_app.py
lambchr: Thanks I'll change it

test/unit/testplan/web_ui/test_web_app.py
lambchr: Thanks I'll change it

test/unit/testplan/web_ui/test_web_app.py
lambchr: No, this endpoint is a little confusing. The UID isn't used by this endpoint, it is actually used by the index.html page returned by the endpoint. The index.html page gets the UID from the URL and makes the /testplan/<uid>/report request based on this. This is confusing, I would prefer it if the flask web app could inject this UID to the index.html page. However I couldn't get that to work and have the index.html page pass this UID to the react component when I tried. If you'd like to give it a go or have another idea to make this less confusing/more let me know :)

testplan/web_ui/web_app.py
lambchr: No as explained above this endpoint doesn't actually make use of the UID, therefore I don't think it should validate it. At most it would just inject the UID into the HTML page. The HTML page would then use this UID to make a /testplan/<uid>/report request which can respond with either the report.json or a 404 (validating the JSON). The HTML page then knows how to handle this and can show an error if a 404 is received.

PR #113, opened 2019-02-20
Add wait_for_connection() method to FixServer

apretori-tic:
Occasionally there is a race when using FixServer in a testcase when the first thing the testcase does is try to receive a message from the FixServer before the connection has been established:
```
msg = env.fixserver.receive()
```
which results in:
```
Exception: Cannot use default connection since more connections active
```
(This error is not entirely correct as there are no connections active yet (server checks if number of connections != 1) - but that does not really concern me).

This patch adds a `wait_for_connection()` method which fixes the race.
BTW. I was wondering if it would be desirable to do this in the FixServer's `receive()` method. I did not want to impact other users/uses of `receive()` so opted to expose this function instead.

### DIFF
--- a/testplan/common/utils/sockets/fix/server.py
+++ b/testplan/common/utils/sockets/fix/server.py
@@ -145,6 +145,7 @@ class Server(object):
         self._socket = None
         self._recv_thread = None
         self._lock = threading.Lock()
+        self._condvar = threading.Condition(self._lock)
         self._pobj = select.poll()
 
     @property
@@ -324,6 +325,7 @@ class Server(object):
         elif _has_logon_tag(msg):
             self._logon_connection(fdesc, conn_name)
             self._no_lock_send(msg, conn_name)
+            self._condvar.notify_all()
         else:
             raise Exception(
                 'Connection {} sent msg before logon'.format(conn_name))
@@ -383,6 +385,25 @@ class Server(object):
         :rtype: ``bool``
         """
         return conn_name in self._conndetails_by_name
+
+    def wait_for_connection(self, timeout = 60, conn_name = None):
+        """
+        Waits for the given connection to appear, or any connection if None.
+
+        :param timeout: Time in seconds to wait for connection.
+        :type timeout: float
+        :param conn_name: Connection name to waited for, or None for any connection.
+        :type conn_name: ``tuple`` of ``str`` and ``str``
+
+        :return: ``True`` if connection is active, ``False`` if timed out waiting for it.
+        :rtype: ``bool``
+        """
+        if conn_name:
+            f = lambda: self.is_connection_active(conn_name)
+        else:
+            f = lambda: self.active_connections()
+        with self._condvar:
+            return self._condvar.wait_for(f, timeout)
 
     def stop(self):
         """
--- a/testplan/common/utils/sockets/fix/server.py
+++ b/testplan/common/utils/sockets/fix/server.py
@@ -386,7 +386,7 @@ class Server(object):
         """
         return conn_name in self._conndetails_by_name
 
-    def wait_for_connection(self, timeout = 60, conn_name = None):
+    def accept_connection(self, timeout = 10, conn_name = None):
         """
         Waits for the given connection to appear, or any connection if None.
 
--- a/testplan/testing/multitest/driver/fix/server.py
+++ b/testplan/testing/multitest/driver/fix/server.py
@@ -113,6 +113,13 @@ class FixServer(Driver):
         return self._server.is_connection_active(conn_name)
     is_connection_active.__doc__ = Server.is_connection_active.__doc__
 
+    def wait_for_connection(self, timeout = 60, conn_name = None):
+        """
+        Docstring from Server.wait_for_connection
+        """
+        return self._server.wait_for_connection(conn_name)
+    wait_for_connection.__doc__ = Server.wait_for_connection.__doc__
+
     def send(self, msg, conn_name=(None, None)):
         """
         Docstring from Server.send
--- a/testplan/testing/multitest/driver/fix/server.py
+++ b/testplan/testing/multitest/driver/fix/server.py
@@ -113,12 +113,12 @@ class FixServer(Driver):
         return self._server.is_connection_active(conn_name)
     is_connection_active.__doc__ = Server.is_connection_active.__doc__
 
-    def wait_for_connection(self, timeout = 60, conn_name = None):
+    def accept_connection(self, timeout = 10, conn_name = None):
         """
-        Docstring from Server.wait_for_connection
+        Docstring from Server.accept_connection
         """
-        return self._server.wait_for_connection(conn_name)
-    wait_for_connection.__doc__ = Server.wait_for_connection.__doc__
+        return self._server.accept_connection(conn_name)
+    accept_connection.__doc__ = Server.accept_connection.__doc__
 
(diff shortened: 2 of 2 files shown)


ryan-collingham:
Hi,

Thanks for the contribution. Can you clarify the expected use of this method within a testcase? Maybe if the method should always be called before the first message is `receive()`d it would make sense to make it a private method and have the FixServer automatically wait for the connection to be established before receiving the first message.

I do think we should also fix the error message in the case of 0 connections being active, thanks for spotting that!

Regards,
Ryan

JohnChiotis:
Can we first agree that the expected setup with FixServer and FixClient is:
1. FixServer starts
2. FixClient starts and connects
3. server.accept_connection() is being called in a post_start() MultiTest step or in a setup() in a testsuite.
4. NOTE: if you cannot accept the connection you can sleep and retry.
And then you have your connection!

Example:
    https://testplan.readthedocs.io/en/latest/download/Transports.html#fix
..but I would definately put the accept connection in post_start() of MultiTest.
Check my_tests/mtest.py file of:
    https://testplan.readthedocs.io/en/latest/download/Interactive.html#basic

If you think this scenario does cover your race-condition case please provide more information.

Otherwise I don't think this PR is needed and I would advise before making a PR to raise an issue topic so we have a conversation as a very first thing.

apretori-tic:
Hi! I hope you are all well!

@ryan-collingham having it private and called from `receive()` would be ideal. I was hoping that is an option, I will look into doing that and will update this PR a bit later. In that case the error message would be correct (I think) as `receive()` would block until there is a connection.

@JohnChiotis First just to clarify do you prefer me raising an issue first before opening a PR as a starting point? I certainly can do that next time.

My test at the moment starts a FixServer and our binary application (eventually the test will start several FixServers, our binary application and some clients of our binary application). I have a driver for our binary application that in its `started_check()` method waits for the binary to log the connection details used for clients. At some later point the binary starts connecting to the FixServer(s) depending on configuration (schedule). 
In this particular test I am working on now I want to test the FIX logon sequence used by a specific gateway in our binary to a FixServer. I also plan to test disconnection and re-connection at various points of processing.
So when my test case starts I want to assert correctness of the message received by the FixServer, but sometimes the connection has not occurred yet leading to the failure I described in my first post.

Regarding your suggestions, `accept_connection()` is for `TcpServer` and not `FixServer` it seems.
Up to now I used a polling loop to wait for connections but that is a bit unsatisfactory, I am looking for a cleaner solution that does not impose arbitrary delays on test suites by sleeping, i.e. proceed as soon as the awaited event occurs.

many thanks for helping!
kind regards.

lambchr:
Hi,

I think it looks very useful so we will merge it. 

However I want to point out this shouldn't be called from the receive method. The drivers should have finished setting up, including ensuring all connections have been made, before the tests have started to run. You should call this method from a function given to Mutitest's after_start parameter. [See here for an example of using Multitest's after start parameter](https://testplan.readthedocs.io/en/latest/download/Interactive.html#my-tests-mtest-py). This will ensure that the connection (or multiple connections if you decide) are active before your tests start. Perhaps we should add this to our FIX downloadable example.

Also I'll just confirm a couple of points:
  * If you have an issue like this in future we would prefer that you raise an issue with us before opening a PR. We really appreciate PRs but this allows us to work with you on the best approach to take from the start.
 * Next time you submit a PR could you also submit a testcase to cover the new functionality. Our tests are under the [test directory](https://github.com/Morgan-Stanley/testplan/tree/master/test). We can add the testcase this time.

Thanks again for contributing!

apretori-tic:
Thank you @lambchr , I will reach out via an issue next time.
Regarding unit tests - I am very happy to add one.
I had difficulty running tests - do they only work with python 2?  (We use python 3.7.2).

lambchr:
We [currently test against python 2.7 & 3.4](https://travis-ci.org/Morgan-Stanley/testplan/builds/496563307). We also test on Linux and Windows, however these builds are done internally (so no link to share). 

We should change our builds to run the tests against 3.7 (and fix any compatibility issues), I'll have to add that to the do list. Thanks for raising the issue. Can you try running the tests on Python 3.4 for now?

ryan-collingham:
> I had difficulty running tests - do they only work with python 2? (We use python 3.7.2).

I can run the tests fine with Python 3.7.2. What problems are you seeing?

apretori-tic:
I have a lot of deprecation warnings for various things and 2 errors. Should I be concerned about the deprecation warnings like:
`DeprecationWarning: invalid escape sequence \`
`DeprecationWarning: inspect.getargspec() is deprecated since Python 3.0, use inspect.signature() or inspect.getfullargspec()`
`PytestWarning: cannot collect test class 'Testplan' because it has a __init__ constructor`
`PytestWarning: cannot collect test class 'TestReport' because it has a __init__ constructor`
Too many to paste, should I be concerned about these?

Looking at that now maybe my environment lacks some modules:
`ModuleNotFoundError: No module named 'testplan'` and `ModuleNotFoundError: No module named 'flask'`. I will create another conda environment just for running tests I think, can use python 2 for that maybe.

ryan-collingham:
The warnings are expected. We should fix the DeprecationWarnings at some point, the warnings about __init__ constructors are benign though.

ModuleNotFoundError indicates that you probably haven't set your environment up correctly. Did you follow the instrucitons on https://testplan.readthedocs.io/en/latest/getting_started.html - in particular you need to run:

```
pip install -r requirements.txt
python setup.py develop
```

We also now have a dockerised development environment if you would like to try that, you can run with `docker run -it ryancollingham/dev_env:testplan`.

apretori-tic:
I did set it up before using those instructions and started to use that environment for my own tests. (Mixing conda and pip tripped me up where pip installs packages into conda environment where pip is installed and not the currently active one. I also switched to fish shell recently.)
I created a new conda environment for unit testing only and now it works! Thank you.

At the moment I cannot find any tests for FixServer or FixClient. I guess it is going to be a challenge to satisfy their msgclass and codec requirements in unit tests which is probably why there are no tests for it yet.

apretori-tic:
I need this method or some other mechanism to use in the `after_start` specified method for `FixServer` as @lambchr said. All examples and documentation I can find only uses it for `TCPServer`.
I considered subclassing `FixServer` but that does not seem feasible for something as fundamental as this. `FixServer` delegates to `fix.server.Server` which will need to be subclassed as well.

ryan-collingham:
Currently we have no unit or functional test coverage of FixServer/Client as far as I can tell, though there is some coverage in the example under examples/Transports/FIX. I think it would make sense to update the example to explicitly wait for the connection to be established in a post_start function.

You are right that accept_connection is a method of the `TCPServer` and not the `FixServer`. So this PR is really adding an equivalent of `TCPServer.accept_connection` to the `FixServer`. Even though `wait_for_connection` is a better description of what the method is actually doing I think we should call the new method `accept_connection` for consistency.

I think that waiting for the `threading.Condition` is more elegant than the polling approach used in `TCPServer`, nice one :)

apretori-tic:
Thanks @ryan-collingham , I renamed it and changed the timeout default to 10 seconds too as that is what TCPServer use for a timeout.
I will try to add some unit tests, it is going to take me a while. Would adding pyfixmsg as a dependency for unit testing be an option? Or would it be better to mock it for `FixServer` testing?

JohnChiotis:
There should be no dependency to pydixmsg,

kelliott55:
Closing due to no activity.

PR #115, opened 2019-03-01
Warn and fail if no tests were run and add default ProcessRunnerTest suite

ryan-collingham:
Previously:

`./test_plan.py --pattern BADPATTERN`

Would run no tests and just print that the testplan has passed. This is not very helpful when you mistype a pattern.

This PR changes behaviour so that if no tests were run we make an explicit warning and do not report the testplan as passed. Marking the testplan as failed gives a non-zero exit code and skips report generation in this case.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
   - "3.4"
 
 install:
-  - pip install -r requirements.txt
+  - pip install -r requirements.txt --upgrade
   - python setup.py develop
 
 script:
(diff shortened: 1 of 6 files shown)


ryan-collingham:
I also added/modified some debug logs in the process of figuring out how the TestRunner is working, I've left some of these in in case they are useful.

ryan-collingham:
(private CI has passed)

ryan-collingham:
Added an extra commit to this PR:

For ProcessRunnerTest, added function to create and use a single default suite to contain all testcases if no list_command() is provided. Some process tests do not have a way to list tests and some do not have the concept of test suites so adding a default suite is required.

ryan-collingham:
Also Travis is blocked for infra reasons, not related to this PR. I will look into fixing but the important thing is that the private CI is passing.

ryan-collingham:
Travis is fixed, should be good to merge.

PR #116, opened 2019-03-04
XML exporter fix and add PyTest listing

ryan-collingham:
- Fix XML exporter so that it can handle nested TestGroupReports (used e.g. by parametrized testcases).
- Add listing function to the PyTest runner. Source modules are used as the suite names and test functions as the testcases.


ryan-collingham:
private CI passed :)

PR #117, opened 2019-03-11
Add file logging handler and new Loggable base class

ryan-collingham:
Adds a file logging handler which writes logs to file with a
default level of DEBUG and with extra debug information included
in the formatter (timestamp, thread/process info, log name). Also
adds a new Loggable base class that provides object-based logging
via the self.logger attribute.


ryan-collingham:
Private CI has passed.

PR #119, opened 2019-03-14
Add custom value comparators to dictmatch

ryan-collingham:
Adds a new value_cmp_func kwarg to DictNamespace.match(),
which allows users to set a custom function for comparing
all values in a dict (versus adding a custom function for
comparing a value for a specific key). Built-in functions are defined
in comparison.CompareFunctions enum. The default option is to
use plain operator.eq - previously the default comparison
method was to always stringify.

Adds UTs for dict.match covering the new functionality.

Also update FixNamespace.match for typed vs untyped FIX matches.
With untyped or mixed, values are compared as strings. Else for
typed FIX messages the types will be checked explicitly.

### DIFF
--- a/examples/Assertions/Basic/test_plan.py
+++ b/examples/Assertions/Basic/test_plan.py
@@ -559,6 +559,33 @@ class SampleSuite(object):
         result.dict.match(
             actual, expected, description='Dict match: Custom comparators')
 
+        # You can also specify a comparator function to apply to all values in
+        # your dict. Standard comparators are available under
+        # testplan.common.utils.comparison.COMPARE_FUNCTIONS but any function
+        # f(x: Any, y: Any) -> bool can be used.
+        actual = {'foo': 1, 'bar': 2, 'baz': 3}
+        expected = {'foo': 1.0, 'bar': 2.0, 'baz': 3.0}
+
+        result.dict.match(
+            actual,
+            expected,
+            description='default assertion passes because the values are '
+                        'numerically equal')
+        result.dict.match(
+            actual,
+            expected,
+            description='when we check types the assertion will fail',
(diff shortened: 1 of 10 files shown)


ryan-collingham:
Need to fix python2 build. The python3 build failed for unrelated infra reasons.

ryan-collingham:
Python3 build fixed. I'm not sure the `pip install --upgrade pip` change is actually required or not but it should be harmless to leave in.

lambchr:
Looks really good, I like the idea of letting the user pass in a custom comparison function for dict values.

examples/Assertions/Basic/test_plan.py
```diff
@@ -559,6 +559,33 @@ def test_dict_namespace(self, env, result):
         result.dict.match(
             actual, expected, description='Dict match: Custom comparators')
 
+        # You can also specify a comparator function to apply to all values in
+        # your dict. Standard comparators are available under
+        # testplan.common.utils.comparison.CompareMethods but any function
```
ryan-collingham: need to change this name to COMPARE_FUNCTIONS

examples/Assertions/Basic/test_plan.py
```diff
@@ -559,6 +559,33 @@ def test_dict_namespace(self, env, result):
...
+        # f(x: Any, y: Any) -> bool can be used.
+        actual = {'foo': 1, 'bar': 2, 'baz': 3}
+        expected = {'foo': 1.0, 'bar': 2.0, 'baz': 3.0}
+
+        result.dict.match(
+            actual,
+            expected,
+            description='default assertion passes because the values are'
```
ryan-collingham: space missing from end of this line

PR #122, opened 2019-03-18
Warn if command line argument takes no effect

raoyitao:
* User can specify argument for running testplan by programmatically
  writting it in script or by passing it in command line. If argument
  specified on both side, command line argument will be overridden. We
  should print warning to let user know it.

### DIFF
--- a/testplan/base.py
+++ b/testplan/base.py
@@ -122,6 +122,10 @@ class Testplan(RunnableManager):
         self._parsed_args = self.parser.parse_args()
         self._processed_args = self.parser.process_args(self._parsed_args)
         for key, value in self._processed_args.items():
+            if key in options:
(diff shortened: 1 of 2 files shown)


testplan/base.py
```diff
@@ -122,6 +122,10 @@ def _enrich_options(self, options):
         self._parsed_args = self.parser.parse_args()
         self._processed_args = self.parser.process_args(self._parsed_args)
         for key, value in self._processed_args.items():
+            if key in options:
+                self.logger.warning('WARNING: Command line argument '
+                    'for "{}" will be overridden by that specified in '
+                        'testplan script'.format(key))
```
JohnChiotis: What about that:

"... will be overridden by the one programmatically  defined in {} constructor".format(key, self)

PR #123, opened 2019-03-18
Mock TestRunner to do time-sensitive test for timeout feature

raoyitao:
* Time sensitive testcase should not depend on system time, operating
  system might be busy scheduling processes with CPU and IO resources,
  sometimes the Multitest tasks (especially executed in pool) cannot be
  finished within specified time, testcase would be unstable. So just
  make some tasks which would take long time to execute, and in this
  mocked TestRunner we can detect these tasks remaining in executors,
  then we can start timeout event handling.

### DIFF
--- a/test/functional/testplan/test_timeout.py
+++ b/test/functional/testplan/test_timeout.py
@@ -1,6 +1,9 @@
 import os
+import time
 
 from testplan import Testplan
+from testplan.runnable import TestRunner
+from testplan.runners.base import Executor
 from testplan.runners.pools import ThreadPool, ProcessPool
 from testplan.runners.pools.tasks import Task
 from testplan.report.testing import Status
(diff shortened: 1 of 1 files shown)


test/functional/testplan/test_timeout.py
```diff
@@ -10,29 +13,92 @@
...
+                        pending_work = True
+                    # For each executer, there will be 2 tasks added, one is
+                    # light weight and the other takes long time to run. When
+                    # there is only one task remaining in each executor it is
+                    # indicated that a timeout event should occur.
+                    if self.cfg.timeout and len(resource.ongoing) > 1:
+                        timeout_flag = False
+
+            if not pending_work:
+                # Wll not happen, timeout event shold occur before all tasks
```
JohnChiotis: should - typo

test/functional/testplan/test_timeout.py
```diff
@@ -10,29 +13,92 @@
...
+                # finished. Or there should me something wrong and test fails.
```
JohnChiotis: me - typo

test/functional/testplan/test_timeout.py
```diff
@@ -10,29 +13,92 @@
...
+                    # Poll the resource's health to avoid hanging.
+                    if not resource.is_alive:
+                        self.result.test_report.logger.critical(
+                            'Aborting {} - {} unexpectedly died'.format(
+                                self, resource))
+                        self.abort()
+                        self.result.test_report.status_override = Status.ERROR
+                    if len(resource.ongoing) > 0:
+                        pending_work = True
+                    # For each executer, there will be 2 tasks added, one is
```
JohnChiotis: executor - typo

PR #124, opened 2019-03-25
Replace report_all flag with report_mode option

ryan-collingham:
Instead of using a report_all flag to control reporting of
dict and fix match results, add a report_mode option.
This can be set to one of the values of the new ReportOptions
enum to control the reporting behaviour. Current reporting
options are ALL (default), NO_IGNORED or FAILS_ONLY but more
could be added.


ryan-collingham:
I did this change as part of porting the ignore_passed flag. Since ignore_passed and report_all are both flags to control how comparisons are reported in a dict/FIX match I thought it would be cleaner to combine them into a single enum parameter.

PR #127, opened 2019-03-28
User Interface

lambchr:
* Web server started locally by --ui or --browse CLI args.
* Install & build handled by install-testplan-ui script.
* UI source code written in React components.

Also including a fix to stop adding None to sys.path in child workers.


test/unit/testplan/web_ui/test_web_app.py
```diff
@@ -93,60 +95,75 @@ def setup_method(self, _):
         self.static_dir = tempfile.mkdtemp()
         _create_tmp_static_files(self.static_dir)
         tp_web_app.config['STATIC_PATH'] = self.static_dir
+        tp_web_app.config['TESTPLAN_REPORT_NAME'] = 'report.json'
         tp_web_app.config['TESTING'] = True
         self.client = tp_web_app.test_client()
 
     def teardown_method(self, _):
         """Remove fake static files."""
+        # self.web_server_thread.stop()
```
JohnChiotis: What's that?

lambchr: It was left over from an earlier version of the test. I've removed it thanks.

PR #130, opened 2019-04-01
By default continue executing cases even exception raised

raoyitao:
* The current behavior is that if any exception raised from a testcase,
  then stop executing the following testcases in the same test suite,
  and turn to execute the next test suite. Change this behavior because
  generally all testcases should not depend on each other, but we can
  provide an option in Multitest constructor if use want to stop at
  once to save time.

### DIFF
--- a/testplan/testing/multitest/base.py
+++ b/testplan/testing/multitest/base.py
@@ -95,6 +95,7 @@ class MultiTestConfig(TestConfig):
             ConfigOption('result', default=Result): is_subclass(Result),
             ConfigOption('thread_pool_size', default=0): int,
             ConfigOption('max_thread_pool_size', default=10): int,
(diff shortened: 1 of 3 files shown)


testplan/testing/multitest/base.py
```diff
@@ -95,6 +95,7 @@ def get_options(cls):
             ConfigOption('result', default=Result): is_subclass(Result),
             ConfigOption('thread_pool_size', default=0): int,
             ConfigOption('max_thread_pool_size', default=10): int,
+            ConfigOption('stop_on_exception', default=False): bool,
```
JohnChiotis: Maybe stop_on_error is a more meaningful name.
Also by default we should stop on first exception as always, so the default value should be True.

PR #132, opened 2019-04-01
Serialization fixes

ryan-collingham:
**Serialization fix for custom result types**

Fix so that only types that are known to be serializable
are returned as their native type for result serialization.
Types that inherit from these primitive types (e.g. int)
may not necessarily also be serializable, so replace an
isintance() check with an explicit equality check of type().

**Check if Task is being scheduled from the main module**

Tasks shouldn't be scheduled to ProcessPools from the special
__main__ module. __main__ will be a different module for Process
workers (the child.py module to be exact) and so target from the
original __main__ will not be found - or worse will be replaced
by an attribute from child.py.


ryan-collingham:
Hmm there is some issue with the tests I need to look into before this can be merged.

ryan-collingham:
This is superseded by #139

PR #134, opened 2019-04-04
Altering App driver & LogMatcher

lambchr:
Needed for internal changes.

### DIFF
--- a/testplan/testing/multitest/driver/base.py
+++ b/testplan/testing/multitest/driver/base.py
@@ -126,8 +126,22 @@ class Driver(Resource):
         """Driver uid."""
         return self.cfg.name
 
+    def start(self):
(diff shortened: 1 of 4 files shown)


lambchr:
Internal private job passed.

testplan/testing/multitest/driver/base.py
```diff
@@ -126,8 +126,22 @@ def uid(self):
         """Driver uid."""
         return self.cfg.name
 
+    def start(self):
```
ryan-collingham: Start and stop logic already lives in Resource, from which Driver inherits. Can we move pre_start/pre_stop logic into Resource too?

If not, at least make a super() call here for the status-changing and starting() logic.

PR #139, opened 2019-04-10
ProcessPool fixes

ryan-collingham:
- Fixes pool serialization edge cases:
    Handle serialization of types that inherit from "safe" builtins like int that may not themselves be safe to 
    pickle - e.g. the Boost.Python.enum type was seen to cause pickle failures despite inheriting from int.
    Also, don't allow users to schedule tests from the __main__ module in Process pool, since the __main__
    module is different for the workers than in the parent process.

- Rename test/ dir to tests/
    To avoid a namespace clash with the python3 standard lib "test" package 
    (https://docs.python.org/3/library/test.html)

- Ensure TCP sockets are always cleaned up by ProcessPool
    Make the TCPConnectionManager used by the ProcessPool into a Resource that can be started
    to create TCP sockets and stopped to clean them up after. This avoids hanging seen in tests due
    to lack of resource cleanup.

- Small refactor of Pool internals
    Split up the large message handling function into seperate functions to handle each message type
    expected from workers.


ryan-collingham:
I think this PR has gotten a bit too big and amorphous with various attempts to fix the CI tests. In hindsight renaming test/ -> tests/ was a bigger change than I thought it would be and should be done seperately to the ProcessPool related changes and the other changes related to Travis CI infra.

PR #141, opened 2019-04-11
Fixing thread logging

lambchr:
Some threads were throwing exceptions when we tried to call __repr__() on them in Python 3. Changed to just log the thread name.


ryan-collingham:
Interesting. If this logging causes more problems we can just remove it, it is not essential.

PR #142, opened 2019-04-11
Update setup.py requirements, use them from requirements.txt

ryan-collingham:
Dependencies in setup.py and requirements.txt were out of sync, so I've
updated the setup.py requirements to match. To avoid further
duplication, I've updated the requirements.txt to include the local
setup.py requirements.

This change should fix the Travis CI build issue affecting some PRs.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,12 +5,10 @@ python:
   - "3.4"
 
 install:
-  - pip install -r requirements.txt --upgrade
-  - pip install --upgrade pylint
-  - python setup.py develop
+  - pip install -r requirements.txt -U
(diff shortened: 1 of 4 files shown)


ryan-collingham:
Following feedback from @lambchr I have updated the "getting started" docs to cover direct `pip` installation as used in the Travis job.

We should not encourage users to call setup.py directly due to https://github.com/pypa/setuptools/issues/498

ryan-collingham:
@lambchr OK Travis is happy now it seems, could you pelase review the docs changes I made when you have a moment? Would be good to get this merged fairly quickly to unblock other PRs.

PR #149, opened 2019-04-12
Parallelise tests and cache pip installs

ryan-collingham:
Use pytest-xdist to run the tests in parallel for speed.
Enable pip caching in Travis to reduce install time.


ryan-collingham:
Thought I would try this out - but it seems that our tests are not thread safe. We can look into options for process-based parallelism if needed.

PR #150, opened 2019-04-12
Fix serialization edge cases for Task results

ryan-collingham:
Fix serialisation of result types so that only the "safe" builtin types
such as int, str, NoneType etc. are directly pickled, all other objects
are converted to their string representation. In particular, types that
inherit from a "safe" builtin are no longer directly pickled - this was
causing problems with Boost.Python.enum type.

Also add a simple check to prevent scheduling of tasks from the
__main__ module for Process pool - this will always fail since __main__
is a completely different module for the child worker processes.

### DIFF
--- a/testplan/common/serialization/fields.py
+++ b/testplan/common/serialization/fields.py
@@ -51,11 +51,17 @@ def native_or_pformat(value):
     elif callable(value):
         value = getattr(value, '__name__', _repr_obj(value))
 
-    result = value if isinstance(value, COMPATIBLE_TYPES) else pprint.pformat(value)
-    obj_repr = _repr_obj(result)
+    # For basic builtin types we return the value unchanged. All other types
+    # will be formatted as strings.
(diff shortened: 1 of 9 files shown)


ryan-collingham:
Hmm Travis isn't happy, I need to revisit the pool changes after Easter I think. Still if you could take a look and let me know your thoughts on the design/whether it clashes with your pool changes that would be handy.

I think the key thing is we need a clear design on who owns the monitor thread - currently it is started by the _loop() method but doesn't terminate until after _loop() has exited, and nothing ensures that it does terminate at all. So I think that _loop() needs to wait for the monitor thread to terminate, and this requires a bit of a tweak to the monitor loop.

ryan-collingham:
There is some issue with test collection I need to investigate further.

ryan-collingham:
oh wait, was just missing an __init__.py, fixed.

PR #153, opened 2019-04-18
Webserver exporter improvements

ryan-collingham:
For the webserver exporter:
- Add a timeout to prevent webserver test from blocking indefinitely due to unexpected stdout from the testplan subprocess
- Fixes a bug when adding the WebServerExporter programmatically, if no port is specified the default value (0) will be used.

(removed previous pool-related changes from this PR and updated title/description)

### DIFF
--- a/testplan/exporters/testing/webserver/base.py
+++ b/testplan/exporters/testing/webserver/base.py
@@ -24,13 +24,11 @@ class WebServerExporterConfig(ExporterConfig):
     def get_options(cls):
         return {
             ConfigOption(
-                'ui_port', default=defaults.WEB_SERVER_PORT,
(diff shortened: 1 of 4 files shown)


ryan-collingham:
Hmm Travis isn't happy, I need to revisit the pool changes after Easter I think. Still if you could take a look and let me know your thoughts on the design/whether it clashes with your pool changes that would be handy.

I think the key thing is we need a clear design on who owns the monitor thread - currently it is started by the _loop() method but doesn't terminate until after _loop() has exited, and nothing ensures that it does terminate at all. So I think that _loop() needs to wait for the monitor thread to terminate, and this requires a bit of a tweak to the monitor loop.

PR #155, opened 2019-04-24
Refactor Pool/Worker communication related code

Pyifan:
- New abstract base class Client & Server
- Queue based implementation for base Pool (thread pool): QueueClient & QueueServer
- ZMQ based implimentation for process/remote/treadmill pool: ZMQClient & ZMQServer
- Move related code to testplan/runners/pools/connection.py

### DIFF
--- a/testplan/runners/pools/base.py
+++ b/testplan/runners/pools/base.py
@@ -1,17 +1,14 @@
 """Worker pool executor base classes."""
 
-import abc
+
 import inspect
-import logging
 import numbers
 import os
 import psutil
-
 import threading
 import time
 
 from schema import Or, And
-import six
 
 from testplan.common.config import ConfigOption, validate_func
 from testplan.common import entity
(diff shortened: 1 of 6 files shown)


testplan/runners/pools/base.py
```diff
@@ -799,15 +611,16 @@ def _add_workers(self):
...
             self.logger.debug('Added worker %(index)s (outfile = %(outfile)s)',
                               {'index': idx, 'outfile': worker.outfile})
 
     def starting(self):
         """Starting the pool and workers."""
         with self._pool_lock:
             self._conn.start()
-            for worker in self._workers:
-                self._conn.register(worker)
+            if not len(self._workers):
```
Pyifan: It feels a bit too scholastic to have to add all workers again if we restart the pool, and use case is limited in our own test. I think this would do, for now.

testplan/runners/pools/connection.py
```diff
@@ -2,46 +2,348 @@
...
+
+    def send_and_receive(self, message, expect=None):
+        """
+        Send and receive shortcut. Optionally assert that the response is
+        of the type expected. I.e For a TaskSending message, an Ack is expected.
+
+        :param message: Message sent.
+        :type message: :py:class:`~testplan.runners.pools.communication.Message`
+        :param expect: Assert message received command is the expected.
+        :type expect: ``NoneType`` or
```
ryan-collingham: I'm a bit confused about the "expect" parameter - it is documented as being a `Message` type but is actually compared with the message's `cmd` string. Assuming it is meant to be the command, can we change the name to `expected_cmd` to be more explicit and update the documented type to `str`?

testplan/runners/pools/connection.py
```diff
@@ -2,46 +2,348 @@
...
+        self.connect()  # auto connect
+
+    def connect(self):
+        """Connect to a ZMQ Server"""
+        self._context = zmq.Context()
+        self._sock = self._context.socket(zmq.REQ)
+        self._sock.connect("tcp://{}".format(self._address))
+        self.active = True
+
+    def disconnect(self):
```
ryan-collingham: Could we add a `__del__` method to ensure that the sockets are always closed as a fallback even if `disconnect()` has not been called - like we have for ZMQServer?

testplan/runners/pools/connection.py
```diff
@@ -2,46 +2,348 @@
...
+        """Server stopping logic."""
+
+        self._unregister_workers()
+        self.status.change(self.status.STOPPED)
+
+    def aborting(self):
+        """Abort policy - no abort actions are required in the base class."""
+        pass
+
+    @abc.abstractmethod
```
ryan-collingham: I'm not sure that the register/unregister methods need to be marked as abstract - do they always need to be overridden? The base implementation seems useful.

testplan/runners/pools/connection.py
```diff
@@ -2,46 +2,348 @@
...
+    def accept(self):
+        """
+        Accepts a new message from worker. This method should not block - if
+        no message is queued for receiving it should return None.
+
+        :return: Message received from worker transport, or None.
+        :rtype: ``NoneType`` or
+            :py:class:`~testplan.runners.pools.communication.Message`
+        """
+        pass
```
ryan-collingham: prefer `raise NotImplementedError` here

testplan/runners/pools/connection.py
```diff
@@ -2,46 +2,348 @@
...
-            self._close()
-
     def starting(self):
         """Create a ZMQ context and socket to handle TCP communication."""
         if self.parent is None:
             raise RuntimeError('Parent pool was not set - cannot start.')
 
-        self._context = zmq.Context()
-        self._sock = self._context.socket(zmq.REP)
+        self._zmq_context = zmq.Context()
```
ryan-collingham: Good name change :+1:

PR #159, opened 2019-04-30
Use six.moves to import modules

raoyitao:
* To avoid a lot of try... except... ImportError sentences.
  Better adopt py3 representation notation for module names.

### DIFF
--- a/testplan/vendor/tempita/__init__.py
+++ b/testplan/vendor/tempita/__init__.py
@@ -30,16 +30,17 @@ from __future__ import absolute_import, division, print_function
 
 import re
 import sys
+import os
+import tokenize
+
 try:
-    from urllib.parse import quote as url_quote
     from io import StringIO
     from html import escape as html_escape
 except ImportError:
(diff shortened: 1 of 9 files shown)


testplan/vendor/tempita/__init__.py
```diff
@@ -30,16 +30,17 @@ def foo(bar):
 
 import re
 import sys
+import os
+import tokenize
+
 try:
-    from urllib.parse import quote as url_quote
     from io import StringIO
```
ryan-collingham: `from six import StringIO` to avoid try/except here?

testplan/vendor/tempita/__init__.py
```diff
@@ -30,16 +30,17 @@ def foo(bar):
...
     from html import escape as html_escape
```
ryan-collingham: I think `cgi.escape` exists in both python 2 and 3. Any particular reason to use `html.escape` instead for python3?

PR #161, opened 2019-05-07
Fix instable restart_worker test

Pyifan:
Also move worker.transport.disconnect from worker to pool's stopping - only if pool is restarting would worker need to reconnect to a new server.


ryan-collingham:
Regarding the failure blocking the Travis build: I've seen that failure before, I'm pretty sure it's not caused by your changes. We should temporarily disable that unstable test. I am intending to take a look at the reload functionality to make sure that test is stable when I get the time.

PR #181, opened 2019-05-13
Change pool start sequence

Pyifan:
Change pool start sequence: conn -> loop & monitor -> workers, stop is reverse.
Fix a few corner cases of status transition.
Enable process pool testcase to see if still any random failure.

### DIFF
--- a/testplan/runners/pools/base.py
+++ b/testplan/runners/pools/base.py
@@ -250,6 +250,7 @@ class Pool(Executor):
         self.make_runpath_dirs()
         self._metadata['runpath'] = self.runpath
         self._exit_loop = False
+        self._start_monitor_thread = True
 
         # Methods for handling different Message types. These are expected to
         # take the worker, request and response objects as the only required
@@ -302,10 +303,12 @@ class Pool(Executor):
         started.
         """
 
-        self.logger.debug('Starting worker monitor thread.')
-        self._worker_monitor = threading.Thread(target=self._workers_monitoring)
-        self._worker_monitor.daemon = True
-        self._worker_monitor.start()
+        if self._start_monitor_thread:
+            self.logger.debug('Starting worker monitor thread.')
+            self._worker_monitor = threading.Thread(
+                target=self._workers_monitoring)
+            self._worker_monitor.daemon = True
+            self._worker_monitor.start()
 
         while self.active and not self._exit_loop:
             msg = self._conn.accept()
--- a/testplan/runners/pools/base.py
+++ b/testplan/runners/pools/base.py
@@ -22,7 +22,7 @@ from testplan.runners.base import Executor, ExecutorConfig
 from .communication import Message
 from .connection import QueueClient, QueueServer
 from .tasks import Task, TaskResult
-
+from testplan.common.entity import ResourceStatus
 
 class WorkerConfig(entity.ResourceConfig):
     """
@@ -123,7 +123,8 @@ class Worker(entity.Resource):
     def _loop(self, transport):
         message = Message(**self.metadata)
 
-        while self.active:
+        while (self.active and self.status.tag not in (self.status.STOPPING,
+                                                       self.status.STOPPED)):
             received = transport.send_and_receive(message.make(
                 message.TaskPullRequest, data=1))
             if received is None or received.cmd == Message.Stop:
@@ -248,6 +249,7 @@ class Pool(Executor):
         self._metadata = {}
         self.make_runpath_dirs()
         self._metadata['runpath'] = self.runpath
+        self._exit_loop = False
 
         # Methods for handling different Message types. These are expected to
         # take the worker, request and response objects as the only required
@@ -305,16 +307,12 @@ class Pool(Executor):
         self._worker_monitor.daemon = True
         self._worker_monitor.start()
 
-        while (self.active
-               and self.status.tag is not self.status.STOPPING
-               and self.status.tag is not self.status.STOPPED):
-
+        while self.active and not self._exit_loop:
             msg = self._conn.accept()
             if msg:
                 try:
                     self.logger.debug('Received message from worker: %s.', msg)
-                    with self._pool_lock:
-                        self.handle_request(msg)
+                    self.handle_request(msg)
                 except Exception as exc:
                     self.logger.error(format_trace(inspect.trace(), exc))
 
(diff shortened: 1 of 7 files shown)


testplan/runners/pools/base.py
```diff
@@ -330,10 +328,9 @@ def handle_request(self, request):
         sender_index = request.sender_metadata['index']
         worker = self._workers[sender_index]
         if not worker.active:
```
Pyifan: always respond to worker

testplan/runners/pools/base.py
```diff
@@ -123,7 +123,8 @@ def is_alive(self):
     def _loop(self, transport):
         message = Message(**self.metadata)
 
-        while self.active:
```
Pyifan: break the loop if pool is stopping

testplan/runners/pools/base.py
```diff
@@ -305,16 +307,12 @@ def _loop(self):
...
-        while (self.active
-               and self.status.tag is not self.status.STOPPING
-               and self.status.tag is not self.status.STOPPED):
-
+        while self.active and not self._exit_loop:
             msg = self._conn.accept()
             if msg:
                 try:
                     self.logger.debug('Received message from worker: %s.', msg)
-                    with self._pool_lock:
```
Pyifan: Trying to remove all the pool lock

testplan/runners/pools/base.py
```diff
@@ -471,27 +468,41 @@ def _workers_monitoring(self):
...
-                        'All workers of {} are inactive.'.format(self))
-                    self.abort()
-                    break
+
+            for worker in self._workers:
+                status, reason = self._query_worker_status(worker)
+                if status == 'inactive':
+                    if self.active and self.status.tag not in (
+                            self.status.STOPPING, self.status.STOPPED):
+                        if self._handle_inactive(worker, reason):
```
Pyifan: Handle inactive only if pool is not stopping

testplan/runners/pools/base.py
```diff
@@ -505,56 +516,53 @@ def _query_worker_status(self, worker):
...
     def _handle_inactive(self, worker, reason):
         """
         Handle an inactive worker.
         :param worker: worker object
-        :param reason: reason of call this method
+        :param reason: why worker is considered inactive
         :return: True if worker restarted, else False
         """
+        if worker.status.tag != worker.status.STARTED:
+            return
```
Pyifan: Handle inactive only if worker is in STARTED status

testplan/runners/pools/base.py
```diff
@@ -643,11 +657,14 @@ def workers_requests(self):
...
         """Stop connections and workers."""
         # TODO do we need a lock here?
-        super(Pool, self).stopping()  # stop the loop and the monitor
 
+        self._workers.stop()
         for worker in self._workers:
             worker.transport.disconnect()
-        self._workers.stop()
+
+        self._exit_loop = True
```
Pyifan: set _exit_loop after all workers are stopped.
We cannot use pool.status.tag == stopping for exiting loop because threadmill workers needs the loop running til worker stopped.

testplan/runners/pools/base.py
ryan-collingham: Any particular reason this change was needed? We can try it, but needs testing to make sure the work thread and monitor thread won't interact badly.

testplan/runners/pools/base.py
ryan-collingham: There's still a window here where the status could become STOPPING before/during handling the inactive worker. Is that OK?

testplan/runners/pools/base.py
```diff
@@ -620,19 +628,25 @@ def starting(self):
...
+
         if not self._workers:
             self._add_workers()
         self._start_workers()
 
         if self._workers.start_exceptions:
             for msg in self._workers.start_exceptions.values():
                 self.logger.error(msg)
-            self._workers.stop()
+            self.stop()
```
ryan-collingham: I think should call `self.abort()` - stop is for stopping under normal conditions, abort is for error conditions.

testplan/runners/pools/base.py
```diff
@@ -662,11 +679,12 @@ def aborting(self):
         """Aborting logic."""
         self.logger.debug('Aborting pool {}'.format(self))
 
-        super(Pool, self).stopping()  # stop the loop and the monitor
         for worker in self._workers:
             worker.abort()
+
+        super(Pool, self).stopping()  # stop the loop and the monitor
```
ryan-collingham: This should call aborting() in the parent class not stopping() - we should not wait to join() the work loop when aborting.

tests/functional/testplan/runners/pools/test_pool_base.py
```diff
@@ -31,6 +31,7 @@ def schedule_tests_to_pool(pool, **pool_cfg):
     plan = Testplan(
         name='Plan',
         parse_cmdline=False,
+        logger_level=10
```
ryan-collingham: Did you mean to include changes to this file?

tests/functional/testplan/runners/pools/test_pool_process.py
```diff
@@ -17,7 +17,7 @@
 from .func_pool_base_tasks import schedule_tests_to_pool
 
 
-pytestmark = pytest.mark.skipif(True, reason='Process Pool tests are unstable')
+# pytestmark = pytest.mark.skipif(True, reason='Process Pool tests are unstable')
```
ryan-collingham: Same as above - did you mean to leave this line commented out? I think we will need more testing before we can say the pool tests are stable.

testplan/runners/pools/base.py
```diff
@@ -330,10 +331,9 @@ def handle_request(self, request):
         sender_index = request.sender_metadata['index']
         worker = self._workers[sender_index]
         if not worker.active:
-            self.logger.critical(
-                'Ignoring message {} - {} from inactive worker {}'.format(
+            self.logger.warning(
```
ryan-collingham: To be honest I'm not sure this should be a warning - I think it is quite expected when processing a message that the worker may have died. Let's leave as a warning for now but if it gets too spammy we should reduce to a debug log.

testplan/runners/pools/base.py
ryan-collingham: This hasn't been resolved?

tests/functional/testplan/runners/pools/test_pool_process.py
ryan-collingham: This hasn't been resolved either?

tests/functional/testplan/runners/pools/test_pool_remote.py
```diff
@@ -14,7 +14,7 @@
 
 IS_WIN = platform.system() == 'Windows'
 
-pytestmark = pytest.mark.skipif(True, reason='Remote Pool tests are unstable')
+# pytestmark = pytest.mark.skipif(True, reason='Remote Pool tests are unstable')
```
ryan-collingham: please uncomment

Issue #182, opened 2019-05-14
LogMatcher is slow for matching large files


ryan-collingham:
When matching moderately large log files (thousands of lines or more) the LogMatcher takes a very long time to match lines. This is due to it sleeping for a fixed time between each match iteration. Fix is simple - only sleep when there is no line read (i.e. we are waiting for the application to produce a log line), not when there are unread lines still to process.

PR #183, opened 2019-05-14
Logmatcher improvements

ryan-collingham:
Improve performance of LogMatcher by only sleeping when there are no more lines to read - not when iterating through lines already there. Add test for matching a more realistically sized logfile (10k lines).

Making the above fix exposed a bug in the tests - the seek/tell methods do nothing for objects returned by mock_open! The test_match_only_searches_after_position() testcase was only passing because somehow it was actually timing out after 0.5s. Resolved by simplifying tests to read from real (temporary) logfiles, not by using mock_open.

Also includes a small change to the match() method to accept a regex string as an alternative to a compiled regex - it will compile the regex automatically.

Fixes https://github.com/Morgan-Stanley/testplan/issues/182

### DIFF
--- a/testplan/common/utils/match.py
+++ b/testplan/common/utils/match.py
@@ -3,6 +3,9 @@ Module of utility types and functions that perform matching.
 """
 import os
 import time
+import re
+
+import six
 
 from . import timing
 from . import logger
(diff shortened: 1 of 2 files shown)


testplan/common/utils/match.py
```diff
@@ -95,28 +101,42 @@ def match(self, regex, timeout=5):
...
                     match = regex.match(line)
                     if match:
-                        self.position = log.tell()
                         break
-                if time.time() > end_time:
-                    self.position = log.tell()
-                    break
-                time.sleep(LOG_MATCHER_INTERVAL)
+                else:
+                    time.sleep(LOG_MATCHER_INTERVAL)
```
Pyifan: Use get_sleeper utility?

ryan-collingham: The `get_sleeper()` utility is for sleeping after every iteration I think? Which is not what this change is doing: we only want to sleep when no lines are read (i.e. we are waiting on the application to write a line), not when lines are being read - so that we can iterate quickly through thousands of lines without unnecessary sleeping.

Issue #199, opened 2019-05-30
Propogate sys.path down into process pool workers


ryan-collingham:
Currently when process pool workers are started they have a completely different sys.path to the parent process. This is confusing for users who might reasonably expect to be able to schedule modules that are already on their sys.path without explicitly specifying the path. Worse, if a task is scheduled from a module "tasks" it will actually try to find that task under testplan.runners.pools.tasks, since the directory of the child.py script is placed first on the child's sys.path!

As an enhancement we should try to replicate the parent's sys.path in the child at the point the child is spawned. We can do this simply by setting the PYTHONPATH environment variable before spawning the child process.

PR #200, opened 2019-05-30
Match same sys.path in child process as in parent

ryan-collingham:
Addresses https://github.com/Morgan-Stanley/testplan/issues/199


ryan-collingham:
There is some issue with setting such a long PYTHONPATH on Windows, so I'm replacing this PR with the file-based approach in  https://github.com/Morgan-Stanley/testplan/pull/201

PR #203, opened 2019-06-05
Add resource monitor that will export a JSON file includes:

yuxuan-ms:
Add resource monitor that will export a JSON file includes:
* test and Testplan internals (events) start and stop times (seconds)
* resource utilisation per host

To run Testplan with Resource Monitor just run your Testplan script with the --resource-monitor command line argument.

### DIFF
--- a/testplan/common/utils/monitor/__init__.py
+++ b/testplan/common/utils/monitor/__init__.py
@@ -7,21 +7,23 @@ import json
 import socket
 import logging
 import threading
-from testplan.common.utils.logger import TESTPLAN_LOGGER
+from collections import defaultdict
+from testplan.common.utils.logger import Loggable, TESTPLAN_LOGGER
 from testplan.common.utils.path import makedirs
 from .collectors import get_collector
 
 
-class EventMonitor(object):
+class EventMonitor(Loggable):
     """
     Event monitor will record event start and stop times(seconds)
     """
     def __init__(self, parent=None):
-        self.logger = TESTPLAN_LOGGER
+        super(EventMonitor, self).__init__()
         self.parent = parent
         self.hostname = socket.getfqdn()
         self.events_metadata = {}
-        self.events_data = {}
+        self.events_data = defaultdict(list)
+        # The lock protects self.events_metadata
         self.lock = threading.Lock()
         self.log_directory = None
         self.uid = str(uuid.uuid4())
@@ -42,7 +44,8 @@ class EventMonitor(object):
 
     def setup_logger(self, log_path=None):
         """
-        Sets up a unique logger for Monitor each time it is called. Should only be called once or logs will be directed
+        Sets up a unique logger for Monitor each time it is called. Should only
+        be called once or logs will be directed
         to new logger on subsequent call.
 
         :param log_path: Absolute path to file to save logs to.
@@ -51,7 +54,6 @@ class EventMonitor(object):
         :return: ``None``
         :rtype: ``NoneType``
         """
-        self.logger = logging.getLogger(str(uuid.uuid4()))
         if log_path:
             directory = log_path.rsplit(os.sep, 1)[0]
             self.log_directory = directory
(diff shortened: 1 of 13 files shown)


testplan/common/utils/monitor/collectors.py
```diff
@@ -0,0 +1,141 @@
...
+                        'cpu_percent',
+                        'io_counters',
+                        'create_time',
+                        'io_counters'
+                    ])
+
+            except psutil.NoSuchProcess:
+                continue
+            monitor_result['cpu'] += raw_data['cpu_percent'] \
+                if raw_data['cpu_percent'] > 0 else 0
```
ryan-collingham: CPU percent can be negative...?

testplan/common/utils/monitor/collectors.py
```diff
@@ -0,0 +1,141 @@
...
+                        'memory_info',
+                        'cpu_percent',
+                        'io_counters',
+                        'create_time',
+                        'io_counters'
+                    ])
+
+            except psutil.NoSuchProcess:
+                continue
+            monitor_result['cpu'] += raw_data['cpu_percent'] \
```
ryan-collingham: This monitor function is very bloated... can you split it into smaller functions?

In particular, it might be good to have a function called something like self._read_process_info which takes the raw_data and returns a tuple of cpu, memory, iops, read and write data for that process that you can add to the totals.

testplan/common/utils/monitor/collectors.py
```diff
@@ -0,0 +1,141 @@
...
+                            cgroup.split(':')[-1][1:],
+                            'memory.limit_in_bytes')
+                        with open(path, 'r') as memory_file:
+                            self.metadata['memory'] = int(memory_file.read())
+                except (IOError, AttributeError, ValueError):
+                    pass
+
+
+class WindowsCollector(Collector):
+    def __init__(self, director):
```
ryan-collingham: pointless method?

testplan/exporters/testing/monitor/__init__.py
```diff
@@ -0,0 +1,158 @@
...
+        ]
+        and order by start time.
+
+        :param source: Exporter instance
+        :type source: ``Exporter``
+
+        :return: ``None``
+        :rtype: ``NoneType``
+        """
+        run_path = self.cfg.runpath
```
ryan-collingham: This function is huuuge... please split into smaller chunks?

tests/functional/testplan/testing/test_resource_monitor.py
```diff
@@ -0,0 +1,150 @@
...
+    for meta, value in mock_host_metadata.items():
+        assert resource_monitor.host_metadata[meta] == value
+
+
+@testsuite
+class Alpha(object):
+
+    @testcase
+    def test_memory_a(self, env, result):
+        _tmp = 'testplan' * 1024 * 1024 * (1024 // 8)  # 1GB memory
```
ryan-collingham: We have these sort of tests internally that attempt to allocate real memory and test the resource monitor and they are very flaky. Please don't repeat in the OSS - mock out the real OS resource usage.

testplan/common/utils/monitor/__init__.py
```diff
@@ -0,0 +1,295 @@
...
+
+    def stop(self):
+        """
+        Stop collect metrics.
+
+        :return: ``None``
+        :rtype: ``NoneType``
+        """
+        if self._poll:
+            self._poll_start = False
```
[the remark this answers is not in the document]
yuxuan-ms: It's a deamon thread, so I don't think we need to join() it.

testplan/common/utils/monitor/collectors.py
yuxuan-ms: It's a known issue of psutil on Linux

ryan-collingham: You mean https://github.com/giampaolo/psutil/issues/392 ? Looks like it was fixed. It's fine to be defensive just in case though, was mostly just curious.

testplan/common/utils/monitor/__init__.py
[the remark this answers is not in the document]
ryan-collingham: You should always join() threads when they are exiting as a matter of good practice. Marking as a daemon is just to ensure that it doesn't block the interpreter from exiting in case we abort (e.g. due to KeyboardInterrupt), but for normal exit we wait for threads to terminate.

PR #209, opened 2019-06-25
Remote worker start/stop in thread pool refactor

Pyifan:
A couple of changes:

- add start/stop_in_pool methods to Environment class to start/stop resources in threadpool.
- add try-except when we fetch result from remote host

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -139,7 +139,7 @@ class Environment(object):
         for resource in self._resources.values():
             try:
                 resource.start()
-                if resource.cfg.async_start is False:
+                if not resource.cfg.async_start:
                     resource.wait(resource.STATUS.STARTED)
             except Exception as exc:
                 msg = 'While starting resource [{}]{}{}'.format(
@@ -179,7 +179,7 @@ class Environment(object):
         """
 
         for resource in self._resources.values():
-            if resource.cfg.async_start is False:
+            if not resource.cfg.async_start:
                 raise RuntimeError(
                     'Cannot start resource {} in thread pool, '
                     'its async_start attr is set to False'.format(
@@ -343,7 +343,7 @@ class EntityConfig(Config):
                 block_propagation=False): Or(None, str, lambda x: callable(x)),
             ConfigOption('initial_context', default={}): dict,
             ConfigOption('path_cleanup', default=False): bool,
-            ConfigOption('status_wait_timeout', default=3600): int,
+            ConfigOption('status_wait_timeout', default=600): int,
             ConfigOption('abort_wait_timeout', default=30): int,
             # active_loop_sleep impacts cpu usage in interactive mode
             ConfigOption('active_loop_sleep', default=0.005): float
@@ -1183,9 +1183,9 @@ class Resource(Entity):
     def restart(self, timeout=None):
         """Stop and start the resource."""
         self.stop()
-        self._wait_stopped(timeout=timeout)
+        self.wait(self.status.STOPPED)
         self.start()
-        self._wait_started(timeout=timeout)
+        self.wait(self.status.STARTED)
 
     def __enter__(self):
         self.start()
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -159,6 +159,39 @@ class Environment(object):
             else:
                 resource.wait(resource.STATUS.STARTED)
 
+    def _log_exception(self, resource, func):
+
+        def wrapper(*args, **kargs):
+            try:
+                func(*args, **kargs)
+            except Exception as exe:
+                msg = 'While executing {} of resource [{}]{}{}'.format(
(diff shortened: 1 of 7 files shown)


Pyifan:
A couple of changes:
- add start/stop_in_pool methods to Environment class to start/stop resources in threadpool.
- add timestamp/thread id to logging
- add try-except when we fetch result from remote host

testplan/runners/pools/remote.py
```diff
@@ -535,23 +544,26 @@ def _write_syspath(self):
...
 
+    def _wait_stopped(self, timeout=None):
+        sleeper = get_sleeper(1, 300)
+        while next(sleeper):
+            if self.status.tag != self.status.STOPPED:
+                self.logger.info('Waiting for workers to stop')
+            else:
+                break
+        else:
+            self.logger.warning(
```
ryan-collingham: Maybe we should raise an exception at this point instead of just continuiing?

testplan/runners/pools/remote.py
```diff
@@ -704,19 +716,54 @@ def _add_workers(self):
...
+
+        sleeper = get_sleeper(1, 300)
+        while next(sleeper):
+            if any([worker.status.tag == worker.status.STARTING
+                   for worker in self._workers]):
+                self.logger.info('Waiting for workers to quit starting')
+            else:
+                break
+        else:
+            self.logger.warning(
```
ryan-collingham: again, I think we should raise an exception and/or call self.abort() here - we don't expect to stop while the workers are still starting?

testplan/common/entity/base.py
```diff
@@ -159,6 +159,39 @@ def start(self):
...
+    def _log_exception(self, resource, func):
+
+        def wrapper(*args, **kargs):
+            try:
+                func(*args, **kargs)
+            except Exception as exe:
+                msg = 'While executing {} of resource [{}]{}{}'.format(
+                    func.__name__, resource.cfg.name, os.linesep,
+                    format_trace(inspect.trace(), exe))
+                self.logger.error(msg)
```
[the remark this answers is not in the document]
Pyifan: copy-paste code, but i think the point here is to add this msg to self.start_exceptions.

testplan/common/entity/base.py
```diff
@@ -159,6 +159,39 @@ def start(self):
...
+
+        return wrapper
+
+    def start_in_pool(self, pool):
+        """
+        Start all resources concurrently in thread pool.
+        """
+
+        for resource in self._resources.values():
+            if resource.cfg.async_start is False:
```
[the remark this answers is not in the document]
Pyifan: copy-paste code, but sure

testplan/runners/pools/remote.py
Pyifan: I think by the time we are stopping, we should best effort stop everything that we can. If we raise here or abort, we are not cleaning up as much as possible. But this is indeed not very neat...

The problem is starting and stopping could overlap. Some treadmill worker could come up and starting executing tests and even finished all tests before other workers are still initializing. I'm thinking whether if we should be more rigid to have all resources readily started before we start to execute test.

testplan/runners/pools/remote.py
Pyifan: The problem is starting and stopping could overlap. Some treadmill worker could come up and starting executing tests and even finished all tests while other workers are still initializing. One worker could also be restarting when testplan determines all tests are finished and it's time to stop.

testplan/runners/pools/remote.py
Pyifan: hmmm, i'm rethinking this. I'm going to make some change to treadmill pool so that start/stop does not overlap.

testplan/common/entity/base.py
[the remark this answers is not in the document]
ryan-collingham: oh I see, msg is also added to start_excepttions - fine. Just bear in mind that in many cases self.logger.exception() will save you effort with manually formatting a stack trace!

And if you are copy-pasting code a lot, maybe it's time for a utility function? ;)

testplan/runners/pools/remote.py
ryan-collingham: I usually prefer to fail noisily (raise an Exception/abort etc.) if something unexpected happens, as that can make it easier to debug - rather than leaving your application in a weird inconsistent state that causes an issue far away and having to search back to find a relevant log!

testplan/runners/pools/remote.py
ryan-collingham: As above, let's try to prevent start/stop overlap.

Simple FSM: STARTING -> STARTED -> STOPPING -> STOPPED :)

Pyifan: this is just best effort, due to treadmill nature, late joiner worker or restarted worker could still be starting while pool's stop is called. So the logic is kept here, see internal PR for treadmill pool change.

PR #210, opened 2019-06-25
Improve public API with explicit argument list

raoyitao:
Improve public API with explicit argument list

* Document string that including arguments usage should be directly
  placed in every entity classes, for sphinx autodoc.
* For public APIs which might be directly used by testplan users, the
  class constructors should list all arguments from its config class
  as well as the required arguments and some significant ones from base
  config classes, to support IDE auto completion.
* Affected APIs include @test_plan, MultiTest and other Test (GTest,
  PyTest, PyUnit, etc), various pool runners and drivers.
* Add some docs for function argumets that are missing.
* Also fix some doc string not put in the correct placed.
* Annotate some testcases temporarily due to the command line arguments
  parsing issue, will be fixed in another pull request.

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -14,10 +14,14 @@ import psutil
 import functools
 from collections import deque, OrderedDict
 
-from schema import Or, And, Use
+try:
+    from collections.abc import MutableMapping, MutableSequence, MutableSet
+except ImportError:
+    from collections import MutableMapping, MutableSequence, MutableSet
 
-from testplan.common.config import Config
-from testplan.common.config import ConfigOption
+from schema import Optional, Or, And, Use
+
+from testplan.common.config import Config, ConfigOption
 from testplan.common.utils.exceptions import format_trace
 from testplan.common.utils.thread import execute_as_thread
 from testplan.common.utils.timing import wait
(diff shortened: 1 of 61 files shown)


testplan/common/entity/base.py
```diff
@@ -315,7 +315,22 @@ class Entity(logger.Loggable):
     CONFIG = EntityConfig
     STATUS = EntityStatus
 
-    def __init__(self, **options):
+    def __init__(self,
```
Pyifan: I don't think Entity is considered 'public API class/function', so it doesn't need to be changed.

testplan/common/entity/base.py
```diff
@@ -1054,7 +1069,11 @@ class Resource(Entity):
     CONFIG = ResourceConfig
     STATUS = ResourceStatus
 
-    def __init__(self, **options):
+    def __init__(self,
```
Pyifan: Same for Resource, sort of internal

testplan/runners/pools/base.py
```diff
@@ -236,7 +238,31 @@ class Pool(Executor):
...
+        size=4,
+        worker_type=Worker,
+        worker_heartbeat=None,
+        heartbeats_miss_limit=3,
+        task_retries_limit=3,
+        max_active_loop_sleep=5,
+        restart_count=3,
+        **options
+    ):
+        """
```
Pyifan: Remove?

testplan/testing/multitest/driver/base.py
```diff
@@ -112,7 +112,8 @@ class Driver(Resource):
 
     CONFIG = DriverConfig
 
-    def __init__(self, **options):
+    def __init__(self, install_files=None, timeout=60, logfile=None, log_regexps=None, stdout_regexps=None, stderr_regexps=None, async_start=False, report_errors_from_logs=False, error_logs_max_lines=10, **options):
```
Pyifan: you may want to format this.

testplan/common/entity/base.py
ryan-collingham: No definitely not. Don't worry about converting abstract bass classes like Entity, Resource etc. Only the ones that testplan users will be importing and initialising directly.

testplan/common/entity/base.py
```diff
@@ -315,7 +315,22 @@ class Entity(logger.Loggable):
...
+        runpath=None,
+        initial_context=None,
+        path_cleanup=False,
+        status_wait_timeout=3600,
+        abort_wait_timeout=30,
+        active_loop_sleep=0.005,
+        **options
+    ):
+        initial_context = initial_context or {}
+        options.update({'runpath': runpath,
```
ryan-collingham: As discussed, we might be able to filter locals() to remove this boilerplate.

testplan/common/utils/doc.py
```diff
@@ -0,0 +1,64 @@
...
+import os
+import re
+import six
+
+from testplan.common.config import Config, ConfigOption
+from testplan.common.entity import Entity
+from testplan.common.utils import logger
+
+
+def gendoc(cls):
```
ryan-collingham: This function is clever but unfortunately it's a bit too clever. We need docstrings to be defined statically so that a) they can be read by us when reading the code as reference and b) they can be understood by IDE highlighters and the builtin help() function.

testplan/testing/multitest/driver/base.py
ryan-collingham: yeah line is too long here and below.

PR #215, opened 2019-07-08
Command Line arguments take precedence over those defined in script

raoyitao:
* Generally, command line arguments should overwrite the corresponding
  ones defined in testplan script if they are explicitly specified by
  users, which provides flexibility for testing.
* We plan to provide public APIs with completely argument list for the
  ease of users with IDE, then this change is very important because
  the default value in API always takes precedence than that in command
  line. So, even a user does nnt explicitly assign a instance value for
  an argument in api, but specify it in command line, it has no effect.
* Also fix a typo (help message in testplan parser).


Pyifan:
I would suggest that we can use DefaultValueWrapper for command line option defaults. For simplicity and consistency.

Issue #218, opened 2019-07-11
LocalRunner does not print full stacktrace on materialization error


ryan-collingham:
Fix is simple - just make sure to log the full stack trace when handling the error.

PR #220, opened 2019-07-11
Log full stack for local materialization errors

ryan-collingham:
logger.exception automatically includes the formatted traceback when called in the except: branch of an exception handler.

Fixes https://github.com/Morgan-Stanley/testplan/issues/218


ryan-collingham:
@kelliott55 FYI this fixes the issue with error handling you reported.

PR #221, opened 2019-07-12
Set cmdline defaults from programmatic args

ryan-collingham:
Proof-of-concept:

Instead of defining defaults for programmatic and cmdline args separately and merging them, let the programmatic args be the defaults for the cmdline args, and provide constant default values for the programmatic args. That way, the cmdline args can be used to override the programmatic args.

This way, the --help will (dynamically) display the correct default value based on the programmatic args, and there is no need for any "default value wrapper" utility.


testplan/base.py
```diff
@@ -76,24 +78,170 @@ class Testplan(RunnableManager):
...
+            shuffle = []
+        if extra_deps is None:
+            extra_deps = []
+        if report_tags is None:
+            report_tags = []
+        if report_tags_all is None:
+            report_tags_all = []
+
+        super(Testplan, self).__init__(
+            name=name,
```
Pyifan: can we use the filter_locals utils so that we don't copy them again?

testplan/base.py
```diff
@@ -157,127 +299,14 @@ def run(self):
...
-        test_sorter=None,
-        test_lister=None,
-        verbose=None,
-        debug=None,
-        timeout=None,
-        interactive_handler=TestRunnerIHandler,
-        extra_deps=None,
-        **options
-    ):
+    def main_wrapper(cls, name, **options):
```
Pyifan: If we use **options here, IDE won't prompt when user write @test_plan...

testplan/base.py
ryan-collingham: not really a big fan of using that - explicit is better than implicit ;)

testplan/base.py
ryan-collingham: Yes that is true. However absolutely we mustn't duplicate parameters and their defaults.

I did think about having the full parameter list and defaults on the test_plan decorator and just leaving the Testplan class as **kwargs. However this has the problem that defaults won't be set correctly if the Testplan class is used directly (as we do in a lot of tests). So I think this is the only option to avoid duplication.

Pyifan: Didn't you just say 'explicit is better than implicit' in the previous comments, lol. Our current solution of unfolding **option is duplicating parameters and default from the Config to __init__, so this is unavoidable. Especially the duplication here is going to improve user experience in IDE, thus i think it's worthwhile.

On the contrary, the duplication in the other place - is totally internal, and i think that will be a good place to apply some utility functions to minimize copy & paste.

testplan/base.py
Pyifan: Actually, i'm having a second thought, we cannot really do the filter_locals directly here as it will eliminate pair with None-value. And user might explicitly set a cmdline option to None.

ryan-collingham: OK yeah I'll leave it as is then.

testplan/base.py
ryan-collingham: OK, I'm just worried what will happen when e.g. we add a new param in Testplan class but forget to add it into the decorator, or change default in one but forget to change in another. But maybe you're right that this extra maintenance burden is worth it to improve user IDE experience.

ryan-collingham:
Updated: modified interactive mode logging and docs so that it makes it clear that only API access is given currently. Fix up default abort signals so that SIGINT and SIGTERM are caught correctly.

ryan-collingham:
Sorry I realise this PR is getting a bit too big, maybe should have opened another. However we need the cmdline defaults fix so that the -i flag works at all, so I thought it was just easier to bolt interactive fixes onto this!

PR #224, opened 2019-07-18
Add a http exporter for test report

raoyitao:
* Just like other exporters like JSONExporter, XMLExporter, etc.

### DIFF
--- a/testplan/exporters/testing/base.py
+++ b/testplan/exporters/testing/base.py
@@ -1,9 +1,8 @@
 import os
 from shutil import copyfile
 
-from schema import Schema, Use
+from schema import Schema, Use, Or
 
-from testplan import defaults
 from testplan.common.config import ConfigOption
 from testplan.common.exporters import BaseExporter, ExporterConfig
 from testplan.common.utils.logger import TESTPLAN_LOGGER
@@ -166,6 +165,7 @@ class TagFilteredExporter(Exporter):
             source=source,
             tag_dicts=self.cfg.report_tags_all,
             filter_type=self.ALL)
+
 
 def save_attachments(report, directory):
     """
--- a/testplan/exporters/testing/http/__init__.py
+++ b/testplan/exporters/testing/http/__init__.py
@@ -1,0 +1,72 @@
+"""
+HTTP exporter for uploading test reports via http transmission. The web server
+must be able to handle POST request and receive data in JSON format.
+"""
+import re
+import json
+
+import requests
+from schema import Or, Regex
+
+from testplan.common.config import ConfigOption
+from testplan.common.exporters import ExporterConfig
+from testplan.report.testing.schemas import TestReportSchema
+
+from ..base import Exporter
+
+
+class HTTPExporterConfig(ExporterConfig):
+    """
+    Configuration object for
+    :py:class:`HTTPExporter <testplan.exporters.testing.json.HTTPExporter>`
+    object.
+    """
+    @classmethod
+    def get_options(cls):
+        return {
+            ConfigOption('url', default=None): Or(None,
+                Regex(r'^http[s]?://[\w\d_-]+(:\d{2,5})?$', flags=re.I))
+        }
+
+
+class HTTPExporter(Exporter):
+    """
+    Json Exporter.
+
+    :param url: Http url for posting data.
+    :type url: ``str``
+
+    Also inherits all
+    :py:class:`~testplan.exporters.testing.base.Exporter` options.
+    """
+    CONFIG = HTTPExporterConfig
+
+    def __init__(self, **options):
+        super(HTTPExporter, self).__init__(**options)
+
+    def export(self, source):
+
+        if self.cfg.url is None:
+            raise ValueError('`url` cannot be None.')
+
+        if len(source):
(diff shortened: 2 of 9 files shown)


testplan/exporters/testing/base.py
```diff
@@ -167,6 +167,28 @@ def export(self, source):
             tag_dicts=self.cfg.report_tags_all,
             filter_type=self.ALL)
 
+
+class NetworkExporterConfig(ExporterConfig):
```
ryan-collingham: What's the point of this config class?

testplan/exporters/testing/http/__init__.py
```diff
@@ -0,0 +1,57 @@
...
+            data = test_plan_schema.dump(source).data
+            headers = {
+                'Content-Type': 'application/json; charset=UTF-8',
+                'Accept': 'application/json, text/javascript, */*; q=0.01'
+            }
+            try:
+                response = requests.post(
+                    url=self.cfg.url, data=json.dumps(data), headers=headers)
+                response.raise_for_status()
+            except requests.exceptions.RequestException as exp:
```
ryan-collingham: nice :+1:

testplan/runnable/base.py
```diff
@@ -126,6 +128,9 @@ def get_options(cls):
             ConfigOption(
                 'pdf_style', default=defaults.PDF_STYLE,
                 block_propagation=False): Style,
+            ConfigOption(
+                'http_url', default=None,
+                block_propagation=False): Or(str, None),
```
ryan-collingham: Where does this configuration need to propagate to exactly?

testplan/exporters/testing/base.py
```diff
@@ -167,6 +167,28 @@ def export(self, source):
...
+
+    @classmethod
+    def get_options(cls):
+        return {
+            ConfigOption('url', default=None): Or(str, None)
+        }
+
+
+class NetworkExporter(Exporter):
```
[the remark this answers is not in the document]
raoyitao: I guess we could have other class like DBExporter also inherit this NetworkExporter, then we can have SomeDBExporter inherit DBExporter, they all need a URL in config.

testplan/runnable/base.py
raoyitao: HTTPExporter  works like WebServerExporter, in `get_default_exporters` will check this argument in config, then initialized with url=config.http_url

testplan/exporters/testing/base.py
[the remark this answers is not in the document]
ryan-collingham: A URL is specific to HTTP and a small number of related protocols though, other forms of network transfer will require a different address (e.g. exporting to a MySQL DB will not require a URL). I don't think this is the right abstraction.

Couch DB is unique in that it uses HTTP - I think that when we add a Couch DB exporter it can just inherit from HTTPExporter.

testplan/runnable/base.py
ryan-collingham: Sure, I remember that WebServerExporter needed to propagate its configuration somewhere too I just couldn't quite remember the details.

testplan/exporters/testing/http/__init__.py
```diff
@@ -0,0 +1,59 @@
+"""
+    HTTP exporter for test reports via http. In the other end web server
+    must be able to handle post request and receive data in JSON format.
```
kelliott55: please improve wording.

testplan/parser.py
```diff
@@ -159,10 +159,17 @@ def generate_parser(self):
...
+            default=None, metavar='URL',
+            help='Web URL for posting report.'
+        )
+
         report_group.add_argument(
             '--report-dir',
-            help='Target directory for tag filtered report output.',
-            default=defaults.REPORT_DIR, metavar='PATH')
+            default=defaults.REPORT_DIR, metavar='PATH',
+            help='Target directory for tag filtered report output.'
```
kelliott55: Why are these lines changing? I cannot see a difference other than ordering. 
Please do not change more than is actually necessary in a PR review, lets keep PR's focused on the change at hand.

tests/functional/exporters/testing/test_json.py
```diff
@@ -60,8 +58,8 @@ def test_json_exporter(tmpdir):
 
 def test_implicit_exporter_initialization(tmpdir):
     """
-        An implicit JSON should be generated if `json_path` is available
-        via cmdline args but no exporters were declared programmatically.
+    An implicit JSON should be generated if `json_path` is available
+    via cmdline args but no exporters were declared programmatically.
     """
     json_path = tmpdir.mkdir('reports').join('report.json').strpath
```
kelliott55: Please make these changes in a separate review if they are not functionally part of the http exporter

testplan/exporters/testing/base.py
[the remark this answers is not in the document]
Pyifan: I agree with Ryan. Let's keep the class hierarchy simply until we really need to branch.

tests/functional/exporters/testing/test_json.py
raoyitao: It's a non-functional small change so I modified it by the way. maybe next we can collect these thing and made a PR including all of them

PR #229, opened 2019-07-23
base.Graph PDF renderer to display result.graph assertions in PDF reports

Mariam-Abbas-MS:
This pull request is looking to merge the following additions:
1. A GraphRenderer for base.Graph
2. Methods to convert input data for react-vis graphs to a static MatPlot image
3. Adjustment of base.Graph to allow for rendering of DiscreteCharts


doc/en/design.rst
```diff
@@ -36,11 +36,13 @@ default values or user input with \*\*options after performing a validation stag
 Config objects can inherit other Config objects and alter or define additional
 ConfigOption items:
 
-  * Config instances may have container references that are used to retrieve a
```
Pyifan: This is not intended change, you may need to rebase to latest

ryan-collingham: This should not be changed?

setup.py
```diff
@@ -17,7 +17,7 @@
     'psutil',
     'six',
     'future',
-    'schema<0.7.0',
```
ryan-collingham: This also should not be changed

testplan/exporters/testing/pdf/renderers/entries/baseUtils.py
```diff
@@ -0,0 +1,161 @@
...
+Creates MatPlot plot for any graph requiring 
+axis that can use get_xy_coords function
+"""
+def plot_graph(source, graph_type):
+    data = source['graph_data']
+    graph_options = source['graph_options']
+    series_options = source['series_options']
+
+    # Special logic for multi-bar graph
+    if graph_type is 'Bar':
```
ryan-collingham: Use `==` rather than `is` for string comparison. `is` checks that the objects themselves are the same which can sometimes produce surprising results: https://stackoverflow.com/questions/1504717/why-does-comparing-strings-using-either-or-is-sometimes-produce-a-differe

testplan/exporters/testing/pdf/renderers/entries/baseUtils.py
```diff
@@ -0,0 +1,161 @@
...
+        fig, ax = plot.subplots()
+        num_of_subplots = len(data)
+        width = 0.7/num_of_subplots
```
ryan-collingham: As a best practice, it's better not to have "magic numbers" in code too much and extract them as NAMED_CONSTANTS.

Obviously you shouldn't go overboard on this, e.g. 1 and 0 should usually just be 1 and 0 and not ONE and ZERO.

testplan/testing/multitest/entries/base.py
```diff
@@ -272,13 +272,18 @@ class Graph(BaseEntry):
...
         self.graph_options = graph_options
-        if graph_type != 'Pie':
-            self.type = 'Graph'
-        else:
-            self.type = 'DiscreteChart'
+        self.type = 'Graph'
+        if graph_type in valid_chart_types:
+            self.discrete_chart = True
+        elif graph_type in valid_graph_types:
+            self.discrete_chart = False
```
ryan-collingham: What happens if the graph_type is not in `valid_chart_types` or `valid_graph_types`? I think you should probably raise a `ValueError` since we can only plot these types.

testplan/web_ui/testing/src/AssertionPane/Assertion.js
```diff
@@ -87,8 +87,7 @@ class Assertion extends Component {
       DictMatch: DictMatchAssertion,
       FixLog: FixLogAssertion,
       FixMatch: FixMatchAssertion,
-      Graph: XYGraphAssertion,
-      DiscreteChart: DiscreteChartAssertion
+      Graph: this.props.assertion.discrete_chart? DiscreteChartAssertion: XYGraphAssertion
```
ryan-collingham: Nice :+1:

testplan/testing/multitest/entries/base.py
```diff
@@ -285,5 +285,7 @@ def __init__(self, graph_type, graph_data, description=None,
             self.discrete_chart = True
         elif graph_type in valid_graph_types:
             self.discrete_chart = False
+        else:
+            self.type = 'UnimplementedGraph - ' + graph_type
```
ryan-collingham: So what happens here when we try to render such an UnimplementedGraph in the UI or in the PDF exporter?

Mariam-Abbas-MS: In the UI:
The default for an 'type' that is not yet implemented will be to display a message saying that this assertion has not yet been implemented and to contact the developers if you want it to be.
(This gets handled in Assertion.js, and this logic gets applied to ever type not implemented, so i just needed to make it an type not recognized in Assertion.js)

In the PDF:
Here nothing except the description gets displayed, as by the default again of an un-recognized/unimplemented type in base.py in exporters

ryan-collingham: I think I would prefer to raise a ValueError here instead of just accepting it and not rendering. I'm thinking that it's quite likely a user could mistype the graph name ("Scattre" etc.) and then be confused why their graph does not get displayed. Raising an exception straight away makes it much more obvious.

PR #230, opened 2019-07-25
An example change of unfolding **options for public API classes

Pyifan:
- List the config options of the class (excl. those inherited from parents) in __init__ method
- Move documentation from config class to entity class
- Note that __init__ param of None value will be filtered out before passing to parent class __init__ for building options, so they will take the default value defined in ConfigOption
- Some format correction for docstring

### DIFF
--- a/testplan/runners/pools/remote.py
+++ b/testplan/runners/pools/remote.py
@@ -51,15 +51,6 @@ class RemoteWorkerConfig(ProcessWorkerConfig):
     """
     Configuration object for
     :py:class:`~testplan.runners.pools.remote.RemoteWorker` resource entity.
-
-    :param workers: Number of remote workers of remote pool of child worker.
-    :type workers: ``int``
-    :param pool_type: Remote pool type that child worker will use.
-    :type pool_type: ``str``
-
-    Also inherits all
-    :py:class:`~testplan.runners.pools.process.ProcessWorkerConfig`
-    options.
     """
 
     @classmethod
@@ -89,6 +80,17 @@ class RemoteWorker(ProcessWorker):
     """
     Remote worker resource that pulls tasks from the transport provided,
     executes them in a local pool of workers and sends back task results.
+
+    :param workers: Number of remote workers of remote pool of child worker.
+    :type workers: ``int``
+    :param pool_type: Remote pool type that child worker will use.
+    :type pool_type: ``str``
+    :param remote_host: Remote hostname to connect to.
+    :type remote_host: ``str``
+
+    Also inherits all
+    :py:class:`~testplan.runners.pools.process.ProcessWorkerConfig`
+    options.
     """
 
     CONFIG = RemoteWorkerConfig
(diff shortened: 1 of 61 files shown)


testplan/runners/pools/remote.py
```diff
@@ -577,19 +579,70 @@ class RemotePoolConfig(PoolConfig):
...
+    default_workspace_root = workspace_root()
+
+    @classmethod
+    def get_options(cls):
+        """
+        Schema for options validation and assignment of default values.
+        """
+        return {
+            'hosts': dict,
+            ConfigOption('abort_signals', default=[signal.SIGINT,
```
ryan-collingham: I'm not sure if we need to use ConfigOption at all if we define the default value in the class constructor?

Doesn't need changing for this PR but something to bear in mind in future, I think we can potentially reduce our use of ConfigOption quite a bit.

tests/functional/testplan/test_timeout.py
```diff
@@ -62,3 +62,4 @@ def test_runner_timeout():
         assert current_proc.children() == start_procs
     finally:
         os.remove(output_json)
+        pass
```
ryan-collingham: Why is this needed...?

tests/unit/testplan/common/config/test_generic_config.py
```diff
@@ -157,3 +158,95 @@ def test_getattr_propagation():
...
+
+    @property
+    def options(self):
+        return self._options
+
+
+def test_filter_locals():
+    """Test that Entity.filter_locals() works correctly."""
+    bottom1 = Bottom('Bottom1', baz='you', bar='bye', boo=None)
+    # Arguments defined explicitly in __init__() should appear and for mutable
```
ryan-collingham: So what's the expected behaviour for mutable types? If I append 4 to bottom1.cfg.boo does that mean if I create a new Bottom instance its boo will default to [1, 2, 3, 4] or just [1, 2, 3]? I think the former would be very confusing.

tests/unit/testplan/common/config/test_generic_config.py
```diff
@@ -157,3 +158,95 @@ def test_getattr_propagation():
...
+    assert bottom1.options['baz'] == 'you'
+    assert bottom1.options['bar'] == 'bye'
+    assert bottom1.options['foo'] == 9
+
+    # takes the value of immediate container
+    assert bottom1.cfg.boo == [1, 2, 3]
+
+    bottom2 = Bottom(
+        'Bottom2', description='An example',
+        boo=None, zoo={10: 'a', 20: 'b', 30: 'c'}, bar='barbar'
```
ryan-collingham: So what does bottom2.cfg.boo get set to - None or [1, 2, 3]...?

testplan/runners/pools/remote.py
Pyifan: i think this is for differentiating optional and mandatory options

tests/unit/testplan/common/config/test_generic_config.py
Pyifan: hmmm, maybe the comment is just misleading. For init param that has a None default, they are going to take the default value defined in its ConfigOption. And this is the recommended usage for options that has mutable type of default value.

tests/unit/testplan/common/config/test_generic_config.py
Pyifan: [1, 2, 3] from immediate container

tests/functional/testplan/test_timeout.py
Pyifan: probably added for debugging purpose, removed.

testplan/runners/pools/remote.py
ryan-collingham: We don't need that differentiation when we are unfolding **options and providing a default value in the constructor.

PR #232, opened 2019-07-31
Summary testcase rendering in the UI - @testcase(summarize=true...)

Mariam-Abbas-MS:
This pull request is looking to merge the following additions:
1. Display Testcases with 'summarize=true' (summarized test cases) in the UI
2. Extend the documentation on summary assertions
3. Update documentation for @tescase docstring


doc/en/assertions.rst
```diff
@@ -696,6 +696,46 @@ however this can be overridden by explicitly passing ``category`` argument while
               result.equal(i * 2, i * 2, category='Multiples')
 
 
+This schema highlights the structure of a summarised output
+
+    .. code-block:: bash
```
ryan-collingham: Pretty minor but can you change to `code-block:: none`? This disables syntax highlighting, since the code isn't actually bash you don't want to highlight keywords like `for`, the `|` symbol etc.

doc/en/assertions.rst
```diff
@@ -696,6 +696,46 @@ however this can be overridden by explicitly passing ``category`` argument while
...
+
+        Testplan Summary
```
ryan-collingham: Nice diagram :+1:

testplan/testing/multitest/entries/base.py
```diff
@@ -272,9 +272,15 @@ class Graph(BaseEntry):
...
     def __init__(self, graph_type, graph_data, description=None,
                  series_options=None, graph_options=None):
-        valid_graph_types = ['Line', 'Scatter', 'Bar', 'Whisker',
+        """
+        NOTE:
+        When adding functionality to Graph, VALID_GRAPH_TYPES,
+        VALID_CHART_TYPES, VALID_GRAPH_OPTIONS and
+        VALID_SERIES_OPTIONS must be kept updated
+        """
+        VALID_GRAPH_TYPES = ['Line', 'Scatter', 'Bar', 'Whisker',
```
ryan-collingham: Nice, it's better to store conceptual constants as ALL_CAPS (python doesn't have real constants, so marking them clearly by convention is useful)

Only comment is that maybe these constants could be placed on the class itself (i.e. same indentation as the `def __init__(...):` and accessed through `self.VALID_CHART_TYPES` etc. That's quite a common pattern for constants that belong to a particular class.

testplan/testing/multitest/suite.py
```diff
@@ -447,17 +447,74 @@ def testcase(*args, **kwargs):
...
+    :param num_failing:The number of failing assertions reported per
+                        category per assertion type
+    :type num_failing: ``int`
+    :param key_combs_limit: Max number of failed key combinations on fix/dict
+                            summaries.
+    :type key_combs_limit: ``int``
+    :param execution_group: group of test cases to run in parallel with,
+                            (groups overall are executed serially)
+    :type execution_group: ``str``
+    :param timeout: time elapsed until TimeoutException raised
```
ryan-collingham: Can you mention that the value is given in seconds e.g. "time elapsed in seconds". It's a little annoying when timer-related functions don't make it clear if they take seconds, ms, ns etc.

testplan/web_ui/testing/src/AssertionPane/AssertionSummary.js
```diff
@@ -0,0 +1,104 @@
...
+import PropTypes from 'prop-types';
+import {Col, Row} from 'reactstrap';
+import AssertionGroup from './AssertionGroup';
+/**
+ * Component used to render a large test case when summarize = true
+ */
+
+class SummaryBaseAssertion extends Component  {
+    displayAssertionGroup(assertion_entries){
+        let group_assertion_jsx = [];
```
ryan-collingham: As an exercise you might want to think how you could transform this loop-based approach to a more functional approach using assertion_entries.map(...). Not required, just if you are curious :)

Mariam-Abbas-MS: When I was implementing this i looked up the best practice for looping through arrays, and i was surprised that this was the recommended way (and i was slightly sad :[ ) i originally did the 'for(in)' approach but found out it wasn't recommended (python has converted me) D: i think almost every implementation looks cleaner than this :')

https://stackoverflow.com/questions/6208964/how-to-loop-over-an-array-with-javascript

ryan-collingham: I think the answer to this SO is more up-to-date for modern JS: https://stackoverflow.com/questions/3010840/loop-through-an-array-in-javascript

In summary: `for ... in` is bad. Arrays have a `.forEach()` method you can use for iteration, though `.map()` is the recommended way to build a new array from an existing one (which is what you are doing here).

testplan/web_ui/testing/src/Nav/NavList.js
```diff
@@ -55,7 +55,7 @@ class NavList extends Component {
       navButtons.push(
         <ListGroupItem
           tabIndex={tabIndex.toString()}
-          key={entry.uid}
+          key={entry.uid+entry.name}
```
ryan-collingham: Why is this needed? Do some entries have the same UID...?

testplan/web_ui/testing/src/AssertionPane/AssertionSummary.js
```diff
@@ -7,83 +7,74 @@ import AssertionGroup from './AssertionGroup';
...
-            //Display the Assertion type as text e.g AssertionType: Equal
-            summary_jsx.push(
-                               <Row>
-                                 <Col sm={{ offset: 1 }}>
-                                     {assertion_type.description}
-                                 </Col>
-                               </Row>
-                              )
+    // Go through categories e.g Category: DEFAULT
+    return summaries.map(category =>{
```
ryan-collingham: nice :+1:

ryan-collingham:
Looks like you just need to fix up the JS unit tests before this can be merged, I think maybe the snapshots just need updating.

testplan/web_ui/testing/src/Nav/NavList.js
Mariam-Abbas-MS: actually, they should have unique uids, i though in some cases they might not but it was actually just because of an error i made with the fakeReport for testing, so i will revert this and fix the fake report

PR #234, opened 2019-08-02
A small fix for App driver

raoyitao:
* On App aborting, try closing file descriptors.
* On App stoping, `self.proc` could be None due to starting failure.

### DIFF
--- a/testplan/testing/multitest/driver/app.py
+++ b/testplan/testing/multitest/driver/app.py
@@ -245,7 +245,7 @@ class App(Driver):
         except Exception as exc:
             warnings.warn('On killing driver {} process - {}'.format(
                 self.cfg.name, exc))
-            self._retcode = self.proc.poll()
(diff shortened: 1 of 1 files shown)


testplan/testing/multitest/driver/app.py
```diff
@@ -245,7 +245,7 @@ def stopping(self):
         except Exception as exc:
             warnings.warn('On killing driver {} process - {}'.format(
                 self.cfg.name, exc))
-            self._retcode = self.proc.poll()
+            self._retcode = self.proc.poll() if self.proc else 0
```
ryan-collingham: Not sure we should set the return code to 0 if self.proc is None, doesn't that indicate some error has occured? Return code of 0 says that everything is OK...

In general I think we need to think about the start/stop management of Resource a bit more, stop() should not run unless start() has completed without error.

PR #235, opened 2019-08-05
Update the URL router of web ui

yuxuan-ms:
* Add /api/v1 prefix for web ui reports api


ryan-collingham:
Nice, I need to update https://github.com/Morgan-Stanley/testplan/pull/238 now

PR #236, opened 2019-08-05
Fix HTTP exporter test for python 3.4

ryan-collingham:
Need to explicitly decode bytes to unicode before parsing
JSON from a string in python versions between 3.0-3.6.
Enhance test to store parsed JSON in-memory instead of
writing to a temp file and use a shared HTTP server fixture.

### DIFF
--- a/tests/functional/exporters/testing/test_http.py
+++ b/tests/functional/exporters/testing/test_http.py
@@ -1,121 +1,109 @@
-import os
 import json
 import threading
-import tempfile
 
+import pytest
 from six.moves.BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
 
-from testplan.testing.multitest import MultiTest, testsuite, testcase
-
+from testplan.testing import multitest
 from testplan import Testplan
-from testplan.common.utils.testing import (
-    log_propagation_disabled, argv_overridden
-)
+from testplan.common.utils.testing import argv_overridden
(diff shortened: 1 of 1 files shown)


tests/functional/exporters/testing/test_http.py
```diff
@@ -1,121 +1,109 @@
...
 import json
 import threading
-import tempfile
 
+import pytest
 from six.moves.BaseHTTPServer import HTTPServer, BaseHTTPRequestHandler
 
-from testplan.testing.multitest import MultiTest, testsuite, testcase
-
+from testplan.testing import multitest
```
raoyitao: This change looks fine to me.
Only one question why not directly import MultiTest, testsuite & testcase? of course it makes no difference, then in test_json.py and test_xml.py we 'd better have the same style. These test files are similar in structure.

ryan-collingham: Importing testsuite/testcase at top levels was causing pytest to think that they were also pytest testcases (it just looks for any callable starting with "test"), it's not a big deal but was causing the output to be a bit confusing since pytest claimed to run more testcases than I expected it to. Importing the multitest module avoids that issue. I didn't bother to change all of the other imports, I agree consistency is better though.

In general, it is considered better practice to import whole modules rather than individual functions/classes/objects (e.g. the Google python style guide recommends it: http://google.github.io/styleguide/pyguide.html#22-imports)

PR #237, opened 2019-08-05
Custom comparator UI rendering for Regex conditions

Mariam-Abbas-MS:
This PR is looking to:
1) Display the user set condition (lambda functions/comparison.(...) ) for Regex assertions when pattern matching


testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
```diff
@@ -278,7 +278,25 @@ function prepareBasicContent(assertion) {
       'RegexFindIter'
     ].indexOf(assertion.type) >= 0
   ) {
-    content['leftContent'] = <span>{assertion.pattern}</span>;
+    content['leftContent'] = (
+        <div>
+            <span>{assertion.pattern}</span>
+            <span>
+                {
+                    assertion.condition?
```
ryan-collingham: the CONDITION ? VALUE : OTHER_VALUE construct is OK for a one-liner but here it is a little confusing to follow, could you replace with an if/else branch?

Mariam-Abbas-MS: I don't think you're able to do if/else statements within JSX, so I used logical operators, which ended up being cleaner anyway :D (yay for less lines of code)

https://reactjs.org/docs/conditional-rendering.html

ryan-collingham: OK yeah using `&&` like that is OK. It's also possible to assign a variable using if/else outside of the JSX e.g.

```
let conditionContent;
if (assertion.condition) {
    conditionContent = <...>;
} else {
    conditionContent = null;
}

...
// Then reference {conditionContent} within JSX
```

It's a little more verbose this way, just worth bearing in mind if you e.g. wanted to render one of two completely different things conditionally you can break it into variables like this. As it is I think the `&&` approach is neat and a little more concise.

PR #238, opened 2019-08-05
Implement basic result.attach() method

ryan-collingham:
- Implement result.attach() method for attaching files to the report.
- Pipe the attachment from the Result object into the report and up to the top-level Report container.
- Implement a basic UI component to render a download link for the attachment.


testplan/report/testing/base.py
```diff
@@ -297,17 +303,20 @@ def bubble_up_attachments(self):
...
-
-            # Add logic to find result.attach assertions.
+                self._bubble_up_fix_spec(child)
+            for attachment in child.attachments:
+                self.attachments[attachment.dst_path] = attachment.source_path
+
+    def _bubble_up_fix_spec(self, child):
+        """Bubble up a "fix_spec_path" from a child report."""
+        real_path = child.fix_spec_path
+        hash_dir = hashlib.md5(real_path.encode('utf-8')).hexdigest()
```
Mariam-Abbas-MS: Can you clarify what this code is doing?
What I understand at the moment is that the child.fix_spec_path is location of the TestGroupReport?
And the hash_path is where the UI will locate the attachment?

testplan/report/testing/base.py
```diff
@@ -450,6 +459,12 @@ def merge(self, report, strict=True):
         super(TestGroupReport, self).merge(report, strict=strict)
         self.propagate_tag_indices()
 
+    @property
+    def attachments(self):
+        """Return all attachments from child reports."""
+        return itertools.chain.from_iterable(
```
Mariam-Abbas-MS: > itertools.chain.from_iterable 

is this returning a array[arrays[attachements]] for  all the TestCaseReports below it?
Why did you choose to implement it with itertools? :D

testplan/testing/multitest/base.py
```diff
@@ -606,6 +606,7 @@ def _run_case_related(method):
 
         # native assertion objects -> dict form
         testcase_report.extend(case_result.serialized_entries)
+        testcase_report.attachments.extend(case_result.attachments)
```
Mariam-Abbas-MS: I'm not too sure what this is doing, is this converting from multi test attachment -> TestCaseReport attachments?

testplan/testing/multitest/result.py
```diff
@@ -1935,3 +1943,13 @@ def __iter__(self):
...
+
+
+class Attachment(object):
+    """Represents a file attached to the result object."""
+
+    def __init__(self, filepath):
+        self.source_path = filepath
+        self.uuid = str(uuid.uuid4())
+        self.filename = os.path.basename(filepath)
+        self.dst_path = os.path.join(self.uuid, self.filename)
```
Mariam-Abbas-MS: why make the file path uuid as opposed to a some standard /attachments/tmp/FILENAME

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
```diff
@@ -153,6 +153,49 @@ function prepareRegexMatchLineContent(assertion) {
...
+  // When running the development server, the real Testplan back-end is not
+  // running so we can't GET the attachment. Stick in a button that
+  // gives a debug message instead of the real link.
+  if ((paths.length >= 2) && (paths[1] === '_dev')) {
+    downloadLink = (
+      <button onClick={() => alert("Would download: " + assertion.dst_path)}>
+        {assertion.filename}
+      </button>
+    );
+  } else if (paths.length >= 3) {
```
Mariam-Abbas-MS: Does the url when hosted on a server change as you view the testplan reports? I think in the dev mode hosted with npm, the url never changes from localhost:XXXX/_dev no matter where you browse in the UI. (or at least from what ive seen)

testplan/report/testing/base.py
ryan-collingham: To be honest I'm not 100% sure, I mostly just took the existing code and moved it into a sub-function. I _think_ that the fix_spec_path is a path to a file that is specific to the FIX protocol. FIX is a protocol used in trading, it can have multiple encodings and custom tag values so I think the spec file has something to do with that.

testplan/report/testing/base.py
ryan-collingham: This code is flattening the attachment lists from each child report object into one single list. So for example if our report object has 3 children that each have 3 attachments, it will return a single list containing all 9 attachment objects.

ryan-collingham: I used itertools.chain just because it's a part of the standard library that can be used to flatten multiple iterables into one: https://docs.python.org/3.7/library/itertools.html#itertools.chain.from_iterable

testplan/testing/multitest/base.py
ryan-collingham: In general, this section of code is converting from a `Result` object into a `Report` object, they are subtly different. Here we are copying the list of attachments from the result object onto the report object.

testplan/testing/multitest/result.py
ryan-collingham: I used a UUID to avoid clashes, different testcases might want to attach files with the same name, which would cause problems if we only used the filename as the key.

There are other possibilities, maybe I could have created a directory hierarchy that matches the testcase hierarchy (e.g. attachments/MultiTestName/TestSuiteName/TestCaseName/filename.txt), I thought that would be more complicated though. Maybe it would be better to hash the file to generate a UID instead of using uuid, then attaching the same file multiple times could be done more efficiently.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
ryan-collingham: yes, normally when you view a testplan report the URL is like `localhost:XXXX/testplan/[report UID]`, which we can inspect to extract the report UID. Running the development server and accessing ` localhost:XXXX/_dev` is a special case, which is why I have the branch above to check for that and render a button instead of a link.

ryan-collingham:
TODO: need to update UI code to handle /api/v1 prefix

testplan/testing/multitest/result.py
Mariam-Abbas-MS: does hashing files that have the same problem as using a static file_path, different files with the same name would then clash?

ryan-collingham: I mean hashing the contents of the files, not the names. Though maybe that would be too slow for very large attachments...

ryan-collingham: Now that I think about it, I'm not sure it's worth it. There could still (potentially) be hash collisions for different files, UUID is guaranteed to be unique. I don't think it would be a common requirement to attach the same file from multiple testcases (if it ever is, we can revisit).

PR #239, opened 2019-08-07
Make pre_testcase & post_testcase work with generated testcase

raoyitao:
* @pre_testcase and @post_testcase decorators for test suites cannot
  work properly with testcase parametrization feature, fix it and let
  it support generated testcases with pre-defined parameters.
* Add a test for that.


tests/functional/testplan/testing/multitest/test_suite_decorators.py
```diff
@@ -7,22 +7,33 @@
...
 
-def post(name, self, env, result):
+def post1(name, self, env, result):
     result.equal(2, 2)
     result.contain('case', name)
 
 
-@pre_testcase(pre)
-@post_testcase(post)
+def pre2(name, self, env, result, a=None, b=None):
```
ryan-collingham: Is my understanding correct, that `a` and `b` will be set when the pre/post testcase functions are run before the parametrized testcase, but will be None for other non-parametrized functions in the suite?

raoyitao: Yes, that's right. If the suite only has parametrized testcases based on "case4", then we don't need that default value for 'a' and 'b', but the suite also has a normal "case5".

ryan-collingham: Makes sense, thanks.

PR #240, opened 2019-08-07
UI Cosmetic changes

Mariam-Abbas-MS:
This PR looks to:
- Improve pie chart cosmetics
- Assign graph colors from set colour palette prior to being random
- Relocate axis titles in a better location


testplan/web_ui/testing/src/AssertionPane/AssertionTypes/GraphAssertions/DiscreteChartAssertion.js
```diff
@@ -3,49 +3,66 @@ import PropTypes from 'prop-types';
...
       return (
            <div>
               {plots}
+              <DiscreteColorLegend
+                orientation='horizontal'
+                width={750}
+                items={legend}
+              />
+              <br/>
+              <p>{"(Hover over chart to see labels)"}</p>
```
ryan-collingham: Not sure you need {} for strings, is there any difference to just having ` <p>(Hover over chart to see labels)</p`?

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/GraphAssertions/graphUtils.js
```diff
@@ -27,24 +27,49 @@ export function returnXType(graph_type){
...
+ * (tinted darker/blue)
  */
-export function returnColour(options){
-    if(options == null){
-        let colour = '';
-        for (let i = 0; i < 4; i++) {
-         colour += (Math.round(Math.random() * 15)).toString(16);
+export function returnColour(series_options, data){
+    let series_colours = {};
+    let colour_options = COLOUR_PALETTE.slice(0);
```
ryan-collingham: I think you can just do `array.slice()`, 0 is the default

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/GraphAssertions/graphUtils.js
```diff
@@ -27,24 +27,49 @@ export function returnXType(graph_type){
...
+
+    for(let series_key in data){
```
ryan-collingham: What exactly is `data` here? If it is an array you should use data.forEach() instead.

Mariam-Abbas-MS: `data` is an object containing the { series_name: data_array[] }, it is here in order to assign a colour to each of the series, it is needed to know every series the user created, (they don't have to create series_options for a series) so data is the only structure that has every series name.

I should make this more clear and should use `Object.keys(data)` and then iteratre through all the series_name like this with .forEach(), i think this is better and will do this :]

ryan-collingham: OK great. I've seen some areas of the code use JSDoc notation to document parameter/return types for functions, maybe that would be useful to add in some places where the types are not obvious.

https://devhints.io/jsdoc

ryan-collingham:
FYI there are merge conflicts preventing this PR from being merged. Let me know if you have any problem resolving them.

PR #241, opened 2019-08-07
Refactor basicAssertionUtils

ryan-collingham:
Refactor if/else spaghetti into switch cases so that the
basic assertions are easier to extend and maintain.

### DIFF
--- a/testplan/web_ui/testing/src/AssertionPane/AssertionTypes/BasicAssertion.js
+++ b/testplan/web_ui/testing/src/AssertionPane/AssertionTypes/BasicAssertion.js
@@ -18,7 +18,7 @@ import {prepareBasicContent} from './basicAssertionUtils';
  * |~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~|
  * | preContent    | rightContent  |
  * |~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~|
- * | postContent                   |
+ * | postTitle                     |
  * |~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~|
  * | postContent                   |
  * |_______________________________|
--- a/testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
+++ b/testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
@@ -4,6 +4,451 @@ import {hashCode} from '../../Common/utils';
 /** @module basicAssertionUtils */
 
 /**
+ * Content required to render a basic assertion. The content is rendered by the
+ * BasicAssertion component, refer to that component for a diagram of how each
+ * content section is displayed.
+ *
+ * @typedef {Object} AssertionContent
+ * @property {object|string|null} preTitle - Content above the assertion title
+ * @property {object|string|null} preContent - Content between the preTitle and
+ *                                             the title
+ * @property {object|string|null} leftTitle - Left side of the title
+ * @property {object|string|null} rightTitle - Right side of the title
+ * @property {object|string|null} leftContent - Left side of the main content
+ * @property {object|string|null} rightContent - Right side of the main content
+ * @property {object|string|null} postTitle - Content immediately below the
+ *                                            title
+ * @property {object|string|null} postContent - Final content of assertion
+ */
+
+/**
+ * Prepare the content for the Log assertion.
+ * @param {object} assertion
+ * @param {AssertionContent} defaultContent
+ * @return {AssertionContent} Content for Log assertion
+ * @private
+ */
+function prepareLogContent(assertion, defaultContent) {
+  const preContent = (
+    <span>
+      {assertion.message !== undefined ? assertion.message : null}
+     </span>
+  );
+
+  return {
+    ...defaultContent,
+    preContent: preContent,
+    leftTitle: null,
+    rightTitle: null,
+  };
+}
+
+/**
+ * Prepare the content for the Equal assertion.
+ * @param {object} assertion
+ * @param {AssertionContent} defaultContent
+ * @return {AssertionContent} Content for Equal assertion
+ * @private
+ */
+function prepareEqualContent(assertion, defaultContent) {
+  const leftContent = <span>{assertion.second}</span>;
+
+  return {
+    ...defaultContent,
+    leftContent: leftContent,
+  };
+}
+
+/**
+ * Prepare the content for the NotEqual assertion.
+ * @param {object} assertion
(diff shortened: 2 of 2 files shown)


testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
```diff
@@ -3,6 +3,448 @@ import {hashCode} from '../../Common/utils';
...
+ * Content required to render a basic assertion.
+ * @typedef {Object} AssertionContent
+ * @property {object|string|null} preTitle - Content above the assertion title
+ * @property {object|string|null} preContent - Content between the preTitle and
+ *                                             the title
+ * @property {object|string|null} leftTitle - Left side of the title
+ * @property {object|string|null} rightTitle - Right side of the title
+ * @property {object|string|null} leftContent - Left side of the main content
+ * @property {object|string|null} rightContent - Right side of the main content
+ * @property {object|string|null} postTitle - Content immediately below the
```
Mariam-Abbas-MS: postTitle and postContent dont get used at all, I guess it's not a problem to keep them in anyway as they are only taking up 4 lines of code and could be used in the future - just a note

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
```diff
@@ -3,6 +3,448 @@ import {hashCode} from '../../Common/utils';
...
+ * @private
+ */
+function prepareLogContent(assertion, defaultContent) {
+  const preContent = (
+    <span>
+      {assertion.message !== undefined ? assertion.message : null}
+     </span>
+  );
+
+  return {
```
Mariam-Abbas-MS: I like this system of adjusting the default content with `{...defaultContent }` :D much cleaner

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
```diff
@@ -224,176 +586,84 @@ function prepareBasicContent(assertion) {
     rightContent: assertion.first,
     postTitle: null,
     postContent: null,
-  };
+  }
 
-  var uid = hashCode(JSON.stringify(assertion));
+  // Fan out to the relevant function to prepare content for each assertion
+  // type.
+  switch (assertion.type) {
```
Mariam-Abbas-MS: Yay! its much nicer

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
```diff
@@ -3,6 +3,448 @@ import {hashCode} from '../../Common/utils';
...
+ * Prepare the content for the Equal assertion.
+ * @param {object} assertion
+ * @param {AssertionContent} defaultContent
+ * @return {AssertionContent} Content for Equal assertion
+ * @private
+ */
+function prepareEqualContent(assertion, defaultContent) {
+  const leftContent = <span>{assertion.second}</span>;
+
+  return {
```
Mariam-Abbas-MS: Maybe these smaller functions can be placed directly inside the switch case? Maybe this looks cleaner though  ¯\_(ツ)_/¯

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/BasicAssertion.js
```diff
@@ -18,7 +18,7 @@ import {prepareBasicContent} from './basicAssertionUtils';
  * |~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~|
```
Mariam-Abbas-MS: This diagram could also be placed in the basicAssertionsUtils.js, but maybe it's redundant to have it twice

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
ryan-collingham: Good point, I did think that there were perhaps a few too many options here...

I think removing those content sections would be a bigger change than a simple refactor, and require updating the BasicAssertion module too to not render those sections. Maybe a change we can make in future, though as you note it's not exactly doing much harm to leave them in...

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
ryan-collingham: Thanks, I find the object-spread syntax is really useful in React - particularly when you want to update just a small part of the component's state.

ryan-collingham: I also think that it generally fits into the React "design philosophy" to create an entirely new object instead of mutating the original object here.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
ryan-collingham: Thanks! Using switch to "fan out" an incoming object into subtypes is quite a common pattern, I've seen it used before in the context of receiving a message and switching based on the message type to decide how to handle it. I think it's a lot more readable than if/else (sadly Python has no switch statement - you can simulate them using a dict of functions though).

I remember reading about a design principle a while ago that you should generally split functions into three categories - 1) "deciding" functions that branch based on some condition, 2) "Pure" functions that just calculate a value based on some input with no side effects, and 3) "Business logic" functions that perform some well-defined side effect (e.g. writing to a file). Here I've tried to only use 1) (for the main "switch" function) and 2) (for the prepareContent helpers).

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
ryan-collingham: It wouldn't be unreasonable to do that, I thought consistency was best here though. Especially as the switch statement has a _lot_ of cases (and is only likely to increase if we add more assertions), I wanted to reduce its length as much as possible.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/BasicAssertion.js
ryan-collingham: I agree that having the diagram near the typedef would be useful! I think I might move it, no point in duplicating it.

PR #242, opened 2019-08-07
Simplify fakeReport attachments test

ryan-collingham:
Remove extra test suite and just have a single testcase, to match the other entries in the fakeReport.


ryan-collingham:
> I found the navigation when there are nested suits confusing on first use, was this the same for you or is this similar to the previous UI. Or maybe it's me :D

The hierarchy is normally Testplan -> MultiTest -> Suite -> Testcase. Previously I accidentally inserted an extra Suite under the SampleSuite, which the UI can handle fine but now it's simpler.

We redesigned the navigation of the UI slightly from the old version, previously you would have three columns displaying the MultiTest / Suite / Testcase. Whereas now just the one column, takes up less screen space for the navigation. The bar at the top lists the current position in the hierarchy. If you have suggestions for improvement let me know :)

PR #245, opened 2019-08-15
An extra kill_process_psutil utility

raoyitao:
* Add a function that terminate a `psutil.Process` instance and all of
  its child processes.
* The original kill_process function also use `psutil` to kill child
  processes, but we need to catch NoSuchProcess exception because after
  a process is terminated some related processes may also terminate.

### DIFF
--- a/testplan/common/utils/process.py
+++ b/testplan/common/utils/process.py
@@ -29,7 +29,6 @@ def kill_process(proc, timeout=5, signal_=None, output=None):
     If alive, kills the process.
     First call ``terminate()`` or pass ``signal_`` if specified
     to terminate for up to time specified in timeout parameter.
-
     If process hangs then call ``kill()``.
 
     :param proc: process to kill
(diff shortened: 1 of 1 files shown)


testplan/common/utils/process.py
```diff
@@ -84,6 +87,57 @@ def kill_process(proc, timeout=5, signal_=None, output=None):
     return proc.returncode
 
 
+def kill_process_ex(proc, timeout=5, signal_=None, output=None):
```
ryan-collingham: As discussed, please can we change the name of this function to something like `kill_process_psutil()` - to highlight the difference is that it takes a psutil object.

testplan/common/utils/process.py
```diff
@@ -84,6 +87,57 @@ def kill_process(proc, timeout=5, signal_=None, output=None):
...
+    :param timeout: timeout in seconds, defaults to 5 seconds
+    :type timeout: ``int``
+    :param output: Optional file like object for writing logs.
+    :type output: ``file``
+    :return: List of processes which are still alive
+    :rtype: ``list`` or ``psutil.Process``
+    """
+    _log = functools.partial(_log_proc, output=output)
+
+    procs = proc.children(recursive=True) + [proc]
```
ryan-collingham: I know that this is the same logic as used in `kill_process`. It doesn't need changing for this PR, but in future I think we should change the high-level logic so that it does:

1. Send SIGTERM to parent process __only__. Normally, the parent process should handle terminating its children.
2. Wait for all processes to exit.
3. Kill any remaining processes with SIGKILL

Probably we could share some of the waiting / kill remaining logic with `kill_process`.

raoyitao: Not sure how the parent process deal with child, for CouchDB, if the parent process receives a TERM signal, then all of its children will also be terminated, that sounds reasonable. So I agree with this logic, if a parent process does not handle its children on receiving TERM or INT, it means the code is not correct, then we can arbitrarily kill all of its children. I'll update the PR in this way.

ryan-collingham: Yes exactly, a well-behaved process should terminate its children before it exits, we just need to kill them as a backup. On the other hand, a process may not expect its children to terminate  and may e.g. try to restart them if it detects that, which maybe could cause problems.

ryan-collingham:
Looks great! Thanks.

PR #246, opened 2019-08-16
Enhance and fix interactive reload feature

ryan-collingham:
Enhances interactive reload feature to build up a graph of dependencies,
ensuring that modules are reloaded in the correct order and that modules
are reloaded when their dependencies are modified. Add a unit test to
test this behaviour. Fixes and re-enables the interactive mode functional
test for python 3.

### DIFF
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -480,9 +480,9 @@ class TestRunnerIHandler(RunnableIHandler):
         """Add an environment from the created environment maker instance."""
         self.target.add_environment(self._created_environments[env_uid])
 
-    def reload(self, rebuild_dependencies=True):
+    def reload(self, rebuild_dependencies=False):
         """Reload test suites."""
         tests = (self.test(test, runner_uid=runner_uid)
                  for test, runner_uid in self.all_tests())
-        self._reloader.reload(tests, rebuild_dependencies=rebuild_dependencies)
+        self._reloader.reload(tests, rebuild_dependencies)
 
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -13,6 +13,7 @@ from testplan.report.testing import TestReport
 from testplan.runnable.interactive.http import TestRunnerHTTPHandler
 from testplan.runners.base import Executor
 from testplan.runnable.interactive.reloader import ModuleReloader
+from testplan.runnable.interactive.resource_loader import ResourceLoader
 
 
 class TestRunnerIHandlerConfig(RunnableIHandlerConfig):
@@ -82,10 +83,7 @@ class TestRunnerIHandler(RunnableIHandler):
     def __init__(self, **options):
         super(TestRunnerIHandler, self).__init__(**options)
         self._created_environments = {}
-        self._reloader = ModuleReloader(logger=self.logger,
-                                        extra_deps=self.cfg.extra_deps)
-
-        from testplan.runnable.interactive.resource_loader import ResourceLoader
+        self._reloader = ModuleReloader(extra_deps=self.cfg.extra_deps)
         self._resource_loader = ResourceLoader()
 
     def _execute_operations(self, generator):
--- a/testplan/runnable/interactive/reloader.py
+++ b/testplan/runnable/interactive/reloader.py
@@ -164,7 +164,7 @@ class ModuleReloader(logger.Loggable):
 
     def _suites_by_class(self, tests):
         """
-        Creates a {class --> [suite_instance, ...]} mapping.
+        Creates a {module_name: {class_name: [suite_instance, ...]}} mapping.
 
         :param tests: iterator of Tests.
         :type tests: ``Iterable[Test]``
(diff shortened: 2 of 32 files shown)


tests/functional/testplan/runnable/interactive/interactive_executable.py
```diff
@@ -26,54 +32,59 @@
      'description': 'Assertion', 'utc_time':  re.compile('\d+\.?\d*')}
 
 
-# CREATE AN INTERACTIVE TESTPLAN
-plan = Testplan(name='MyPlan', interactive=True, interactive_block=False)
+def main():
```
Pyifan: Was the original file meant to be executed in Jupyter Notebook?

testplan/runnable/interactive/reloader.py
```diff
@@ -1,219 +1,530 @@
...
-                suite_obj.__class__ = attr
-                suite.set_testsuite_testcases(suite_obj)
+        :param extra_deps: Modules to register as extra dependencies to reload,
+            despite not being directly imported by __main__.
+        :type extra_deps: ``Iterable[ModuleType]``
+        """
+        main_module_file = sys.modules['__main__'].__file__
+        if not main_module_file:
+            raise RuntimeError(
+                "Can only use interactive reloader when the __main__ module "
```
Pyifan: I saw originally _get_reload_dirs() allows absence of __main__ module, then we just reload from pwd() - why are we adding this constraint?

tests/functional/testplan/runnable/interactive/test_interactive.py
```diff
@@ -535,7 +542,6 @@ def add_second_client_after_environment_started():
                     'server': 'STOPPED'})
 
 
-@pytest.mark.skipif(six.PY3, reason='Reload test is unstable on python 3')
```
Pyifan: Do we know why it was unstable, and how this PR fixes it?

testplan/runnable/interactive/base.py
```diff
@@ -482,9 +480,9 @@ def add_created_environment(self, env_uid):
         """Add an environment from the created environment maker instance."""
         self.target.add_environment(self._created_environments[env_uid])
 
-    def reload(self, rebuild_dependencies=True):
+    def reload(self, rebuild_dependencies=False):
```
Pyifan: Is dependencies built the first time reload is called? If we set rebuild_dependencies default False, when are we going to build dependencies?

testplan/runnable/interactive/reloader.py
```diff
@@ -1,219 +1,530 @@
...
-                try:
-                    mod_filepath = self._module_filepath(
-                        inspect.getfile(sys.modules[name]))
-                    if any(mod_filepath.startswith(directory)
-                           for directory in self.reload_dirs):
-                        self._dependencies[filepath].add(mod_filepath)
-                except:
-                    pass
-
-        # Manipulating dependencies so that
```
Pyifan: Do we no longer need this piece of logic？

testplan/runnable/interactive/reloader.py
ryan-collingham: There needs to be a main module to be the root of the dependency graph. It would significantly complicate the logic otherwise, since the dependency graph would not have a single root. The only use-case where there wouldn't be a __main__ module is if a user is importing testplan from the command-line interpreter and defining a testplan there dynamically (instead of just running your normal testplan script with -i) - I'm not sure if this is really something people would want to do.

We can always support no __main__ as a future enhancement, but for now I think this is simplest.

The old OSS reloader didn't have this restriction because it doesn't build a graph of dependencies, it just finds all modules in sys.modules that are locally and reloads them in arbitrary order when modified without considering which module depends on which other.

tests/functional/testplan/runnable/interactive/test_interactive.py
ryan-collingham: It was unstable because it made some assumptions that are only valid on Python 2. It works with Python 3 fine now.

testplan/runnable/interactive/base.py
ryan-collingham: Dependencies are built when the reloader object is constructed. Generally you don't want to rebuild dependencies every time, because for a large project that could take a while (~seconds) especially on an NFS.

ryan-collingham: The only time you should need to reload dependencies is if you add (or delete) an entire module, which is not the common use-case where someone just tweaks a few existing testcases.

testplan/runnable/interactive/reloader.py
ryan-collingham: No, this is the old logic for building dependencies. It is fundamentally incorrect - modulefinder here will find all dependencies of a given script including indirect ones, whereas we only want to build a graph of direct dependencies. Also some modules cannot be run as a script by modulefinder if e.g. it is part of a package and uses relative imports, so we will miss those modules completely from the dependency graph.

By subclassing modulefinder and override the import_hook, we can detect only the direct dependencies and build a graph.

PR #247, opened 2019-08-19
Simplify and clean up WebUI url

yuxuan-ms:
* Simplify /static url
* Use static_folder feature
* Clean up unused url

### DIFF
--- a/testplan/web_ui/testing/package.json
+++ b/testplan/web_ui/testing/package.json
@@ -37,5 +37,5 @@
     "lint": "eslint --ext .js src",
     "eject": "react-scripts eject"
   },
-  "homepage": "/testplan/static"
+  "homepage": "/"
 }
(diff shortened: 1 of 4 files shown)


testplan/web_ui/testing/src/Report/BatchReport.js
```diff
@@ -55,8 +55,8 @@ class BatchReport extends Component {
     if ((paths.length >= 2) && (paths[1] === '_dev')) {
       const r = propagateIndices([fakeReportAssertions]);
       setTimeout(() => {this.setState({report: r, loading: false});}, 1500);
-    } else if (paths.length >= 3) {
-      const uid = paths[2];
+    } else if (paths.length >= 2 && paths[1].length > 0) {
+      const uid = paths[paths.length-1];
```
ryan-collingham: Why does this need to be changed? The URL is still /testplan/[UID] no?

yuxuan-ms: for short url like /[UID]

ryan-collingham: I thought we agreed not to have that? Otherwise how do you distinguish between a resource name like /testplan and a UID?

PR #249, opened 2019-08-19
Matplot UI rendering and attachment content displayed

Mariam-Abbas-MS:
This PR looks to:

- Allow MatPlot graphs to be displayed as static images in the UI, and be downloaded (using base.Attachment logic)

- Modify attachment rendering in the UI based on file type:        

      1.  Image files are rendered as static downloadable images with captions
      2.  .txt files are rendered as previews displaying the last 20 lines, with the ability to 'expand' to 
           view the full file with line numbers - also with the ability to download


testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TextAttachment.js
```diff
@@ -0,0 +1,270 @@
...
+    * @public
+    */
+    getTextAttachment() {
+    const paths = window.location.pathname.split('/');
+
+    // Check if dev mode and display test_string
+    if ((paths.length >= 2) && (paths[1] === '_dev')) {
+        const TEST_TEXT = "test1\ntest2\ntest3\ntest4\ntest5\ntest6\ntest7\n";
+        let text = TEST_TEXT;
+        let lines = text.split('\n');
```
ryan-collingham: It's a bit sad that we are duplicating logic between the dev mode and regular branches. I know it's not very straightforward because the regular mode is async. But maybe you can refactor into an extra helper function that takes the text string and handles setting the state from that - could call it something like `handleText(text)`. Then you can call that same helper function from the axios callback with the text data and also with the special dev test data.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TextAttachment.js
```diff
@@ -0,0 +1,270 @@
...
+        let length = lines.length;
+
+        if(lines.pop() === ""){
+            text = text + "<newline>"
+            lines[length-1] = "<newline>"
+        }
+
+        this.setState({
```
ryan-collingham: Can we merge these two setState calls into one? I think that each setState call can potentially trigger a separate re-render, so it is more efficient to just call it once.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TextAttachment.js
```diff
@@ -0,0 +1,270 @@
...
+            let lines = text.split('\n');
+            let length = lines.length
+
+            if(lines.pop() === ""){
+                text = text + "<newline>"
+                lines[length-1] = "<newline>"
+             }
+
+            let display = this.getCollapsedText(lines);
+            this.setState({
```
ryan-collingham: Strictly I don't think that you should rely on `this.state` when calling `this.setState` - from the React docs, setState updates may be asynchonous (i.e `this.state` may not be immediately updated after calling `this.setState`). I think the effect of this would be that if someone clicked the expand/collapse button _really_ quickly, not all clicks would register correctly (and possible lead to some weird inconsistent state).

https://reactjs.org/docs/state-and-lifecycle.html#state-updates-may-be-asynchronous

The docs suggest a form where you pass a callback that takes state and props as args - maybe we could use that form instead? (JS loves callbacks after all)

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
```diff
@@ -541,35 +543,126 @@ function prepareDictCheckContent(assertion, defaultContent) {
...
-    );
+  const uid = paths[2];
+  let attachment_content;
+  let file_type = assertion.orig_filename.split('.').pop()
+  let file_path = assertion.dst_path;
+  let description = assertion.description;
+  let get_path = `/api/v1/reports/${uid}/attachments/${file_path}`
+
+  switch(file_type){
+    case 'txt':
```
ryan-collingham: Nice use of switch statement. Could potentially add some more text file extensions here - .log, .out and .err are reasonably common.

examples/Data Science/test/test_plan.py
```diff
@@ -0,0 +1,167 @@
+#!/usr/bin/env python
```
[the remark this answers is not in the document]
Mariam-Abbas-MS: nope, whoops! this was used to help me test the matplot function but i forgot to delete it

examples/Data Science/basic_models/test_plan.py
```diff
@@ -45,6 +46,21 @@ def create_image_plot(title, img_data, rows, columns, index):
 
 @testsuite
 class ModelExamplesSuite(object):
+    def __init__(self):
```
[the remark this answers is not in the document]
Mariam-Abbas-MS: Yes, whoops! Again, i added these files by mistake, was just to help me test out the adjusted methods all in one file

testplan/testing/multitest/entries/base.py
```diff
@@ -227,13 +227,18 @@ def __init__(self, message, description=None):
...
         self.width = float(width)
         self.height = float(height)
         self.image_file_path = image_file_path
-        pyplot.savefig(image_file_path, dpi=dpi, pad_inches=0, transparent=True)
-        pyplot.close()
+
+        # Filter unneeded information from base.Attachment for image
+        RELEVENT_KEYS = ['source_path', 'dst_path', 'orig_filename', 'hash',
+                         'filesize', 'file_path', 'description']
+        modified_attachment = {key: attachment.__dict__[key]
```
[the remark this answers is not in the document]
Mariam-Abbas-MS: Yeah, I think whilst the filtering removed unnecessary attributes, it's probably not worth the maintenance for the few lines of json saved.

Ahh I looked up `getattr()` vs` __dict__`, and I saw that `getattr()` works in more cases than ` __dict__`  :D I guess you should only use ` __dict__` if you really need the dictionary object

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TextAttachment.js
```diff
@@ -0,0 +1,270 @@
...
+
+    /**
+    * Fetch the text attachment.
+    *   * Handle UID errors.
+    *   * Make a GET request for the .txt attachment file.
+    *   * Prepare the display content
+    * @public
+    */
+    getTextAttachment() {
+    const paths = window.location.pathname.split('/');
```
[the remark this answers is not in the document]
Mariam-Abbas-MS: everything is now all squished and correct :[

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TextAttachment.js
Mariam-Abbas-MS: I saw there was quite a bit of code repetition, but I didn't think it was enough to refactor out, but turns out as there is logic repeated 3 times throughout the code, i'm now also convinced I should factor it out :D plus when it's right next to each other like that it looks even more horrible

https://stackoverflow.com/questions/2298272/how-much-duplicated-code-do-you-tolerate

testplan/testing/multitest/entries/base.py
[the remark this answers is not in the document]
ryan-collingham: Yeah, essentially `__dict__` only stores the _instance_ attributes - anything set on the object after it is created, including in the `__init__` constructor. It doesn't include _class_ attributes - the class' methods or any variables created in "class scope" (same indentation level as method `def`s). `__dict__` also doesn't include any values that could be returned if the magic `__getattr__` or `__getattribute__` methods are overridden. `getattr` on the other hand works for any of these.

As a general rule, directly accessing "dunder" methods or attributes from outside the object is best avoided, most of them are considered "private by convention" in the same way as any other attribute starting with one or more underscores is. Most magic methods have some non-magic function, operator or statement that can be used to access them - `__dict__` has `getattr`, `__eq__` has the `==` operator, `__enter__` and `__exit__` have the `with` statement, etc.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TextAttachment.js
ryan-collingham: It is a question of judgement, sometimes a small amount of repetition can be OK if it helps readability - but this is the exception rather than the norm. In this case I think having a well-named function is more readable. A good rule of thumb is to think about what the duplicated lines are doing and if you can think of a good function name to describe what they do - if so that suggests you should factor them out into a function. A good function should mostly do one thing, and its name should tell you what that thing is.

In some cases the lines might do two (or more) distinct things, in which case you might consider factoring into two (or more) functions.

How far to take DRY is to a certain extent language dependent. In more verbose languages like C/C++, the "overhead" for creating a new function is non-trivial because you have to also create a function prototype in some other header file and worry more about collisions in the namespace, so developers tend to be lazier with allowing some repetition of a few lines here and there. In Python or JS the overhead of creating a new function is virtually nil, so there is less excuse for having even a couple of lines of repetition.

examples/Data Science/basic_models/test_plan.py
```diff
@@ -25,6 +25,7 @@
 matplotlib.use('agg')
 import matplotlib.pyplot as plot
 import numpy as np
+import tempfile
```
ryan-collingham: I think this can be removed

testplan/testing/multitest/entries/base.py
```diff
@@ -342,3 +328,17 @@ def __init__(self, filepath, description):
...
         super(Attachment, self).__init__(description=description)
+
+
+class MatPlot(Attachment):
+    """Display a MatPlotLib graph in the report."""
+    def __init__(self, pyplot, image_file_path, width=2, height=2,
+                 description=None):
+        self.width = float(width)
+        self.height = float(height)
+        self.image_file_path = image_file_path
```
ryan-collingham: Nice! Just one thing: the Attachment entry already stores the original filepath via the `source_path` attribute, so I think that also storing it in the `image_file_path` attribute is redundant.

PR #251, opened 2019-08-22
Fix a small issue in `testplan.common.utils.process`

raoyitao:
* In function `kill_process`, the while condition (verify timeout) has a typo.

### DIFF
--- a/testplan/common/utils/process.py
+++ b/testplan/common/utils/process.py
@@ -1,7 +1,6 @@
 """System process utilities module."""
 
 import time
-import signal
 import psutil
 import warnings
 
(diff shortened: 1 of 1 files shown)


testplan/common/utils/process.py
```diff
@@ -56,9 +56,8 @@ def kill_process(proc, timeout=5, signal_=None, output=None):
     begin = time.time()
     intervals = exponential_interval(initial=0.05, multiplier=1.1, maximum=1)
 
-    while time.time() - begin >= timeout and retcode is None:
```
Pyifan: Let's use get_sleeper() utility here.

raoyitao: yes used it instead.

PR #252, opened 2019-08-22
Fix race condition in parallel MTest example

ryan-collingham:
Should not rely on time.sleep for synchronisation - use Barrier and
Event primitives instead. Add a Lock to protect the shared refcounts
counter inside the ExclusiveResourceManager driver.

### DIFF
--- a/examples/Multitest/Parallel/parallel_tasks.py
+++ b/examples/Multitest/Parallel/parallel_tasks.py
@@ -28,8 +28,23 @@ class SampleTest(object):
     """
 
     def __init__(self):
-        self._test_g2_1_done = threading.Event()
+        # A Barrier is a synchronisation primitive which allows a fixed number
+        # of threads (in our case, 2) to wait for each other. We use it here
+        # to demonstrate that testcases are run concurrently and how they may
+        # be synchronised with each other.
+        #
+        # Note that on Python 3 you can use the Barrier class from the standard
+        # library:
+        # https://docs.python.org/3.7/library/threading.html#barrier-objects .
+        # Here we use a backported Barrier provided by Testplan, which works
+        # on both Python 2 and 3.
         self._barrier = thread_utils.Barrier(2)
+
+        # The Event synchronisation primitive allows one thread to signal to
+        # another that is waiting on the first thread to do some work. We use
+        # it here to demonstrate another way testcases within the same
+        # execution group may be synchronised with each other.
+        self._test_g2_1_done = threading.Event()
 
     @testcase(execution_group='first')
     def test_g1_1(self, env, result):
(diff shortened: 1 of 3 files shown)


ryan-collingham:
This isn't a theoretical problem - I saw this test fail in CI due to the poor synchronisation. This change should make the example much more robust.

examples/Multitest/Parallel/parallel_tasks.py
```diff
@@ -19,46 +22,53 @@ class SampleTest(object):
...
     @testcase(execution_group='first')
     def test_g1_1(self, env, result):
-        """Wait for test_g1_2 to also acquire the first resource."""
-        with env.resources['first'] as res:
-            result.true(res.active)
-            while res.refcount < 2:
-                result.log('Waiting for test_g1_2...')
-                time.sleep(1)
-
-    @testcase(execution_group='second')
```
Pyifan: From the look of the original example, testplan logic is to sort the testcases by group before execution. I think this behavior is very implicit, we should not do that. All testcases should be executed in the definition order, then adjacent testcases of the same execution_group can be run in parallel.

ryan-collingham: As discussed, the behaviour to run testcases in the order of definition applies only when running a MultiTest serially. When run in parallel, tests are run in order of their groups, and ordering within a group is undefined since the tests are run in parallel.

I will move back the ordering of testcases and add a comment to the example to describe this behaviour.

ryan-collingham:
> The example wants to demonstrate 2 cases of groups 1 (test_g1_1 & test_g1_2) can be run in parallel, and cases of different group cannot be run in the same time. And I think user is going to lost in this overly advanced implementation.
> 
> A quick thought of mine is: all we need to test is res.refcount has ever reached 2 - meaning 2 cases are using the resource in the same time, so just add a res.max_ref_count and check it?

Parallel MultiTest is an advanced feature, it should not be used unless you understand multithreaded programming reasonably well. I don't think that using a Barrier and Event (both part of python 3 standard library) is that "overly-advanced". In fact I would say it's a lot clearer to understand than the old polling logic (I wrote that originally, so I can say it was pretty bad with hindsight).

Checking the "maxref" would still have race conditions if we don't use ensure synchronisation of threads, since e.g. thread 1 might release the resource before thread 2 acquires it or vice-versa, so the max refcount would only be 1. Using the appropriate synchronisation primitives is the standard way to avoid obscure race conditions.

examples/Multitest/Parallel/parallel_tasks.py
ryan-collingham: In any case, changing the behaviour of parallel MTests is not in scope of this PR and would require more careful thought/design. I'm only trying to fix the example to be less flaky so it doesn't break our CI, as well as to follow best practices for multithreading.

PR #265, opened 2019-09-02
(WIP) Add Windows test environment.

yuxuan-ms:


.travis.yml
```diff
@@ -1,9 +1,23 @@
...
+      os: linux
+      dist: xenial
+      python: 2.7
+    - name: "Python 3.7 on Xenial Linux"
+      os: linux
+      dist: xenial
+      python: 3.7
+    - name: "Python 3.7.4 on Windows"
+      os: windows
+      language: node_js
```
ryan-collingham: Why do we have to set language to node_js and not python?

yuxuan-ms: Because travis doesn't support set python on windows. But we need npm, so I set the language to node_js
https://docs.travis-ci.com/user/languages/python/

ryan-collingham: I see, makes sense. Thanks for adding!

PR #268, opened 2019-09-03
Change the behavior of exporters

raoyitao:
* The original behavior is that if the exporters is programmatically
  specified then arguments from command line (such as --pdf, --json)
  is ignored. Change this behavior that all those exporters can take
  effect and work independently.
* Refactor `HTTPExporter` that code for uploading report is extracted
  as a private function.
* Fix typo.

### DIFF
--- a/testplan/exporters/testing/http/__init__.py
+++ b/testplan/exporters/testing/http/__init__.py
@@ -18,7 +18,7 @@ from ..base import Exporter
 class HTTPExporterConfig(ExporterConfig):
     """
     Configuration object for
-    :py:class:`HTTPExporter <testplan.exporters.testing.json.HTTPExporter>`
+    :py:class:`HTTPExporter <testplan.exporters.testing.http.HTTPExporter>`
     object.
     """
     @classmethod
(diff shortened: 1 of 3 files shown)


testplan/exporters/testing/http/__init__.py
```diff
@@ -18,14 +18,14 @@
...
-    :py:class:`HTTPExporter <testplan.exporters.testing.json.HTTPExporter>`
+    :py:class:`HTTPExporter <testplan.exporters.testing.http.HTTPExporter>`
     object.
     """
     @classmethod
     def get_options(cls):
         return {
-            ConfigOption('url', default=None): Or(None,
-                Regex(r'^http[s]?://[\w\d_-]+(:\d{2,5})?.+$', flags=re.I))
+            ConfigOption('url', default=None):
```
ryan-collingham: I don't think default should be None. I think url is a required param so I don't think we need a default?

ryan-collingham:
Looks good otherwise, thanks for refactoring this.

PR #269, opened 2019-09-04
Fix an issue of config option of exporters

raoyitao:
* Each exporter has a config option (e.g. `json_path`, `http_url`) and
  this option is not repulsive because the exporter can get its value
  from parent(`TestRunner` instance) during runtime. Since it has no
  default value, if not explicitly specified in exporter constructor and
  the parent does not has that attribute, then the option is missing
  in exporter config. Before exporting we have a check to make sure
  the option cannot be `None`, otherwise raise exception, actually it
  will not work, Config.__getattr__() can directly raise exception,
  what we need to do is to make that error message clear.

### DIFF
--- a/testplan/exporters/testing/http/__init__.py
+++ b/testplan/exporters/testing/http/__init__.py
@@ -24,8 +24,8 @@ class HTTPExporterConfig(ExporterConfig):
     @classmethod
     def get_options(cls):
         return {
-            ConfigOption('url', default=None): Or(None,
-                Regex(r'^http[s]?://[\w\d_-]+(:\d{2,5})?.+$', flags=re.I))
+            ConfigOption('url'):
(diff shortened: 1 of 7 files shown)


testplan/exporters/testing/http/__init__.py
```diff
@@ -24,8 +24,8 @@ class HTTPExporterConfig(ExporterConfig):
     @classmethod
     def get_options(cls):
         return {
-            ConfigOption('url', default=None): Or(None,
```
ryan-collingham: Do you need to make similar changes to the Config options for the other exporters?

raoyitao: Other exporters already changed 2 months ago. simply like **ConfigOption('pdf_path'): str**

Issue #258, opened 2019-08-30
testplan ui-generated JSON report for test_plan.view the json report in browser:Cannot view the browser


hemapriya123-sys:
i run test plan using command python test_plan.py --ui.test runs and it prints in the terminal as Json generated at my folder .view the JSON report in browser:http://fe80::cd3e:db0e:1c79:867b%11:62434/testplan/local.but i could not see anything in browser.Please help to see testplan in ui

ryan-collingham:
Looks like that might be a bug, the IP address you give looks like IPv6 so it should be enclosed in square brackets when printed to terminal. Can you try running again and insert square brackets into the URL manually like:

http://[fe80::cd3e:db0e:1c79:867b%11]:62434/testplan/local

And I think you should be able to see the UI.

So that we can fix the bug, could you provide us with more information about:

- Your OS type and version
- The version of Python you have installed (`python --version` output)
- What output do you get from running: `python -c "import socket; print(socket.gethostname())"`

Thanks for your interest in Testplan!

hemapriya123-sys:
Thanks for your immediate response.

I tried manually with square bracket in URL.still the issue is same

information for bug fix:
1.Windows 7 professional
2.python 3.6.4
3.got my PC name 

could you ensure the command which i am giving for ui is correct or not?
If I run in cmd terminal,python test_plan.py --pdf x.pdf. I can view my report in pdf.
for UI,python test_plan.py --ui 
After my testplan runs,it outputs JSON generated at my folder called report.json.I can view it
view the JSON report in the browser:http://fe80::cd3e:db0e:1c79:867b%11:**56554**/testplan/local
but I cant view this url in browser and also every run it generates different number in the place of 56554.
Is there anything i need to give specific with --ui?

hemapriya123-sys:
I forget to mention about the warning i receive,
![testplan](https://user-images.githubusercontent.com/54711670/64037254-75c4c800-cb55-11e9-9ac0-492c0850b496.jpg)

Please help as soon as possible to view tesplan in UI

hemapriya123-sys:
I forget to mention this warning.
![testplan](https://user-images.githubusercontent.com/54711670/64037444-e66be480-cb55-11e9-933c-7f9bbdc31d08.jpg)

please help as soon as possible

ryan-collingham:
Thanks for the information. First of all, don't worry about the warnings about command-line args being overridden, they are expected (we should remove them).

[...]

PR #270, opened 2019-09-05
UI server listens on 0.0.0.0

ryan-collingham:
Fixes https://github.com/Morgan-Stanley/testplan/issues/258

### DIFF
--- a/testplan/exporters/testing/webserver/base.py
+++ b/testplan/exporters/testing/webserver/base.py
@@ -7,8 +7,7 @@ from __future__ import absolute_import
 import ipaddress
 import os
 import json
-
-from schema import Or
+import socket
 
 from testplan import defaults
 from testplan.common.utils.timing import wait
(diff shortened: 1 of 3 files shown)


testplan/exporters/testing/webserver/base.py
```diff
@@ -94,20 +93,40 @@ def export(self, source):
...
-        self.logger.exporter_info(
-            'View the JSON report in the browser: {}'.format(self.url))
+        # Check if we are bound to the special (and default) 0.0.0.0 address -
+        # in that case, the UI can be accessed both from localhost or from
+        # any IP address this machine listens on.
+        if host == "0.0.0.0":
+            local_url = 'http://localhost:{}/testplan/local'.format(port)
+
+            try:
+                local_ip = socket.gethostbyname(socket.getfqdn())
```
ryan-collingham: Maybe I should comment in the code, but here local_ip is guaranteed to be IPv4 so we don't have to bother checking for IPv6. If there is no IPv4 address available then socket.gethostbyname() will raise socket.gaierror - which we catch and then only print the localhost URL.

I used socket.getfqdn() instead of socket.gethostname(), since apparently the latter can cause some issue on Linux if machines have their own hostname in /etc/hosts: https://stackoverflow.com/questions/166506/finding-local-ip-addresses-using-pythons-stdlib

PR #271, opened 2019-09-06
Add a property `exporters` for `TestRunner`

raoyitao:
* A `TestRunner` instance generates exporters only once, and it only
  depends on the constructor arguments. Add a private member to store
  these exporters and it can be accessed from outside as a property.
  Then the exporters know each other through their parent.
* After exporting, the log info should contain the full path of saved
  report rather than a single file name relative to the current path.
* All exporters use self.logger instead of `TESTPLAN_LOGGER`, and use %
  notation in `exporter_info`.

### DIFF
--- a/testplan/exporters/testing/xml/__init__.py
+++ b/testplan/exporters/testing/xml/__init__.py
@@ -279,6 +279,7 @@ class XMLExporter(Exporter):
                     encoding='UTF-8'
                 )
 
-        TESTPLAN_LOGGER.exporter_info(
-            '%s XML files created at: %s', len(source), xml_dir)
+        self.logger.exporter_info(
+            '%s XML files created at %s', len(source), os.path.abspath(xml_dir)
+        )
 
--- a/testplan/exporters/testing/http/__init__.py
+++ b/testplan/exporters/testing/http/__init__.py
@@ -41,17 +41,15 @@ class HTTPExporter(Exporter):
     """
     CONFIG = HTTPExporterConfig
 
-    def _upload_report(self, url, source):
+    def _upload_report(self, url, data):
         """
-        Upload Json report, then return the response from server with an
+        Upload Json data, then return the response from server with an
         error message (if any).
         """
         response = None
         errmsg = ''
 
-        if len(source):
-            test_plan_schema = TestReportSchema(strict=True)
-            data = test_plan_schema.dump(source).data
+        if data and data['entries']:
             headers = {
                 'Content-Type': 'application/json',
                 'Accept': 'application/json, text/javascript, */*; q=0.01'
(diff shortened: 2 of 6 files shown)


testplan/exporters/testing/xml/__init__.py
```diff
@@ -280,5 +280,6 @@ def export(self, source):
                 )
 
         TESTPLAN_LOGGER.exporter_info(
-            '%s XML files created at: %s', len(source), xml_dir)
+            '{} XML files created at {}', len(source), os.path.abspath(xml_dir)
```
ryan-collingham: Replacing %s with {} won't work, you have to use %-style formatting with python logging when the format values are specified as *args since it uses the % operator to do the format: https://docs.python.org/3/library/logging.html#logging.Logger.debug

Generally with logging it is slightly preferable to use % formatting and pass the format options as *args rather than formatting explicitly using either % operator or .format(). This is because the logging module can choose not to perform the string formatting at all if e.g. for a debug log if debug logging is not enabled, as an optimization. Whereas when you format a log message with .format() the string formatting always has to be done, even if the message is then discarded. The real impact of performing a few unnecessary string formats is almost certainly negligible, but it's just a nice detail to be aware of.


In any case, putting {} here without a .format will cause an Exception to be raised, so you need to fix - simplest to just revert back to %s flags.

Logging is just a special case in this regard - outside of logging, remember {} and .format are definitely preferable to % formatting!

raoyitao: I saw **PDFExporter** also use "{}" in **TESTPLAN_LOGGER.exporter_info**, so maybe we also need to change that. Another thing is that **JSONExporter** use **self.logger.exporter_info** instead of **TESTPLAN_LOGGER.exporter_info**, is it better? if we use **self.logger.exporter_info** then we also prefer %s notation?

ryan-collingham: Currently the code uses a mix of {} and % notation for logging in different parts. I personally prefer % in logging for the optimization reasons noted above. For all other string formatting (e.g. exception messages, print() calls) {} formatting is preferred.

It is better to use self.logger if you can, since in the log file that will give information about which class is making the log. TESTPLAN_LOGGER is only meant for code that can't use self.logger (i.e. outside of a class).

ryan-collingham: As you've seen, there is a lot of code that doesn't follow the above guidelines. I don't care strongly enough to go through and fix up everything at once, I've mostly just been fixing up the logging in code that I touch.

raoyitao: that makes sense. so for all exporters I removedd **TESTPLAN_LOGGER**, and use %s notation instead of '{}'.format(...).

By the way, there's another small change in this commit, **HTTPExporter::_upload_report** accepts a json data as argument instead of a TestReport instance, because the MS internal exporter has its own TestReport class and corresponding schema.

PR #273, opened 2019-09-10
Fix "None" runpath issue on Process Pool

ryan-collingham:
Fixes https://github.com/Morgan-Stanley/testplan/issues/274 where the runpath used for Process Pool workers is the string
"None", resulting in a None/ directory being created in the local working
dir. self.runpath is not set until self.make_runpath_dirs() is called from
within self.starting().


ryan-collingham:
Run private tests internally to verify that nothing else pool-related is broken by this change.

PR #277, opened 2019-09-11
Update to react-scripts 3

ryan-collingham:
Fixes security vulnerabilities reported by npm audit and gives us latest
react-scripts features and updates. Update Snapshot tests to work with latest
Jest 24 required by react-scripts 3 by using the enzyme-to-json package (see
https://github.com/facebook/jest/issues/7802).


ryan-collingham:
For now I've pinned react-scripts to 3.0.1 to avoid a linting bug in 3.1.1: https://github.com/facebook/create-react-app/issues/7566

Seems like the fix for that is merged and should be in next release of CRA, so we could bump the version of react-scripts then.

testplan/web_ui/testing/package.json
```diff
@@ -14,30 +14,46 @@
...
+    "production": [
+      ">0.2%",
+      "not dead",
+      "not op_mini all"
+    ],
+    "development": [
+      "last 1 chrome version",
+      "last 1 firefox version",
+      "last 1 safari version"
+    ]
```
morsedav: Why are we opting for these particular versions?

ryan-collingham: These are the default values set by react-scripts 3.

testplan/web_ui/testing/src/AssertionPane/Assertion.js
```diff
@@ -20,9 +20,12 @@ import NotImplementedAssertion from './AssertionTypes/NotImplementedAssertion';
 import AssertionHeader from './AssertionHeader';
 import AssertionGroup from './AssertionGroup';
 import {BASIC_ASSERTION_TYPES} from '../Common/defaults';
-import XYGraphAssertion from './AssertionTypes/GraphAssertions/XYGraphAssertion';
-import DiscreteChartAssertion from './AssertionTypes/GraphAssertions/DiscreteChartAssertion';
+import XYGraphAssertion
+  from './AssertionTypes/GraphAssertions/XYGraphAssertion';
+import DiscreteChartAssertion
+  from './AssertionTypes/GraphAssertions/DiscreteChartAssertion';
```
morsedav: is it feasible to hoist the exports like:

DiscreteChartAssertion.js => GraphAssertions/index.js => AssertionTypes.js 

so we can just do `import DiscreteChartAssertion from './AssertionTypes';` ?

ryan-collingham: I was only adding in line-breaks here to fix new line-length linting rules added by the new react-scripts version.

Hoisting the exports sounds reasonable though.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/basicAssertionUtils.js
```diff
@@ -610,7 +610,7 @@ function getImageContent(image_path, description){
             <div>
               <figure className={css(styles.caption)}>
                   <img src={image_path}
-                       class="img-responsive ..."
+                       className="img-responsive ..."
```
morsedav: why ellipsis?

ryan-collingham:
I've explicitly tested that the web UI can be isntalled, started in dev mode and built in production mode using internal npm artifactory using this branch.

PR #282, opened 2019-09-18
Use react-router for UI routing

ryan-collingham:
Adds the react-router-dom package and use it to route within the UI App
based on the URL. react-router handles parsing the URL and extracting the
UID for a batch report so we do not need no directly inspect the Window
object.

Adds a placeholder "NotImplemented" page for the interactive UI, and a
404 Not Found page for URLs that don't match an expected path.

Note: for now I am going to implement the Interactive mode component
as part of the main UI, the idea is it can be accessed via localhost:xxxx/interactive.
In future it might be better to split out into a completely separate react App,
however that would be a bigger change and require more work on build,
setting up a common shared "library" of components between batch report
App and interactive App. This is definitely doable, but I'm planning to start with
the simpler single App option and iterate from there.


testplan/web_ui/testing/package.json
```diff
@@ -55,5 +57,6 @@
     "snapshotSerializers": [
       "enzyme-to-json/serializer"
     ]
-  }
+  },
+  "proxy": "http://localhost:4000"
```
ryan-collingham: Note: this setting allows us to start the UI in dev mode using `npm start` and have it communicate with the backend of a testplan, assuming the testplan is started in UI mode on port 4000 like `$ ./test_plan.py --ui 4000`. It's very useful as it allows you to iterate quickly on the frontend UI and test how it interacts with the real backend, instead of just loading the static fakeReport.js.

See https://create-react-app.dev/docs/proxying-api-requests-in-development for more detail.

ryan-collingham:
I've updated this PR with some other UI changes I've made, since they depend on the changes already made here. Since we now use React Router to parse URL I've gone and removed the parsing of window.paths from the attachment components, and made some refactor of the component.

    * Move the Attachment components out of basicAssertions and into their
      own Assertion type.
    * Propogate down the report UID into the assertions
      from the top level batch report, instead of parsing the global window
      path.
    * Add tests for new attachment components and for existing TextAttachment
      component.

PR #288, opened 2019-10-01
Add some utils needed for LBM & UME drivers

morsedav:


testplan/common/utils/networking.py
```diff
@@ -8,10 +8,105 @@
...
+    """
+    Finds free TCP ports in the given range by binding to random ports in the range,
+    until <nb_ports> free ports have been found.
+
+    This does not guarantee that by the time caller tries to bind the ports will
+    still be available.
+
+    .. warning::
+
+        You should *not* use this utility. Ever. That is because it is impossible to guarantee
```
ryan-collingham: I don't think we should be porting this utility here if we can avoid it - as the warning says, it is not a reliable way of acquiring free ports.

testplan/common/utils/networking.py
```diff
@@ -8,10 +8,105 @@
...
+            __ip_local_port_range_cache = int(start), int(end)
+    return __ip_local_port_range_cache
+
+
+def available_port():
+    """
+    Finds free TCP port, this is a shortcut to :py:func:`unoccupied_ports`
+    for just one port, in the range of user accessible ports.
+
+    .. warning::
```
ryan-collingham: I don't think we should port this utility for the same reason.

testplan/common/utils/networking.py
```diff
@@ -8,10 +8,105 @@
...
+
+__ip_local_port_range_cache = None
+def ip_local_port_range():
+    """
+    Return local ip port range on the machine
+
+    :return: pair of integer port values
+    :rtype: ``tuple`` of ``int``, ``int``
+    """
+    global __ip_local_port_range_cache
```
ryan-collingham: please try to avoid globals. This function could be replaced with a constant. If lazy evaluation is really needed, there are better ways to do it that don't require use of globals.

Though as per comment below, I'm not sure we even need this utility.

testplan/common/utils/path.py
```diff
@@ -301,3 +302,56 @@ def hash_file(filepath):
...
+    :return: path to the archived file
+    :rtype: ``str``
+    """
+    new_path = path + timestamp
+    if os.path.isfile(path):
+        os.rename(path, new_path)
+    return new_path
+
+
+def copy(source, target, using_subprocess=False):
```
ryan-collingham: Not sure if we really need this utility instead of just using shutil.copy() directly. We certainly don't need the using_subprocess option, as that won't work on Windows.

PR #289, opened 2019-10-03
Implement prototype interactiv navigation

ryan-collingham:
Add components for navigating the report in interactive mode. Each report
entry has a "play" button that can be used to run the test. Since this is
a prototype, we don't yet make any API calls to the backend to run the tests,
instead a fake interactive report is used. For demonstration purposes only,
the tests are assumed to pass after 3 seconds have passed.


testplan/web_ui/testing/src/Nav/navUtils.js
```diff
@@ -119,10 +125,137 @@ function _parseNavSelection(entries, selected, depth, parentUid) {
  * has the entries name & type.
  * @returns {{navBreadcrumbs: Array, navList: Array}}
  */
-function parseNavSelection(entries, selected) {
-  return _parseNavSelection(entries, selected, 0, undefined);
+function ParseNavSelection(entries, selected) {
```
morsedav: why not `parseTopLevelNavSelection`?

ryan-collingham: Why do you think that would be clearer? I was planing to take a closer look at this function at some point in future, it's strange that it is supposed to take an array of just 1 element (the root of the report tree) rather than just the root report node itself.

morsedav: The name _parseTopLevelNavSelection_ is just a name to better distinguish the functionality of  [ParseNavSelection](https://github.com/Morgan-Stanley/testplan/blob/7a173b8af4cdf453f8c8d5fb43631c3b9c4b3de7/testplan/web_ui/testing/src/Nav/navUtils.js#L128) from that of [parseNavSelection](https://github.com/Morgan-Stanley/testplan/blob/7a173b8af4cdf453f8c8d5fb43631c3b9c4b3de7/testplan/web_ui/testing/src/Nav/navUtils.js#L74). Plus it feels icky to have two functions identically named save one capitalized character. It's easy to imagine a dev using one when they meant to use the other > intellisense doesn't tell them anything is wrong > the dev can't figure out why things aren't behaving correctly.

I'd be happy with any name change that better distinguished these two functions from one another.

ryan-collingham: OK yes I agree that having the names differ by one letter capitalisation is bad. I've renamed the recursive function to `parseNavSelectionRecur` so as to keep the extenal interface of the module unchanged.

testplan/web_ui/testing/src/Nav/__tests__/InteractiveNav.test.js
```diff
@@ -20,7 +20,7 @@ describe('InteractiveNav', () => {
   it('shallow renders and matches snapshot', () => {
     const renderedNav = shallow(
       <InteractiveNav
-        report={[FakeInteractiveReport]}
```
morsedav: We don't know why it was being passed as an array?

ryan-collingham: looks like it was a convenience to make the recursive step of parseNavSelection simpler, since it recurses each time into the entries array. But representing the report by a single element is much clearer IMO. Having a report represented by more than one root nodes makes no sense.

PR #293, opened 2019-10-14
Simplify auto-select for batch and interactive reports

ryan-collingham:
Simplify auto-select logic for both batch and interactive reports.
Entries are automatically selected when the report is first loaded only,
not on every re-render.

Lift the selection state up into the report
components to make the navigation components fully-controlled and
stateless.

Add report loading and assertion rendering into interactive report. Add tests.

Some other minor refactor/tidy up.

### DIFF
--- a/testplan/web_ui/testing/src/Common/utils.js
+++ b/testplan/web_ui/testing/src/Common/utils.js
@@ -40,6 +40,17 @@ function getNavEntryDisplayData(entry) {
 }
 
 /**
+ * Convert a report entry into a nav entry type.
+ *
+ * @param {object} reportEntry - entry from report object.
+ * @returns {object}
+ */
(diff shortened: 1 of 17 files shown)


dobragab:
Out of pure curiosity, how can I try the prototyped interactive navigation?

ryan-collingham:
> Out of pure curiosity, how can I try the prototyped interactive navigation?

Hi, the interactive GUI is still in early development, so it doesn't do very much of use and is likely to change significantly. In particular, the frontend doesn't yet communicate with the backend API and instead renders a fake test tree

If you are still curious, you can check out latest master and run:

```
cd <repo_root>/testplan/web_ui/testing/
npm install
npm start
```

Then navigate to localhost:4000/interactive to see the basic mock-up.

Thanks for taking an interest!

PR #302, opened 2019-10-22
Implement REST API for interactive backend

ryan-collingham:
Implements a documented, HTTP REST API for interactive mode. To view the swagger documentation, run any testplan script in interactive mode with -i flag and navigate to localhost:XXXX/ in the browser.

Includes both unit and functional tests of the new API.


testplan/report/testing/base.py
```diff
@@ -66,13 +68,17 @@ class Status(object):
     INCOMPLETE = 'incomplete'
```
Pyifan: What does a 'incomplete' status mean?

testplan/report/testing/base.py
```diff
@@ -66,13 +68,17 @@ class Status(object):
...
     SKIPPED = 'skipped'
+    READY = 'ready'
+    RUNNING = 'running'
 
     STATUS_PRECEDENCE = (
+        RUNNING,
         ERROR,
         FAILED,
         INCOMPLETE,
         PASSED,
```
Pyifan: Not sure about READY having the lowest precedence, so if one testcase PASSES and other entries are not run yet, the entire report will show as PASSED?

testplan/common/report/base.py
```diff
@@ -235,6 +235,41 @@ def get_by_uid(self, uid):
...
+        Unfortunately, for updating an existing entry requires an O(n) lookup
+        in the list of existing entries, since the index is read-only.
+        Inserting a new entry is O(1).
+
+        :param uid: `uid` for the child report.
+        :type uid: ``hashable``
+        :param item: entry to update or insert into the report.
+        :type item: ``Report``
+        """
+        if uid in self._index:
```
Pyifan: Maybe we can save the position of the item in _index instead? In that way we can access the entry as self.entries[self._index[uid]], O(1)

testplan/report/testing/base.py
ryan-collingham: I don't know I didn't add it. It's used somewhere. Probably means a testcase is interrupted or not run for some reason.

testplan/report/testing/base.py
ryan-collingham: READY has the highest precedence.

testplan/common/report/base.py
ryan-collingham: Maybe that would be better yeah. TBH storing entries in a list is a mistake, they should be stored in a dict keyed by UID. But I'm wary to make big changes in this area because I know that most of the rest of testplan depends on it, so I'm just making the minimal changes I can.

testplan/runnable/interactive/http.py
```diff
@@ -49,162 +313,34 @@ class TestRunnerHTTPHandler(Entity):
...
     """
+
     CONFIG = TestRunnerHTTPHandlerConfig
 
     def __init__(self, **options):
         super(TestRunnerHTTPHandler, self).__init__(**options)
-        self._ip = None
-        self._port = None
+        self._server = None
+        self.task_pool = None
```
yuxuan-ms: self.pool? I didn't see self.task_pool be used but self.pool defined in `def run`

tests/functional/testplan/runnable/interactive/test_api.py
```diff
@@ -0,0 +1,288 @@
...
+    assert host == "0.0.0.0"
+
+    report_url = "http://localhost:{}/api/v1/interactive/report".format(port)
+    rsp = requests.get(report_url)
+    assert rsp.status_code == 200
+    report_json = rsp.json()
+
+    # Trigger all tests to run by updating the report status to "running"
+    # and PUTting back the data.
+    report_json["status"] = "running"
```
yuxuan-ms: Status.RUNNING

testplan/runnable/interactive/http.py
```diff
@@ -1,21 +1,284 @@
...
+
+                if should_run(ihandler.report.status):
+                    new_report.status = report.Status.RUNNING
+                    http_handler.pool.apply_async(ihandler.run_tests)
+
+                ihandler.report = new_report
+                return ihandler.report.shallow_serialize()
+
+    @api.route("/api/v1/interactive/report/tests")
+    class Tests(flask_restplus.Resource):
```
yuxuan-ms: It's easy to mistake Test for Tests, Suite for Suites. Could you rename these classes?

testplan/common/report/base.py
ryan-collingham: I guess that self._index is "private" to the Report object so changing it to store index should be a safe change, I'll take a look. Good suggestion!

testplan/report/testing/base.py
ryan-collingham: Now I think about it, maybe READY should be a higher priority than PASSED/FAILED. That way, interactive testplan will only be green when all testcases have passed at least once.

testplan/runnable/interactive/http.py
```diff
@@ -1,21 +1,284 @@
...
+        """
+        Tests endpoint. Represents all Test objects in the report. Read-only.
+        """
+
+        def get(self):
+            """Get the UIDs of all tests defined in the testplan."""
+            with ihandler.report_mutex:
+                return [test.uid for test in ihandler.report]
```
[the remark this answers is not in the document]
ryan-collingham: The lock is required when reading/updating the report object, since otherwise other threads could mutate the report object halfway through when we are iterating through the tests it contains.

testplan/runnable/interactive/http.py
```diff
@@ -24,13 +287,14 @@ class TestRunnerHTTPHandlerConfig(EntityConfig):
...
     @classmethod
     def get_options(cls):
         return {
-            'ihandler' : RunnableIHandler,
-            ConfigOption(
-                'host', default=defaults.WEB_SERVER_HOSTNAME): str,
-            ConfigOption('port', default=defaults.WEB_SERVER_HOSTNAME): int
+            "ihandler": RunnableIHandler,
+            ConfigOption("host", default=defaults.WEB_SERVER_HOSTNAME): str,
+            ConfigOption("port", default=defaults.WEB_SERVER_HOSTNAME): int,
```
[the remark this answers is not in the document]
ryan-collingham: damn, how does this work currently then?!

testplan/runnable/interactive/http.py
ryan-collingham: Woops, yes I changed the name from task_pool -> pool but forgot to update the __init__ declaration.

testplan/runnable/interactive/http.py
ryan-collingham: Hmm maybe. How does `AllTests` and `SingleTest` sound?

yuxuan-ms: > Hmm maybe. How does `AllTests` and `SingleTest` sound?

It sounds good!

testplan/runnable/interactive/http.py
```diff
@@ -24,13 +287,14 @@ class TestRunnerHTTPHandlerConfig(EntityConfig):
...
+            ConfigOption("pool_size", default=4): int,
```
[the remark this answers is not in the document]
ryan-collingham: I think that I may be able to remove the thread pool entirely at some point, I need to look more at how the ihandler queues tasks. For now I agree that it is better to put this magic number in the defaults file.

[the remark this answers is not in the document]
ryan-collingham: I would like to remove the thread pool entirely at some point, but for now it's the simplest solution I think. I would prefer if the InteractiveIHandler class could provide a better interface for queuing async tasks.  For now I agree that it is better to put this magic number in the defaults file.

PR #303, opened 2019-10-23
Remove Pillow from dependencies

ryan-collingham:
Pillow is not used anywhere in the testplan package, so it should not be listed as a dependency in setup.py. For the one example that uses it, note that Pillow needs to be installed manually.


morsedav:
For future reference this CVE motivated this PR: https://nvd.nist.gov/vuln/detail/CVE-2019-16865

PR #304, opened 2019-10-24
Optimize result.log display

yuxuan-ms:
* convert no-string object to pretty string
* use pre instead of span to keep the format of string


testplan/testing/multitest/entries/base.py
```diff
@@ -217,9 +220,12 @@ def _summarize(self, entries, num_passing, num_failing, key_combs_limit):
 
 
 class Log(BaseEntry):
-
+    """Log a str to the report."""
     def __init__(self, message, description=None):
-        self.message = message
+        if isinstance(message, basestring):
```
ryan-collingham: Do you need this check? I think you can just pass a string to pprint.pformat() and it will return the same string unchanged.

yuxuan-ms: > Do you need this check? I think you can just pass a string to pprint.pformat() and it will return the same string unchanged.

pprint.pformat()  wil replace \n to \\n. I want to keep the raw string format.

PR #306, opened 2019-10-25
Serve static interactive files

ryan-collingham:
Serve the testplan web app alongside the interactive API when running
interactive mode. Log how to access the API schema. Since the UI
is still a work in progress, don't yet log details on how to access it.

### DIFF
--- a/testplan/common/utils/networking.py
+++ b/testplan/common/utils/networking.py
@@ -19,7 +19,7 @@ def port_to_int(port):
         return socket.getservbyname(port)
 
 
-def format_access_urls(host, port, url_path):
+def format_access_urls(host, port, url_path, ssl=False):
     """
     Format a string to log to give HTTP access to a URL endpoint listening
     on a particular host/port. Handles special 0.0.0.0 IP and formats IPV6
@@ -29,23 +29,30 @@ def format_access_urls(host, port, url_path):
     :type host: ``str``
     :param port: port number
     :type port: ``int``
-    :param url_path: path to format after host:port part of URL (expected to
-        contain leading "/" e.g "/index.html")
+    :param url_path: path to format after host:port part of URL. A leading /
+        will be inserted if there isn't already one present.
     :type url_path: ``str``
     :return: a formatted string containing the connection URL(s)
     :rtype: ``str``
     """
+    # Strip any whitespace and insert a leading / to paths if required.
+    url_path = url_path.strip()
+    if not url_path.startswith("/"):
+        url_path = "/" + url_path
+
+    scheme = "https" if ssl else "http"
+
     # Handle 0.0.0.0 as a special case: this means that the URL can be accessed
     # both via localhost or on any IP address this host owns.
     if host == "0.0.0.0":
-        local_url = "http://localhost:{port}{path}".format(
-            port=port, path=url_path
+        local_url = "{scheme}://localhost:{port}{path}".format(
+            scheme=scheme, port=port, path=url_path
         )
 
         try:
             local_ip = socket.gethostbyname(socket.getfqdn())
-            network_url = "http://{host}:{port}{path}".format(
-                host=local_ip, port=port, path=url_path
+            network_url = "{scheme}://{host}:{port}{path}".format(
+                scheme=scheme, host=local_ip, port=port, path=url_path
             )
 
             return (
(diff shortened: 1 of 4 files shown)


testplan/common/utils/networking.py
```diff
@@ -8,10 +8,65 @@
...
+    :param port: port number
+    :type port: ``int``
+    :param url_path: path to format after host:port part of URL (expected to
+        contain leading "/" e.g "/index.html")
+    :type url_path: ``str``
+    :return: a formatted string containing the connection URL(s)
+    :rtype: ``str``
+    """
+    # Handle 0.0.0.0 as a special case: this means that the URL can be accessed
+    # both via localhost or on any IP address this host owns.
```
morsedav: Shouldn't this be "this means that the URL can be accessed from any IP address"? As long as a client is able to access the server's listening port (i.e. not blocked by firewall) 0.0.0.0 means any subnet can connect.

testplan/common/utils/networking.py
```diff
@@ -8,10 +8,65 @@
...
+    Format a string to log to give HTTP access to a URL endpoint listening
+    on a particular host/port. Handles special 0.0.0.0 IP and formats IPV6
+    addresses. The returned string is indented by 4 spaces.
+
+    :param host: hostname or IP address
+    :type host: ``str``
+    :param port: port number
+    :type port: ``int``
+    :param url_path: path to format after host:port part of URL (expected to
+        contain leading "/" e.g "/index.html")
```
morsedav: It's good this is mentioned in the docstring but isn't it trivial to just add this check?

```py
path = path.strip()
path = ('/' if not path.startswith('/') else '') + path
```

testplan/common/utils/networking.py
```diff
@@ -8,10 +8,65 @@
...
+        try:
+            local_ip = socket.gethostbyname(socket.getfqdn())
+            network_url = "http://{host}:{port}{path}".format(
+                host=local_ip, port=port, path=url_path
+            )
+
+            return (
+                "    Local: {local}\n"
+                "    On Your Network: {network}".format(
+                    local=local_url, network=network_url
```
morsedav: This is neat. People will like this.

testplan/common/utils/networking.py
```diff
@@ -8,10 +8,65 @@
...
+        # Check for an IPv6 address. Web browsers require IPv6 addresses
+        # to be enclosed in [].
+        try:
+            if ipaddress.ip_address(host).version == 6:
+                host = "[{}]".format(host)
+        except ValueError:
+            # Expected if the host is a host name instead of an IP address.
+            pass
+
+        url = "http://{host}:{port}{path}".format(
```
morsedav: I'd like if this checked the scheme too so in case we ever want to add HTTPS support for whatever reason we wouldn't need to modify this function to return accurate information. Not a big deal though.

testplan/common/utils/networking.py
ryan-collingham: 0.0.0.0 in this context means "all local addresses on the machine". So if a server has multiple physical ethernet links, each with their own IP address, then the HTTP request can be sent to any of those IP addresses. It's not about the IP address of the client.

testplan/common/utils/networking.py
ryan-collingham: yeah it doesn't hurt to add a check

testplan/common/utils/networking.py
ryan-collingham: So do you mean we should make it configurable whether to use http:// or https:// ? I don't mind to add that as a parameter, I think it should default to http for simplicity.

testplan/common/utils/networking.py
```diff
@@ -8,10 +8,65 @@
     absolute_import,
 )
 import socket
+import ipaddress
```
morsedav: Looks like `ipaddress` was only [added in Python 3.3](https://docs.python.org/3.5/library/ipaddress.html?highlight=New%20in%20version%203.3.). Perhaps you could add this [ipaddress](https://pypi.org/project/ipaddress/) as a [conditional dependency](https://stackoverflow.com/a/35614580)?

```txt
ipaddress==1.0.23 ; python_version < '3.3'
```

ryan-collingham: We already have ipaddress as a dependency in setup.py. Good point on making it a conditional dependency though :+1:

testplan/common/utils/networking.py
morsedav: Right, in context makes sense.

testplan/common/utils/networking.py
morsedav: Ya come to think of it there's not really a simple way to detect if we're running over https from here. I'd be happy with adding `is_ssl=False` to the function signature and then just using 'http' / 'https' depending on that value.

PR #307, opened 2019-10-28
Interactive examples modifications

ryan-collingham:
- Modify the basic example to recommend viewing the Swagger UI schema
- Remove other examples from docs temporarily until they are fixed to
  work with the new REST API.


examples/Interactive/Basic/test_plan.py
```diff
@@ -28,26 +28,7 @@ def main(plan):
...
-#     curl -X POST http://127.0.0.1:$PORT/sync/run_tests
-#
-# Run only one suite:
-#     curl -X POST http://127.0.0.1:$PORT/sync/run_test_suite -d '{"test_uid": "Test1", "suite_uid": "BasicSuite"}'
-#
-# .. and all other operations that Testplan interactive provides.
-
+# View the API schema at either the localhost or LAN address printed when this
+# testplan script is run. The schema is interactive so can be used to try out
+# the API. To run a test, set its status to "running" on a PUT update.
```
morsedav: Reads better:

You can browse the API schema at either localhost or at the LAN address that's 
printed when running this testplan script. The API schema is interactive so you can test out
what functionality is available. In order to run a test, issue a PUT request
with the test's status set to "running".

ryan-collingham: Thanks for the suggestion, I've made the change.

PR #309, opened 2019-11-04
Empty testcase report of pyTest means PASS

raoyitao:
* There are new status 'READY' and 'RUNNING' added for test report,
  mainly for interactive mode. If a test case report has no entry then
  the default status is 'READY'. We also support integrating pyTest
  into our report, and adding assertion using `Result` object. If no
  assertion added then there is no entry in the testcase report, but
  actually it has passed because no error raised in pyTest.

### DIFF
--- a/testplan/testing/py_test.py
+++ b/testplan/testing/py_test.py
@@ -315,18 +315,26 @@ class _ReportPlugin(object):
                     'Cannot store testcase results to report: no report '
                     'object was created.')
 
-            # Add the assertion entry to the case report
-            for entry in self._current_result_obj.entries:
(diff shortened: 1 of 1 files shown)


yuxuan-ms:
It would be better if you can add some testcases

Pyifan:
> Looks good to me. I made a similar change for PyUnit already but missed PyTest, thanks for fixing it up.
> 
> Personally, I much prefer to explicitly set the status of an empty testcase to Passed rather than implicitly expecting the report entry to default to Passed.

I don't think it is a decent solution to add such logic to all different types of test that we have. The test's state and their precedence/aggregation needs more thoughts and better design.

PR #310, opened 2019-11-05
Fix 2 small issues of HTTPExporter

raoyitao:
* Use argument `json` instead of `data` when uploading json report.
* Fix a bug in formatting string.

### DIFF
--- a/testplan/exporters/testing/http/__init__.py
+++ b/testplan/exporters/testing/http/__init__.py
@@ -49,24 +49,18 @@ class HTTPExporter(Exporter):
         errmsg = ''
 
         if data and data['entries']:
-            headers = {
-                'Content-Type': 'application/json',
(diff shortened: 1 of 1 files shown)


testplan/exporters/testing/http/__init__.py
```diff
@@ -49,26 +49,20 @@ def _upload_report(self, url, data):
...
-                    timeout=self.cfg.timeout)
+                    url=url, json=data, timeout=self.cfg.timeout)
                 response.raise_for_status()
             except requests.exceptions.RequestException as exp:
                 errmsg = 'Failed to export to {}: {}'.format(url, exp)
         else:
             errmsg = (
                 'Skipping exporting test report via http '
-                'for empty report: %s', data.name)
+                'for empty report: {}'.format(data.name))
```
yuxuan-ms: Is `data` dict type? data.name or data.get('name') ?

PR #316, opened 2019-11-14
Interactive API tweaks

ryan-collingham:
- Return all tests/suites/testcases from their respective API resources, not just a list of UIDs (this is to make polling for updates simpler)
- Remove the HTTP handler's thread pool and instead allow the interactive handler to run tests asynchronously (it already has a thread pool internally, so no point duplicating that).


testplan/runnable/interactive/base.py
```diff
@@ -410,9 +419,11 @@ def all_tests(self, runner_uid=None):
                 for test_uid in runner.added_items:
                     yield test_uid, runner.uid()
 
-    def run_tests(self, runner_uid=None):
+    def run_tests(self, runner_uid=None, await_results=True):
```
morsedav: Can we change the name of this to `run_all_tests`? It takes a 2nd or 3rd look to see that this isn't `run_test`, especially since the signature is almost the same.

PR #317, opened 2019-11-15
Fix navigation bar loses focus highlighter issue

yuxuan-ms:


testplan/web_ui/testing/src/Nav/Nav.js
```diff
@@ -18,6 +18,11 @@ const Nav = (props) => {
     props.selected,
   );
 
+  let selectedUid = undefined;
```
ryan-collingham: slightly prefer `null` to `undefined` here. `undefined` means not defined (usually a mistyped variable name), `null` means defined but empty.

testplan/web_ui/testing/src/Nav/navUtils.js
```diff
@@ -154,12 +158,15 @@ const CreateNavButtons = (props, createEntryComponent) => {
     );
 
     const tabIndex = entryIndex + 1;
-
+    let cssName = [styles.navButton, styles.navButtonInteract];
```
ryan-collingham: prefer `const` to `let` where possible

yuxuan-ms:
> Can you explain more about what issue this is fixing? When does the navigation button lose focus?

The navList entries added :focus pseudo class in CSS for displaying chose navButton(background grey). But if you click on where not in the navList, the :focus will not be enabled. So I add some code on CreateNavButtons function. If the navButton's uid == selected, then added "backgroundColor: DARK_GREY" css style.

ryan-collingham:
> > Can you explain more about what issue this is fixing? When does the navigation button lose focus?
> 
> The navList entries added :focus pseudo class in CSS for displaying chose navButton(background grey). But if you click on where not in the navList, the :focus will not be enabled. So I add some code on CreateNavButtons function. If the navButton's uid == selected, then added "backgroundColor: DARK_GREY" css style.

OK, so the issue is that if you click anywhere outside of the nav bar (e.g. in the assertion pane) the button loses focus? Whereas with this PR it will keep the dark grey highlight. Makes sense.

PR #323, opened 2019-11-22
Add parent_uids to report entries

ryan-collingham:
Adds new parent_uids attribute to all report entries.
This is a list of the UIDs of all parent entries in the report
tree from root (first element) downwards. Knowing both the UID
of an entry and the UIDs of its parents allows an entry to be
found in the report tree in O(log(n)) time.

parent_uids are automatically updated whenever a child report entry
is appended to its parent.

This new attribute is used to find and update report entries in the
interactive UI, instead of relying on the current selection state.


testplan/web_ui/testing/src/Report/__tests__/InteractiveReport.test.js
```diff
@@ -8,6 +8,68 @@ import InteractiveReport from '../InteractiveReport.js';
...
+  "meta": {},
+  "status_override": null,
+  "attachments": {},
+  "tags_index": {},
+  "name": "Assertions Example",
+  "parent_uids": [],
+  "type": "TestGroupReport",
+  "category": "testplan",
+  "entries": [{
+    "uid": "Assertions Test",
```
yuxuan-ms: uid != name. It's not important, but maybe uid="AssertionsTestUID" would be better to avoid mishandling in the test.

PR #325, opened 2019-11-28
Update Fix Drivers to be compatible with the latest Pyfixmsg

raoyitao:
* Pyfixmsg is updated to be compatible with Python 3, the fix drivers
  should also be updated or they can not work in Python 3 environment.

### DIFF
--- a/examples/Transports/FIX/over_one_session.py
+++ b/examples/Transports/FIX/over_one_session.py
@@ -39,7 +39,6 @@ def fixmsg(source):
     to serialise and parse messages and pass the codec explicitly.
     """
     # python 2 and 3 compatibility
-    source = {tag: val.encode('utf-8') for tag, val in source.items()}
     msg = FixMessage(source)
     msg.codec = CODEC
     return msg
--- a/examples/Transports/FIX/over_two_sessions.py
+++ b/examples/Transports/FIX/over_two_sessions.py
@@ -39,7 +39,6 @@ def fixmsg(source):
     to serialise and parse messages and pass the codec explicitly.
     """
     # python 2 and 3 compatibility
-    source = {tag: val.encode('utf-8') for tag, val in source.items()}
     msg = FixMessage(source)
     msg.codec = CODEC
     return msg
--- a/testplan/common/utils/sockets/fix/client.py
+++ b/testplan/common/utils/sockets/fix/client.py
@@ -107,7 +107,7 @@ class Client(object):
 
         msg[34] = self.out_seqno
         self.out_seqno += 1
-        if msg[35] == b'4':
+        if msg[35] in (b'4', u'4'):
             self.out_seqno = int(msg[36])
         return msg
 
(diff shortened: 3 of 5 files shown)


examples/Transports/FIX/over_one_session.py
```diff
@@ -39,7 +39,7 @@ def fixmsg(source):
     to serialise and parse messages and pass the codec explicitly.
     """
     # python 2 and 3 compatibility
-    source = {tag: val.encode('utf-8') for tag, val in source.items()}
+    # source = {tag: val.encode('utf-8') for tag, val in source.items()}
```
ryan-collingham: This must be wrong - source is now not defined?

testplan/common/utils/sockets/fix/client.py
```diff
@@ -107,7 +107,7 @@ def _populate_tags(self, msg):
 
         msg[34] = self.out_seqno
         self.out_seqno += 1
-        if msg[35] == b'4':
+        if msg[35] in (b'4', u'4'):
```
ryan-collingham: If msg values may be either bytes or unicode, that is an ugly interface. Can we fix that? If not, prefer use of `six.ensure_binary()` here.

testplan/testing/multitest/driver/fix/client.py
```diff
@@ -215,7 +215,7 @@ def logoff(self):
         self._client.sendlogoff()
         rcv = self._client.receive(timeout=self.cfg.logoff_timeout)
         self.file_logger.debug('Received logoff response {}.'.format(rcv))
-        if 35 not in rcv or rcv[35] != b'5':
+        if 35 not in rcv or rcv[35] not in (b'5', u'5'):
```
ryan-collingham: same comment here regarding message values being either bytes or unicode, it's not good. Better to ensure they are always either bytes or unicode as makes the most sense (do the values represent text or non-text data?)

examples/Transports/FIX/over_two_sessions.py
```diff
@@ -39,7 +39,7 @@ def fixmsg(source):
     to serialise and parse messages and pass the codec explicitly.
     """
     # python 2 and 3 compatibility
-    source = {tag: val.encode('utf-8') for tag, val in source.items()}
+    #source = {tag: val.encode('utf-8') for tag, val in source.items()}
```
ryan-collingham: Same comment as above - source is now undefined?

examples/Transports/FIX/over_one_session.py
ryan-collingham: Oh I see, source is coming from the function signature. In that case, can we remove this line completely instead of commenting it out?

raoyitao: removed this line

examples/Transports/FIX/over_two_sessions.py
raoyitao: removed this line

testplan/testing/multitest/driver/fix/client.py
raoyitao: we init the fixServer(name='server',  msgclass=FixMessage, codec=CODEC)
Here we use FixMessage from pyfixmsg library which has been updated to be compatible with Py3, so we can directly write **rcv[35] != '5'** here. Although use can pass their customized class as argument and then we can get b'5', u'5' or anything else, but maybe it's their responsibility to let their parser work properly and write py2/3 compatible code, if they really want to use customized class instead of default one. So I just check if value of tag 35 is '5', also changed some similar code in this PR.

PR #329, opened 2019-12-06
Add xfail feature for dealing with tests that cannot succeed

yuxuan-ms:
* Add `@xfail` decorator to mark failed testcase/testsuite
* `@xfail` decorator will add `__xfail__` attribute into testcase/testsuite
* `@xfail` test will be marked xfail(if fail) xpass(if pass)
* Add xfail testcase
* Add `xfail` document

### DIFF
--- a/testplan/common/utils/logger.py
+++ b/testplan/common/utils/logger.py
@@ -15,7 +15,7 @@ import sys
 import logging
 
 from testplan.common.utils.strings import Color
-
+from testplan.report.testing import Status
 
 # Define our log-level constants. We add some extra levels between INFO and
 # WARNING.
@@ -80,15 +80,26 @@ class TestplanLogger(logging.Logger):
         """Log 'msg % args' with severity 'DRIVER_INFO'"""
         self._custom_log(DRIVER_INFO, msg, *args, **kwargs)
 
-    def log_test_status(self, name, passed, indent=0, level=TEST_INFO):
+    def log_test_status(self, name, status, indent=0, level=TEST_INFO):
         """Shortcut to log a pass/fail status for a test."""
-        pass_label = Color.green('Pass') if passed else Color.red('Fail')
+        if status in Status.PASSED_STATUSES:
+            pass_label = Color.green(status)
+        elif status in Status.FAILED_STATUSES:
+            pass_label = Color.red(status)
+        else:
+            pass_label = Color.yellow(status)
+
         indent_str = indent * ' '
         msg = self._TEST_STATUS_FORMAT
         self._custom_log(
             level,
             msg,
-            {'name': name, 'pass_label': pass_label, 'indent': indent_str})
+            {
+                'name': name,
+                'pass_label': pass_label.title(),
+                'indent': indent_str
+            }
+        )
 
     def _custom_log(self, level, msg, *args, **kwargs):
         """Log 'msg % args' with severity 'level'."""
--- a/testplan/report/testing/base.py
+++ b/testplan/report/testing/base.py
@@ -70,6 +70,8 @@ class Status(object):
     SKIPPED = 'skipped'
     READY = 'ready'
     RUNNING = 'running'
+    XFAIL = 'xfail'
+    XPASS = 'xpass'
 
     STATUS_PRECEDENCE = (
         RUNNING,
@@ -79,6 +81,27 @@ class Status(object):
         INCOMPLETE,
         PASSED,
         SKIPPED,
+    )
+
+    STATUS_PROPAGATION = {
+        ERROR: ERROR,
+        FAILED: FAILED,
+        INCOMPLETE: INCOMPLETE,
+        PASSED: PASSED,
+        SKIPPED: SKIPPED,
+        READY: READY,
+        RUNNING: RUNNING,
+        XFAIL: PASSED,
+        XPASS: PASSED,
+        None: None  # `status_override` can be None in merging report,
+    }
+
+    PASSED_STATUSES  = (
+        PASSED, SKIPPED, XFAIL
+    )
+
+    FAILED_STATUSES = (
+        FAILED, ERROR, XPASS
     )
 
     # `status_override` can be None, so need to add it to precedence rules
@@ -95,7 +118,10 @@ class Status(object):
         :param rule: Precedence rules for the given statuses.
         :type rule: ``sequence``
         """
-        return min(stats, key=lambda stat: rule.index(stat))
+        return min(
+            [cls.STATUS_PROPAGATION[stat] for stat in stats],
+            key=lambda stat: rule.index(stat)
+        )
 
 
 _TestCount = collections.namedtuple('_TestCount', Status.STATUS_PRECEDENCE)
@@ -164,6 +190,7 @@ class BaseReportGroup(ReportGroup):
 
     def __init__(self, *args, **kwargs):
         self.meta = kwargs.pop('meta', {})
+        self.status_reason = kwargs.pop('status_reason', None)
         super(BaseReportGroup, self).__init__(*args, **kwargs)
         self.status_override = None
         self.timer = timing.Timer()
@@ -176,7 +203,7 @@ class BaseReportGroup(ReportGroup):
     @property
     def passed(self):
         """Shortcut for getting if report status is `Status.PASSED`."""
-        return self.status == Status.PASSED
+        return self.status in Status.PASSED_STATUSES
 
     @property
     def failed(self):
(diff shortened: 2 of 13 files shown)


testplan/common/report/base.py
```diff
@@ -52,7 +52,8 @@ def __init__(self,
                  description=None,
                  uid=None,
                  entries=None,
-                 parent_uids=None):
+                 parent_uids=None,
+                 xfail=None):
```
ryan-collingham: I don't think it is the right design to have a dedicated xfail attribute in the report. The report already has a status field, we just need to add XFAIL as a new status and add logic into MultiTest in order to set it correctly.

testplan/common/utils/logger.py
```diff
@@ -82,7 +82,12 @@ def driver_info(self, msg, *args, **kwargs):
 
     def log_test_status(self, name, passed, indent=0, level=TEST_INFO):
         """Shortcut to log a pass/fail status for a test."""
-        pass_label = Color.green('Pass') if passed else Color.red('Fail')
+        if passed is True:
+            pass_label = Color.green('Pass')
+        elif passed is False:
+            pass_label = Color.red('Fail')
+        else:
```
ryan-collingham: This is weird - so `passed` is no longer a simple boolean - what else could it be? I think that this method should be changed to accept the status string instead of a `passed` boolean, in order to correctly handle XFAIL and possible future test states (SKIP, XPASS etc.)

testplan/report/testing/schemas.py
```diff
@@ -81,6 +81,8 @@ class TestCaseReportSchema(ReportSchema):
     timer = TimerField(required=True)
     tags = TagField()
 
+    xfail = fields.Dict(allow_none=True)
```
ryan-collingham: We definitely shouldn't be serializing this field - the XFAIL information will already be encoded in the status field. As per above comment, I don't think that a separate xfail attribute should exist in the report at all.

testplan/testing/multitest/base.py
```diff
@@ -547,6 +549,18 @@ def create_testcase_report(testcase):
         if not testsuite_report.failed:
             testsuite_report.status = Status.PASSED
 
+        # If xfailed testsuite, force set status_override and update result
+        if testsuite_report.xfail:
```
ryan-collingham: At this point, instead of checking the report xfail, I think you should instead check the testcase's `__xfail__` attribute - if it is `True` and the testcase report status is `Status.FAIL`, update it to `Status.XFAIL`.

I'm not sure if possible but it would be better if you can directly update the status instead of using the status_override attribute.

testplan/testing/multitest/base.py
```diff
@@ -857,6 +884,9 @@ def aborting(self):
 
     def _log_status(self, report, indent):
         """Log the test status for a report at the given indent level."""
+        passed = report.passed
+        if report.xfail:
+            passed = report.xfail['result'].capitalize()
```
ryan-collingham: this is wrong, a `boolean` variable shouldn't be reassigned to a `str`. As above, just  pass the `report.status.tag` to the log function and let that deal with the different possible statuses, instead off overloading the `passed` variable.

testplan/common/report/base.py
ryan-collingham: Oh OK I think I understand why you added this now. It is not just a simple boolean flag but contains information on the xfail reason and whether it was strict or not?

For the reason, I wonder if this can be done in a more generic way by adding a "reason" attribute to the status object or a `status_reason` attribute to the report object? That way when user marks a testcase as skipped, it can reuse the same reason attribute instead of adding another `skip` attribute (could also be used to give optional reason for marking testcase as ERROR etc).

As for the `strict` param, I'm not sure the report needs to care about that.

testplan/common/utils/logger.py
```diff
@@ -82,7 +82,12 @@ def driver_info(self, msg, *args, **kwargs):
 
     def log_test_status(self, name, passed, indent=0, level=TEST_INFO):
         """Shortcut to log a pass/fail status for a test."""
-        pass_label = Color.green('Pass') if passed else Color.red('Fail')
+        if passed is True:
```
morsedav: Just FYI for anyone else wondering "Shouldn't it be `if passed:` or `if passed == True`?": As of Python 2.3 `True` and `False` are singletons. [link](https://stackoverflow.com/q/27276610)

ryan-collingham: We have this pattern in a number of places. It's ok when there is a good reason you for distinguising between `True` and other truthy values, but the times you actually need to do that should be quite rare (and a reason given as a comment). Normally `if passed:` is more readable and conventional.

testplan/common/report/base.py
ryan-collingham: One more thing to note as well. The current design is that the report base classes under `testplan/common/report` are super generic and know nothing about _test_ reports, which are covered by the `testplan/report/testing`. So if we _were_ to add information to the report that is specific to xfail, it should go into one or more of the classes under `testplan/report/testing` and not the common base classes.

With that said, I personally do not really see much reason currently for splitting the report classes between the generic "common" location and the test-report specific locations, since there are no other report types that inherit from the common base classes - I think it over complicates things for no real gain. So I would be happy to move all of `testplan/common/report` into `testplan/report/testing` and simplify the inheritance hierarchy as it currently very complex.

testplan/report/testing/base.py
```diff
@@ -70,6 +70,9 @@ class Status(object):
     SKIPPED = 'skipped'
     READY = 'ready'
     RUNNING = 'running'
+    XFAIL = 'xfail'
+    XPASS = 'xpass'
+    XPASS_STRICT = 'xpass(strict)'
```
ryan-collingham: What does XPASS_STRICT mean exactly? Personally I think that if a test is marked as strict xfail and it passes, we should just mark as FAIL.

testplan/report/testing/base.py
```diff
@@ -81,6 +84,20 @@ class Status(object):
         SKIPPED,
     )
 
+    STATUS_PROPAGATION = {
```
ryan-collingham: Can you explain what this is doing exactly? I think this is over-complicating the status precedence mechanism.

testplan/report/testing/schemas.py
```diff
@@ -111,6 +113,8 @@ class TestGroupReportSchema(TestCaseReportSchema):
     part = fields.List(fields.Integer, allow_none=True)
     fix_spec_path = fields.String(allow_none=True)
 
+    xfail = fields.Dict(allow_none=True)
```
ryan-collingham: I don't think this is still required?

testplan/common/utils/logger.py
```diff
@@ -80,9 +80,16 @@ def driver_info(self, msg, *args, **kwargs):
         """Log 'msg % args' with severity 'DRIVER_INFO'"""
         self._custom_log(DRIVER_INFO, msg, *args, **kwargs)
 
-    def log_test_status(self, name, passed, indent=0, level=TEST_INFO):
+    def log_test_status(self, name, passed,
```
ryan-collingham: It's a bit odd that this function now takes both a required `passed` boolean and an optional `status` string. Why not just pass the `status` string?

testplan/report/testing/base.py
yuxuan-ms: If a multitest contains a xfail testsuite and the testsuite status is xfail, the multitest status should be passed, not xfail

testplan/report/testing/base.py
yuxuan-ms: > What does XPASS_STRICT mean exactly? Personally I think that if a test is marked as strict xfail and it passes, we should just mark as FAIL.

Many users are using this feature to mark unstable testcase. The testcase not always fail or pass

testplan/report/testing/base.py
ryan-collingham: I don't think I agree. If all testcases in a multitest are XFAIL, it is reasonable to mark the multitest as XFAIL.. So long as PASS has higher precedence than XFAIL, if there are a mixture of PASS and XFAIL testcases the MultiTest will be PASS. I don't think this extra complication is required.

testplan/report/testing/base.py
ryan-collingham: ? I don't think that answers my question. I get what the `strict=True` flag to xpass does, I'm just not sure that we need a separate XPASS_STRICT status.

yuxuan-ms: > ? I don't think that answers my question. I get what the `strict=True` flag to xpass does, I'm just not sure that we need a separate XPASS_STRICT status.

OK, I misunderstand.  Let's mark is as FAIL

testplan/common/utils/logger.py
yuxuan-ms: > It's a bit odd that this function now takes both a required `passed` boolean and an optional `status` string. Why not just pass the `status` string?

How to display the text color? Add a status color map?

ryan-collingham: Yeah maybe, or just a set of PASSED_STATUSES and FAILED_STATUSES. You could use the same sets in the passed / failed property functions.

PR #330, opened 2019-12-06
Add category to all report nodes and simplify UI logic

ryan-collingham:
- Add "category" to all report nodes to identify them as testplan / multitest / suite or testcase nodes, so that they can be labelled correctly. Remove previous attempt to "guess" the correct labels based on the combination of type and category.

- Simplify UI logic by removing separate navigation entry types. All report entries are represented by their node in the report tree. Represent the current selection by a hierarchy of UIDs. Remove assertions from state and look it up in the currently selected entry instead.


ryan-collingham:
This PR fixes the issue of testcases not being labelled correctly in the interactive UI.

tests/functional/exporters/testing/test_pdf.py
```diff
@@ -44,6 +44,7 @@ def test_create_pdf(tmpdir):
                 entries=[
                     TestGroupReport(
                         name='MySuite',
+                        category='suite',
```
morsedav: Same comment / question wrt using an enum instead of strings

PR #331, opened 2019-12-12
Add an information dialog

yuxuan-ms:
Add an information dialog box to contain report meta information, e.g. username, timestamp, command line string etc.

* Add information attribute in TestReport 
* Add information dialog in web UI
* Update pytest/ui testcase

### DIFF
--- a/testplan/report/testing/base.py
+++ b/testplan/report/testing/base.py
@@ -44,7 +44,10 @@ We will have a report tree like:
     ...
 """
 import os
+import sys
 import copy
+import getpass
+import platform
 import collections
 import inspect
 import hashlib
@@ -327,9 +330,26 @@ class TestReport(BaseReportGroup):
     Only contains TestGroupReports as children.
     """
 
-    def __init__(self, meta=None, attachments=None, *args, **kwargs):
+    def __init__(self,
+                 meta=None,
+                 attachments=None,
+                 information=None,
+                 *args,
+                 **kwargs):
         self.meta = meta or {}
         self._tags_index = None
+        self.information = information or []
+        try:
+            user = getpass.getuser()
+        except (ImportError, OSError):
+            # if the USERNAME env variable is unset on Windows, this fails
+            # with ImportError
+            user = 'unknown'
+        self.information.extend([
+            ('user', user),
+            ('command_line_string', ' '.join(sys.argv)),
+            ('python_version', platform.python_version())
+        ])
 
         # Report attachments: Dict[dst: str, src: str].
         # Maps from destination path (relative from attachments root dir)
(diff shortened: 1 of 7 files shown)


testplan/report/testing/base.py
```diff
@@ -352,6 +361,26 @@ def tags_index(self):
                 *[child.tags_index for child in self])
         return self._tags_index
 
+    @property
```
ryan-collingham: It doesn't really matter but is there a particular reason for this to be a property and not just an attribute set on `__init__`?

testplan/report/testing/base.py
```diff
@@ -352,6 +361,26 @@ def tags_index(self):
...
+    def information(self):
+        if not self._info:
+            try:
+                user = getpass.getuser()
+            except (ImportError, OSError):
+                # if the USERNAME env variable is unset on Windows, this fails
+                # with ImportError
+                user = 'unknown'
+            self._info = [
+                ('user', user),
```
ryan-collingham: why not store as a `dict` instead of a list of tuples? Seems marginally simpler.

testplan/report/testing/base.py
```diff
@@ -327,9 +330,15 @@ class TestReport(BaseReportGroup):
...
-    def __init__(self, meta=None, attachments=None, *args, **kwargs):
+    def __init__(self,
+                 meta=None,
+                 attachments=None,
+                 information=None,
+                 *args,
+                 **kwargs):
         self.meta = meta or {}
         self._tags_index = None
+        self._info = information
```
ryan-collingham: Could you please add a comment to explain what is stored in info? Either here or on the property is fine.

testplan/report/testing/base.py
```diff
@@ -352,6 +361,26 @@ def tags_index(self):
                 *[child.tags_index for child in self])
         return self._tags_index
 
+    @property
+    def information(self):
```
ryan-collingham: nitpick: missing docstring here and below

testplan/report/testing/base.py
yuxuan-ms: > why not store as a `dict` instead of a list of tuples? Seems marginally simpler.

For the order of items. If it's a dict/OrderedDict, it would be a object in JSON. I want to keep the same order on the UI side

ryan-collingham: why does the order matter?

testplan/report/testing/base.py
yuxuan-ms: > It doesn't really matter but is there a particular reason for this to be a property and not just an attribute set on `__init__`?

Because some of the information values would be generated when testplan finished(eg, end time). And also user can add some custom information values.

testplan/report/testing/base.py
yuxuan-ms: > why does the order matter?

For the user experience. For example, if there are "git commit id", "git branch", "git repo" and "command line string" in the information values. Displaying "command line string", "git commit id", "git branch", "git repo" would be better than Displaying  "git commit id", "command line string", "git branch", "git repo".

PR #332, opened 2019-12-16
Refactor interactive backend and add UTs

ryan-collingham:
Change how backend runs tests and updates the test report: now the
interactive handler fully owns the report tree and handles emrging in
partial test results returned from runners. This avoids ending up with
dangling report entries owned by test runners.

Remove `RunnableIHandler` base class and move task running logic into single TestRunnerIHandler class for simplicity.

Add unit tests for the interactive backend runner, to cover running
tests at various levels. These UTs will be extended to cover environment
start/stop and other features in future.

There is still some work to do in tidying up the TestRunnerIHandler class - I think there are a few methods that can be removed or simplified in future.

### DIFF
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -8,46 +8,48 @@ from __future__ import absolute_import
 from builtins import super
 from builtins import next
 from future import standard_library
+
 standard_library.install_aliases()
 import functools
 import re
 import six
 import numbers
 import threading
+import multiprocessing.dummy
+import itertools
 
-from testplan.common.config import ConfigOption
-from testplan.common.entity import (RunnableIHandler,
-                                    RunnableIHandlerConfig,
-                                    ResourceStatus)
+from testplan.common import entity
+from testplan.common import config
+from testplan.common.utils import networking
 import testplan.report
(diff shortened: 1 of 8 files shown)


testplan/runnable/interactive/base.py
```diff
@@ -8,132 +8,276 @@
...
     :py:class:`~testplan.runnable.interactive.base.TestRunnerIHandler` runnable
     interactive handler.
     """
+
     @classmethod
     def get_options(cls):
-
-        return {ConfigOption('http_handler',
-                             default=TestRunnerHTTPHandler): object}
+        return {"target": object, "startup_timeout": int, "http_port": int}
```
Pyifan: I'm not sure about not using the ConfigOption wrapper, but specify the default value in the constructor. For one, this introduce inconsistency in our code base. Secondly, I remember I fixed some occurrence of such but I cannot remember what exactly is not working.

ryan-collingham: This is in line with our recent change in direction to add explicit named parameters to class constructors instead of relying on **kwargs. If we are specifying the default value already in the `__init__()` signature, it doesn't make sense to also define a default using ConfigOption.

PR #333, opened 2019-12-17
Highlight xfail test on Web UI

yuxuan-ms:
* Highlight xfail test as orange
* Update some example report (add category)
* Update test case and snapshot
* Update xfail document


doc/en/multitest.rst
```diff
@@ -1281,10 +1281,12 @@ The ``xfail`` decorator mandates a reason that explains why the test is marked a
     def fail_testcase(self, env, result):
         ...
 
-`Xpass` fails the testsuite/testcase, unless the ``strict`` parameter is passed as ``False``. A common example is a test is unstable:
+If a testcase/testsuite is unstable (passing sometimes, failling other times) then `strict=False` can be used. This means if the testcase/testsuite fails it woll be marked "expected to fail" (`xfail`), if it passes it will be marked as "unexpectedly passing" (`xpass`).
```
ryan-collingham: I thought `strict=False` was the default option like it is in pytest? I don't think a user should need  to specify that explicitly.

testplan/web_ui/testing/src/Common/defaults.js
```diff
@@ -3,6 +3,7 @@
  */
 const GREEN = '#228F1D';
 const RED = '#A2000C';
+const ORANGE = 'orange';
```
ryan-collingham: why is this given as a name when all of the other colours are hex values?

testplan/web_ui/testing/src/Report/reportUtils.js
```diff
@@ -102,8 +102,8 @@ const propagateIndicesRecur = (entries, parentIndices) => {
       caseCount.failed += descendantsIndices.case_count.failed;
     } else {
       // Count testcase's status.
-      caseCount.passed += entry.status === 'passed';
-      caseCount.failed += entry.status === 'failed';
+      caseCount.passed += ['passed', 'xpass'].includes(entry.status);
```
ryan-collingham: This is fine for now but in future I think we should be keeping these pass/fail counts in the backend and not calculating them here.

testplan/web_ui/testing/src/Common/defaults.js
yuxuan-ms: > why is this given as a name when all of the other colours are hex values?

I think the named css colour is more readable. It's ok. I will change the value to #FFA500

yuxuan-ms:
> Looks fine. My only major query is about the default operation of xfail - my impression was that it defaulted to `strict=False` like pytest, but the docs you added make it sound like it actually defaults to `strict=True`...?

I have updated the document. Could you review & check the wording?

doc/en/multitest.rst
```diff
@@ -1269,22 +1269,24 @@ Also keep in mind that testplan will take a little bit of effort to monitor exec
 Xfail
 -----
 
-You can mark testsuite/testcase that you expect to fail to testplan can deal with them accordingly and present as pass in the test report, while keeping the testsuites/testcase `green`.
+You can mark testsuite/testcase that you expect to fail testplan can deal with them accordingly and present as pass in the test report, while keeping the testsuites/testcase `green`.
```
ryan-collingham: This sentence doesn't read very well to me. Also I think it is misleading - if an xfail test fails it will be marked as xfail (orange) not pass (green)?

How about something like:

Testcases and testsuites that you expect to fail can be marked with the `@xfail` decorator. These failures will be visible in the test report, highlighted in orange. Expected failures will not cause the testplan as a whole to be considered a failure.

doc/en/multitest.rst
```diff
@@ -1269,22 +1269,24 @@ Also keep in mind that testplan will take a little bit of effort to monitor exec
...
 
-A Xfail means that you expect a test to fail for some reason. A common example is a test for a feature not yet implemented, or a bug not yet fixed. When a test passes despite being expected to fail(marked with `xfail`), it's an `Xpass` and will be reported as failed in the test report
+The Xfail means that you expect a test to fail for some reason. If a testcase/testsuite is unstable (passing sometimes, failling other times) then `strict=False` (default value is `False`) can be used. This means if the testcase/testsuite fails it will be marked "expected to fail" (`xfail`), if it passes it will be marked as "unexpectedly passing" (`xpass`). 
+Both `xfail` and `xpass` don’t fail the testsuite/testcase.
```
ryan-collingham: "don't cause the parent testsuite or MultiTest to be marked as a failure."

doc/en/multitest.rst
```diff
@@ -1269,22 +1269,24 @@ Also keep in mind that testplan will take a little bit of effort to monitor exec
...
 
 The ``xfail`` decorator mandates a reason that explains why the test is marked as Xfail:
 
 .. code-block:: python
 
-    @xfail(reason='api changes')
-    def fail_testcase(self, env, result):
+    @xfail(reason='unstable test', strict=False)
```
ryan-collingham: since `strict=False` is the default, you don't have to include it here.

doc/en/multitest.rst
```diff
@@ -1269,22 +1269,24 @@ Also keep in mind that testplan will take a little bit of effort to monitor exec
...
+    def unstable_testcase(self, env, result):
         ...
 
-`Xpass` fails the testsuite/testcase, unless the ``strict`` parameter is passed as ``False``. A common example is a test is unstable:
+
+If a test is expect to fail all the time, you can also use the `strict=True` then `xpass` will be considered as `fail`. This will make the unexpectedly passing results to fail test testcase/testsuite.
```
ryan-collingham: "This will cause the unexpectedly passing result to fail the testcase or testsuite."

ryan-collingham:
Looks good, thanks for making the changes!

PR #334, opened 2019-12-17
Control test environments via API and UI

ryan-collingham:
Allow test environments to be stopped and started via the interactive
API. Add new "env_status" field to keep track of the environment state.


ryan-collingham:
Update: I implemented the UI side of the environments control (it turned out not to be as difficult as I thought it might be). Since it depends on this PR, I've added the changes here instead of creating a separate PR.

testplan/report/testing/base.py
```diff
@@ -470,6 +470,7 @@ def __init__(self,
                  tags=None,
                  part=None,
                  fix_spec_path=None,
+                 env_status=None,
```
Pyifan: Does it make sense for Report object to hold the status of env? I think all report related status should go to the Status class, maybe add the UNINIT state to indicate env is not started yet.

testplan/runnable/interactive/base.py
```diff
@@ -279,6 +283,78 @@ def test(self, test_uid, runner_uid=None):
...
+        test = self.test(test_uid, runner_uid=runner_uid)
+        test_irunner = test.cfg.interactive_runner(test)
+
+        results = self._run_all_test_operations(
+            test_irunner.start_resources(), await_results
+        )
+
+        if await_results:
+            results.append(
+                self._set_env_status(test_uid, entity.ResourceStatus.STARTED)
```
Pyifan: How do we know the env is indeed started?

testplan/report/testing/base.py
ryan-collingham: I see where you are coming from and kind of agree that this isn't a perfect fit for the report object. However currently the design of the interactive mode is such that the interactive API reflects the report structure, so any information to be sent over the API should be included in the report.

The problem with adding another UNINIT status is that that doesn't allow the interactive UI to know when the environment is either STARTING or STOPPING - the env_status gives more information (STOPPED -> STARTING -> STARTED ->> STOPPING -> STOPPED cycle) versus a simple UNINIT -> READY switch. So I think for now it is best to keep the env_status.

testplan/runnable/interactive/base.py
ryan-collingham: When `await_results=True` the environment is started synchronously, so by the time we get to here the start actions must have completed successfully (otherwise an exception would be raised).

In the async case, we should add a check that the environment was started without error.

PR #335, opened 2019-12-19
Consolidate Categories and ReportCategories classes

Pyifan:
There 2 classes are overlapping, remove Categories.
Also change TESTCASE = 'testcase' --> CASE = 'case' to be consistent with SUITE = 'suite'


examples/Test Output/Exporters/JSON/json_to_pdf.py
```diff
@@ -5,7 +5,7 @@
 import argparse
 import json
 
-from testplan.report.testing import TestReport
+from testplan.report import TestReport
```
ryan-collingham: If this import was missing before... how did it ever work?

testplan/common/utils/logger.py
```diff
@@ -15,7 +15,7 @@
 import logging
 
 from testplan.common.utils.strings import Color
-from testplan.report.testing import Status
+from testplan.report import Status
```
ryan-collingham: Again, how was this not broken before...?

testplan/report/testing/base.py
```diff
@@ -136,19 +136,19 @@ class ReportCategories(object):
     """
 
     TESTPLAN = "testplan"
-    TESTGROUP = "testgroup"
+    TESTGROUP = "testgroup"     # test group of unspecific type?
     MULTITEST = "multitest"
+    SUITE = "suite"
+    CASE = "case"
```
ryan-collingham: This will break the UI since it expects testcases to have a category of "testcase". Please can you change back the value to "testcase"?

testplan/runnable/interactive/base.py
```diff
@@ -662,7 +662,7 @@ def _initial_report(self):
             for suite in test.suites:
                 suite_name = suite.__class__.__name__
                 suite_report = testplan.report.TestGroupReport(
-                    name=suite_name, uid=suite_name, category="suite"
+                    name=suite_name, uid=suite_name, category='suite'
```
ryan-collingham: Why have you made this change? Double quotes are the standard used by `black`.

testplan/testing/multitest/base.py
```diff
@@ -249,7 +242,7 @@ def pre_post_step_report(self):
         if self._pre_post_step_report is None:
             self._pre_post_step_report = TestGroupReport(
                 name='Pre/Post Step Checks',
-                category=Categories.SUITE,
+                category=ReportCategories.SUITE,
```
ryan-collingham: Nice, better to use the ReportCategories values here for sure.

tests/functional/testplan/runnable/interactive/test_api.py
```diff
@@ -160,7 +160,7 @@ def plan():
         "/report/tests/ExampleMTest/suites/ExampleSuite/testcases",
         [
             {
-                "category": "testcase",
+                "category": "case",
```
ryan-collingham: This will be a problem, the UI won't handle the value of "case".

tests/functional/testplan/testing/fixtures/base/passing/report.py
```diff
@@ -1,4 +1,4 @@
-from testplan.report.testing import TestReport, TestGroupReport
+from testplan.report import TestReport, TestGroupReport
```
ryan-collingham: A lot of unrelated changes being made... it's fine for now but in future could you split out such cosmetic changes into a separate PR?

testplan/report/testing/base.py
Pyifan: Yea, i was debating whether I should change suite to testsuite or change testcase to case. I recon the latter one will have smaller impact as the CASE type is lately added by you recently. I prefer to keep the terminologies consistent, either testsuite/testcase or suite/case, but definitely not testsuite/case. We can do a global replace in the UI code from testcase -> case.

ryan-collingham: as discussed, let's go with testsuite/testcase and be consistent everywhere please.

testplan/web_ui/testing/src/Common/defaults.js
```diff
@@ -45,7 +45,7 @@ const CATEGORIES = {
   'boost-test-suite': 'suite',
   'gtest-suite': 'suite',
   'parametrization': 'parametrization',
-  'testcase': 'testcase'
+  'case': 'case'
```
ryan-collingham: Why is this changing from testcase -> case ...?

testplan/web_ui/testing/src/Common/defaults.js
```diff
@@ -62,7 +62,7 @@ const CATEGORY_ICONS = {
   'boost-test-suite': 'BS',
   'gtest-suite': 'GS',
   'parametrization': 'P',
-  'testcase': 'C'
+  'case': 'C'
```
ryan-collingham: I think this should be updated to `'testcase': 'TC'` now.

testplan/web_ui/testing/src/Common/sampleReports.js
```diff
@@ -55,7 +55,7 @@ const TESTPLAN_REPORT = {
           },
           "entries": [
             {
-              "category": "testcase",
+              "category": 'case',
```
ryan-collingham: Why are these changed...?

testplan/runnable/interactive/base.py
Pyifan: just for consistency with context

PR #341, opened 2020-01-06
Keep the attributes of test suite defined by user in report

raoyitao:
* user can define some useful attributes in test suite and they will be
  saved in report in a dictionary named 'extra_attributes'.
* Add a common util for verifying email address.

### DIFF
--- a/testplan/base.py
+++ b/testplan/base.py
@@ -263,6 +263,11 @@ class Testplan(entity.RunnableManager):
                                 default_options=self._default_options)
 
     @property
+    def runnable(self):
+        """Runnable instance."""
+        return self._runnable
+
+    @property
     def args(self):
         """Parsed arguments."""
         return self._parsed_args
--- a/testplan/common/utils/validation.py
+++ b/testplan/common/utils/validation.py
@@ -29,4 +29,8 @@ def has_method(method_name):
 
 def is_valid_url(url):
     """Validator that checks if a url is valid"""
-    return True if validators.url(url) is True else False
+    return bool(validators.url(url))
+
+
+def is_valid_email(email):
+    return bool(validators.email(email))
--- a/testplan/report/testing/base.py
+++ b/testplan/report/testing/base.py
@@ -491,6 +491,7 @@ class TestGroupReport(BaseReportGroup):
                  category=ReportCategories.TESTGROUP,
                  tags=None,
                  part=None,
+                 extra_attributes=None,
                  fix_spec_path=None,
                  env_status=None,
                  **kwargs):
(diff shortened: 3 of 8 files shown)


testplan/common/utils/validation.py
```diff
@@ -30,3 +30,7 @@ def _validator(kls):
 def is_valid_url(url):
     """Validator that checks if a url is valid"""
     return True if validators.url(url) is True else False
+
+
+def is_valid_email(email):
+    return True if validators.email(email) is True else False
```
ryan-collingham: why not simply `return validators.email(email)`...?

testplan/report/testing/base.py
```diff
@@ -509,6 +510,8 @@ def __init__(self,
         self.part = part  # i.e. (m, n), while 0 <= m < n and n > 1
         self.part_report_lookup = {}
 
+        self.extra_attributes = extra_attributes
```
ryan-collingham: What exactly is the purpose of storing these extra attributes?

testplan/base.py
```diff
@@ -262,6 +262,11 @@ def parser(self):
         return self._cfg.parser(name=self._cfg.name,
                                 default_options=self._default_options)
 
+    @property
+    def runnable(self):
```
ryan-collingham: is this used anywhere?

testplan/common/utils/validation.py
raoyitao: Because the function returns an Exception object if validation fails, sound a little weird but it does in that way

testplan/report/testing/base.py
raoyitao: For now, this is needed by internally used RFBEmailExporter. I think we can record these attributes (if nothing defined it's None) in test report, maybe useful in the future. But I guess maybe we need to check this content should be jsonified.

testplan/common/utils/validation.py
ryan-collingham: Oh right, in that case I think you can convert to bool like `return bool(validators.email(email)` - from the docs, validation failures are Falsey: https://validators.readthedocs.io/en/latest/#basic-validators

testplan/base.py
raoyitao: I used it before, but I changed my code so it's not in use any more, I can remove it. We may also keep it, I saw in `test_plan_base.py` plan._runnable is frequently used.

ryan-collingham: It's fine I was mostly just curious.

testplan/report/testing/base.py
ryan-collingham: ok thanks for clarifying

testplan/testing/multitest/suite.py
```diff
@@ -256,6 +256,13 @@ def _testsuite(klass):
         klass.__tags__ = {}  # used for UI
         klass.__tags_index__ = {}  # used for actual filtering
 
+    klass.__extra_attributes__ = {
```
Pyifan: Add some comment here?

raoyitao: added

PR #345, opened 2020-01-10
Some improvement or fixes

raoyitao:
* Fix an issue that when testcase has a timeout, the exception info
  cannot be caught, and our code itself raises an `IndexError`.
* Enhancement for `HttpExporter` that if the report has some objects
  that cannot be serialized by json, then we provide a customized
  json encoder to deal with it to avoid exception.
* Customized name_func for parametrized testcase cannot ensure no
  deplicated testcase names exist: after we simply renaming the invalid
  names there might still conflict be found, we need to check again.

### DIFF
--- a/testplan/common/utils/timing.py
+++ b/testplan/common/utils/timing.py
@@ -9,6 +9,9 @@ import functools
 import time
 import datetime
 import threading
+import inspect
+
+from .exceptions import format_trace
 
 
 class TimeoutException(Exception):
@@ -92,10 +95,14 @@ def timeout(seconds, err_msg='Timeout after {} seconds.'):
     def timeout_decorator(func):
         """"""
         def _new_func(result, old_func, old_func_args, old_func_kwargs):
-            result.append(old_func(*old_func_args, **old_func_kwargs))
+            try:
+                result.append(old_func(*old_func_args, **old_func_kwargs))
+            except Exception as exp:
+                result[0] = False
+                result.append(format_trace(inspect.trace(), exp))
 
         def wrapper(*args, **kwargs):
-            result = []
+            result = [True]
             new_kwargs = {
                 'result': result,
                 'old_func': func,
@@ -110,7 +117,7 @@ def timeout(seconds, err_msg='Timeout after {} seconds.'):
                 thd.join()
                 raise TimeoutException(err_msg.format(seconds))
             else:
-                return result[0]
+                return result
 
         return functools.wraps(func)(wrapper)
 
--- a/testplan/exporters/testing/http/__init__.py
+++ b/testplan/exporters/testing/http/__init__.py
@@ -3,6 +3,8 @@ HTTP exporter for uploading test reports via http transmission. The web server
 must be able to handle POST request and receive data in JSON format.
 """
 
+import json
+
 import requests
 from schema import Or, And, Use
 
@@ -12,6 +14,12 @@ from testplan.common.exporters import ExporterConfig
 from testplan.report.testing.schemas import TestReportSchema
 
 from ..base import Exporter
+
+
+class CustomJsonEncoder(json.JSONEncoder):
+    """To jsonify data that cannot be serialized by default JSONEncoder."""
+    def default(self, obj):  # pylint: disable = method-hidden
+        return str(obj)
 
 
 class HTTPExporterConfig(ExporterConfig):
@@ -33,8 +41,10 @@ class HTTPExporter(Exporter):
     """
     Json Exporter.
 
-    :param url: Http url for posting data.
-    :type url: ``str``
+    :param http_url: Http url for posting data.
+    :type http_url: ``str``
+    :param timeout: Connection timeout.
+    :type timeout: ``int``
 
     Also inherits all
     :py:class:`~testplan.exporters.testing.base.Exporter` options.
@@ -50,9 +60,12 @@ class HTTPExporter(Exporter):
         errmsg = ''
 
         if data and data['entries']:
+            headers = {'Content-Type': 'application/json'}
             try:
                 response = requests.post(
-                    url=url, json=data, timeout=self.cfg.timeout)
+                    url=url, headers=headers,
+                    data=json.dumps(data, cls=CustomJsonEncoder),
+                    timeout=self.cfg.timeout)
                 response.raise_for_status()
             except requests.exceptions.RequestException as exp:
                 errmsg = 'Failed to export to {}: {}'.format(url, str(exp))
--- a/testplan/testing/multitest/base.py
+++ b/testplan/testing/multitest/base.py
@@ -594,9 +594,13 @@ class MultiTest(Test):
 
                 time_restriction = getattr(testcase, 'timeout', None)
                 if time_restriction:
-                    timeout_deco(
+                    # pylint: disable=unbalanced-tuple-unpacking
+                    executed, execution_result = timeout_deco(
                         time_restriction, 'Testcase timeout after {} seconds'
                     )(testcase)(self.resources, case_result)
+                    if not executed:
+                        testcase_report.logger.error(execution_result)
+                        testcase_report.status_override = Status.ERROR
                 else:
                     testcase(self.resources, case_result)
 
--- a/testplan/testing/multitest/parametrization.py
+++ b/testplan/testing/multitest/parametrization.py
@@ -175,6 +175,10 @@ def _ensure_unique_names(functions):
     """
     name_counts = collections.Counter([f.__name__ for f in functions])
     dupe_names = {k for k, v in name_counts.items() if v > 1}
+
+    if len(dupe_names) == 0:
+        return
+
     dupe_counter = collections.defaultdict(int)
 
     for func in functions:
@@ -184,6 +188,15 @@ def _ensure_unique_names(functions):
             func.__name__ = '{}__{}'.format(name, count)
             dupe_counter[name] += 1
 
+    # Can still have name conflict if user defined name_func does not work well
+    name_counts = collections.Counter([f.__name__ for f in functions])
+    dupe_names = {k for k, v in name_counts.items() if v > 1}
+
+    if len(dupe_names) > 0:
+        msg = (
+            '"name_func" produces duplicated function names and cannot be '
+            'automatically fixed by adding a suffix number, please check it.')
+        raise ParametrizationError(msg)
 
 def _generate_func(
      function, name_func, tag_func, docstring_func, tag_dict, kwargs):


doc/en/multitest.rst
```diff
@@ -1024,11 +1026,18 @@ for sample implementation.
       """
       number_list = kwargs['number_list']
 
-      return '{func_name}__numbers_{joined_list}__result_{expected}'.format(
-          func_name=func_name,
-          joined_list='_'.join(map(str, number_list)),
-          expected=kwargs['expected']
-      )
+      if kwargs.get('for_display', False):
```
ryan-collingham: I think this is over-complicating. Surely the custom name generator can just return a single human-readable name? Why does it need to be a valid python identifier?

testplan/common/utils/timing.py
```diff
@@ -92,10 +95,14 @@ def timeout(seconds, err_msg='Timeout after {} seconds.'):
     def timeout_decorator(func):
         """"""
         def _new_func(result, old_func, old_func_args, old_func_kwargs):
-            result.append(old_func(*old_func_args, **old_func_kwargs))
+            try:
+                result.append(old_func(*old_func_args, **old_func_kwargs))
+            except Exception as exp:
+                result[0] = False
+                result.append(format_trace(inspect.trace(), exp))
```
ryan-collingham: It's not very important but I think that `traceback.format_exc()` would be simpler here, I don't see why we need a custom function for formatting stack traces when it's right there in the standard library.

testplan/exporters/testing/http/__init__.py
```diff
@@ -14,6 +16,12 @@
 from ..base import Exporter
 
 
+class CustomJsonEncoder(json.JSONEncoder):
```
ryan-collingham: Why is this needed exactly?

testplan/report/testing/base.py
```diff
@@ -552,6 +552,7 @@ class TestGroupReport(BaseReportGroup):
 
     def __init__(self,
                  name,
+                 display_name=None,
```
ryan-collingham: We definitely don't need an additional display_name parameter on the report. If a custom name is specified for a testcase/testsuite, that becomes the `name` parameter. Only if no name is specified do we use the function name as a default value. The report object only ever has one name.

testplan/report/testing/schemas.py
```diff
@@ -195,11 +195,15 @@ class TestCaseReportSchema(ReportSchema):
...
     status_override = fields.String(allow_none=True)
 
+    display_name = fields.String(allow_none=True)
+
     entries = fields.List(EntriesField())
 
     status = fields.String(dump_only=True)
     runtime_status = fields.String(dump_only=True)
     counter = fields.Dict(dump_only=True)
+    passed = fields.Bool()
```
ryan-collingham: We don't need these extra fields - we already send the status, this information is redundant.

testplan/testing/multitest/suite.py
```diff
@@ -382,12 +383,13 @@ def wrapper(function):
         function.__tags__ = copy.deepcopy(tag_dict)
 
         if parameters is not None:  # Empty tuple / dict checks happen later
-
+            function.__display_name__ = name
```
ryan-collingham: Here, you should set `function.__display_name__ = name if name else function.__name__`.

Ideally this wouldn't be a dunder attribute (i.e. just `function._display_name`, however since that is the established convention it's probably best to stick with it for now.

testplan/testing/multitest/base.py
```diff
@@ -340,6 +340,7 @@ def dry_run(self, status=None):
                 testcase_report = TestCaseReport(
                     name=testcase.__name__,
```
ryan-collingham: Here and below you should set `name=testcase.__display_name__` and not set any `display_name` parameter in the report.

doc/en/multitest.rst
raoyitao: the name_func is used for generating names for parametrized testcases. Testplan will check if the generated name is not a valid python identifier then a warning message is printed, and testplan will simply reset the name with suffix. So user can know that their customized naming function does not work well. So, this name_func is designed to generate function name, not human readable display name. I also think that this is over-complicating, it is better to use another default name_func for generating hunan-readable names.

The question is: should we provide another another argument to let user customize the name_func for display name? if we should, that seems a little complicated, if we do not, then an unwanted name can be displayed such as "My test function (arg1: 1, arg2: [{'a': 'b'}, {'c': 'd'} ...])"

testplan/common/utils/timing.py
raoyitao: I' m not sure, I saw that the __exit__() function of Exception Logger calls this format_trace. Is it some different ?

testplan/report/testing/base.py
raoyitao: Yes the display name should be the name of testcase here, my concern is , if we totally remove that original function name from report, then it is hard for us to debug or trace with that json report, or maybe one day use want a feature that switch back to testcase identifier on UI. It makes sense that the name of report is 'display name', just like that name of Multitest is a human readable string, but somehow I think we should also keep an attribute 'testcase' or 'testcase_name' in the report.

testplan/report/testing/schemas.py
raoyitao: because in another PR you left a comment that don't check the pass or fail status for a testcase, it already have the attribute. There the report has already been dumped to json so no 'passed' or 'failed' key is kept. But I agree to remove the attributes here, because in a recent commit, it add an unstable status, for example, a skipped testcase is unstable, but if a suite has 2 cases: passed and unstable, then the suite is passed, so RFB cannot reply on 'passed' or 'failed' attributes because in RFB check it has different meaning, finally it is either pass or fail, even all testcases are set skipped, it passes but no unstable status should be reported. So I think I should check the status again in the rfb exporter.

testplan/testing/multitest/suite.py
raoyitao: make sense

doc/en/multitest.rst
```diff
@@ -1011,8 +1011,10 @@ However you can provide custom name generation functions to override this
 functionality via ``@testcase(parameters=..., name_func=custom_name_func)``
 syntax. You just need to implement a function that accepts
 ``func_name`` and ``kwargs`` as arguments, ``func_name`` being a string and
-``kwargs`` being an ``OrderedDict``.
-See
+``kwargs`` being an ``OrderedDict``. You may check if there is an argument
+``for_display`` in `kwargs``, if the value is true, then it is suggested to
```
Pyifan: I agree with Ryan, I don't think we need another for_display here. If this is the display name we discussed, it does not belong here (parameterization session).

doc/en/multitest.rst
Pyifan: Does not seem necessary to me at least for now. The idea is if user specify his own name_func, then he takes full control. I do not think it is super difficult to come up with a pattern that is both human-readable && python identifier valid.

testplan/testing/multitest/parametrization.py
```diff
@@ -325,14 +347,21 @@ def default_name_func(func_name, kwargs):
...
+    else:
+        arg_strings = ['{arg}_{{{arg}}}'.format(arg=arg) for arg in kwargs]
+        template = '{func_name}__' + '__'.join(arg_strings)
     return template.format(func_name=func_name, **kwargs)
 
 
 def generate_functions(
     function,
     parameters,
+    name,
```
Pyifan: doc string?

testplan/testing/multitest/suite.py
```diff
@@ -357,6 +357,7 @@ def _testcase(func):
 
 
 def _testcase_meta(
+    name=None,
```
Pyifan: I think it's confusing to have a name param in @testcase decorator. Intuitively the name of a testcase is the function name. We can have a display_name for the testcase though. I suggest we change name ==> display_name here.

testplan/report/testing/base.py
Pyifan: I prefer that we keep the original testcase name (func name) unchanged in the report, just add one additional field say display_name. Changing the name itself might have other side effect that we cannot predict right now.

Pyifan:
I think adding 'display name' for testcase is a very good addition! I prefer that we add this as an additional field in the report, and leave it to the Exporter to render the user friendly name. On the other hand we need to caution with the complexity when 'display name' encounters 'parameterization'. My suggestion is to redesign @testcase decorator as:

@testcase(display_name=...) for single testcase
@testcase(display_name_func=...) for parameterized testcase
and get rid of the original
@testcase(name_func=...)

testplan/exporters/testing/http/__init__.py
raoyitao: user can define attributes in test suites, but we don't have a check that all their attributes can be serialized by JSON (this rarely happens, but user can define weird things like some objects of non-basic type). To avoid exception in dumping JSON so we can add this customized encoder. Xml pdf exported is not affected because they choose the pre-defined attributes to dump. I think Json report is the most basic if we want to debug or trace, so it should contains all the information. Json exporter also need it so I move this Encoder to exporters/testing/json/__init__.py and HTTPExporter imports it.

doc/en/multitest.rst
ryan-collingham: Why does the name need to be a valid python identifier?

I agree it's probably not needed for now.

testplan/common/utils/timing.py
ryan-collingham: I don't think we need it, it just does the same as standard library function.

testplan/exporters/testing/http/__init__.py
ryan-collingham: ok, I think that if a user defines some weird attribute on a testsuite that is quite unusual and we shouldn't worry about making it work. But if it's only a simple change it's fine to leave in.

testplan/report/testing/base.py
ryan-collingham: No, as discussed absolutely we cannot have more than one name in the report as it will over complicate the UI code too much. One name only please.

testplan/report/testing/schemas.py
ryan-collingham: I think you are confusing with something else. Please can you just remove these passed/failed fields, the JSON should not contain redundant data to minimise its size.

testplan/testing/multitest/suite.py
ryan-collingham: the whole point of this PR is to provide a way to override the name of a testcase via the decorator. The function name is just the default string used for the name, it has no other special meaning.

doc/en/multitest.rst
raoyitao: we have decided to add display name, so the name_func is now used for generating display name for parametrized testcases. and our default name func should be improved a little, because most time the string format of parameter is not simple positive integer or letters, and they will be renamed again.

testplan/common/utils/timing.py
raoyitao: I will check it and if not necessary I will remove it.

testplan/exporters/testing/http/__init__.py
raoyitao: It's only a few lines change, more safe, some team like proxima like to dynamically throw things into modules or classes. Let me keep it.

testplan/report/testing/schemas.py
raoyitao: I just checked with Yuxuan/Yifan about the changes made for Report status. I moved the attributes.

testplan/report/testing/base.py
raoyitao: I think we have different point of views.
**Ryan**: only a `name` attribute added into report, inside testcase it's actually __display_name__, no redundant attributes kept in report.
**Yifan**: `name` is intuitively the defined name of testcase, so a `display_name` is added in both testcase decorator and test report.
**Me**: `name` is added for testcase decorator because it's short and regular, but in json report we keep both original name and display name, when users search on UI, I think both of them are needed, just like when you search a people, the real name and user id will both be considered.

ryan-collingham: I don't really care too much about what attributes you set on the testcase functions but, to reiterate, we must not overcomplicate the report by adding an extra "display_name" field. The name parameter of testcase is an override, that becomes the testcases unique identifier and is the only name used for search and filter.

testplan/common/utils/timing.py
raoyitao: Hi Ryan, as I check again, **format_trace(inspect.trace(), exc))** is used at many places in open source repo. To avoid making it not stable so I won't change it in this pull request. If you though it's necessary then another PR can be opened.

ryan-collingham: that's fine, I wasn't suggesting you remove it in this PR. I was more just asking the broader question of do we need it, and do you need to use it right here. We can look at removing in another PR, if it can be replaced by the traceback module.

PR #347, opened 2020-01-13
Show the testplan name on UI page title

yuxuan-ms:


testplan/web_ui/testing/src/Report/BatchReport.js
```diff
@@ -142,6 +142,11 @@ class BatchReport extends React.Component {
 
   render() {
     const {reportStatus, reportFetchMessage} = GetReportState(this.state);
+
+    if (this.state.report && this.state.report.name) {
+      window.document.title = this.state.report.name;
```
ryan-collingham: just a question but is `window.document.title` any different from just `document.title`? https://developer.mozilla.org/en-US/docs/Web/API/Document/title

yuxuan-ms: > just a question but is `window.document.title` any different from just `document.title`? https://developer.mozilla.org/en-US/docs/Web/API/Document/title

Same thing, document === window.document
https://stackoverflow.com/questions/9895202/what-is-the-difference-between-window-screen-and-document-in-javascript

PR #351, opened 2020-01-16
Test runnning parametrizations interactively in UT

ryan-collingham:
Includes a small change to interactive runner class for command-line usability and
testability purposes. Now, all methods to run a testcase or group of
testcases (suite, MultiTest etc.) will return a single report object
instead of a list of many report objects.

### DIFF
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -189,17 +189,19 @@ class TestRunnerIHandler(entity.Entity):
             default local runner.
         :param await_results: Whether to block until tests are finished,
             defaults to True.
-        :return: If await_results is True, returns a list of test results.
-            Otherwise, returns a list of async result objects.
+        :return: If await_results is True, returns a testplan report.
+            Otherwise, returns a future which will yield a testplan report when
+            ready.
         """
+        if not await_results:
+            return self._run_async(self.run_all_tests, runner_uid)
+
         all_tests = self.all_tests(runner_uid)
 
-        return [
-            self.run_test(
-                test_uid, runner_uid=runner_uid, await_results=await_results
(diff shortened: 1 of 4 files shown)


ryan-collingham:
Update: also pushed a patch to simplify the automatic environment control. When running async, the environment control is all performed in the same thread and not split over multiple async threads, to guarantee correct order.

testplan/runnable/interactive/base.py
```diff
@@ -813,6 +813,33 @@ def _stop_environment(self, test_uid, stop_generator):
...
+                    runner_uid=runner_uid,
+                )
+            self.start_test_resources(
+                test_uid, runner_uid=runner_uid,
+            )
+            resources_started = True
+
+        yield
+
+        if resources_started:
```
Pyifan: I still don't think we should stop the resources after running individual tests.

testplan/runnable/interactive/base.py
```diff
@@ -271,24 +245,36 @@ def run_test_case(
...
         irunner = test.cfg.interactive_runner(test)
         test_run_generator = irunner.run(suite=suite_uid, case=case_uid)
 
-        if await_results:
-            return self._run_all_test_operations(test_run_generator)
-        else:
-            return self._run_async(
-                self._run_all_test_operations, test_run_generator
+        with self._auto_start_stop_environment(test_uid, runner_uid):
+            test_report = self._merge_test_reports(
```
Pyifan: why do we need to do the merge there, we just ran one testcase.

ryan-collingham: If the testcase is parametrized, we will run all parametrizations here - which will generate multiple reports.

PR #352, opened 2020-01-17
Replace format_trace with traceback.format_exc()

ryan-collingham:
We may have a strong tradition of reinventing the wheel, but when there
is a function right there in the standard library that does what we
need, we really need to ask why we would use something else?

See traceback module docs: https://docs.python.org/3.7/library/traceback.html

@raoyitao I'm aware this will clash with your PR, I'm happy for you to merge yours
first and I can update this one.

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -9,20 +9,18 @@ import signal
 import time
 import uuid
 import threading
-import inspect
 import psutil
 import functools
 from collections import deque, OrderedDict
+import traceback
 
 from schema import Or, And, Use
 
 from testplan.common.config import Config, ConfigOption
-from testplan.common.utils.exceptions import format_trace
 from testplan.common.utils.thread import execute_as_thread
 from testplan.common.utils.timing import wait
 from testplan.common.utils.path import makeemptydirs, makedirs, default_runpath
 from testplan.common.utils import logger
-from testplan.common.utils import networking
 
 
 class Environment(object):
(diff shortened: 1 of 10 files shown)


testplan/common/entity/base.py
```diff
@@ -141,10 +139,11 @@ def start(self):
                 resource.start()
                 if not resource.cfg.async_start:
                     resource.wait(resource.STATUS.STARTED)
-            except Exception as exc:
-                msg = 'While starting resource [{}]{}{}'.format(
-                    resource.cfg.name, os.linesep,
-                    format_trace(inspect.trace(), exc))
+            except Exception:
+                msg = 'While starting resource [{}]\n{}'.format(
```
yuxuan-ms: `\n`  is compatible with Windows and Linux on the console. But It's not compatible on the log files. I think `os.linesep` is a good multiplatform solution.

ryan-collingham: If a file is opened in text mode (the default), python will convert any \n to \r\n before writing:

```
Python 3.7.1 (v3.7.1:260ec2c36a, Oct 20 2018, 14:57:15) [MSC v.1915 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> with open("file.txt", "w") as f:
...     f.write('\n'.join(["line", "line", "yet another line"]))
...
26
Python 3.7.1 (v3.7.1:260ec2c36a, Oct 20 2018, 14:57:15) [MSC v.1915 64 bit (AMD64)] on win32
Type "help", "copyright", "credits" or "license" for more information.
>>> with open("file.txt", "rb") as f:
...     byte_contents = f.read()
...
>>> print(byte_contents)
b'line\r\nline\r\nyet another line'
```

See https://docs.python.org/3/tutorial/inputoutput.html#reading-and-writing-files

> When writing in text mode, the default is to convert occurrences of \n back to platform-specific line endings.

ryan-collingham: The _only_ reason to use `os.linesep` is if you are writing to a file or socket opened in _binary_ mode. I'm not aware of anywhere in Testplan in which we do that.

PR #353, opened 2020-01-22
Refactor MultiTest into reusable component methods

ryan-collingham:
Refactor MultiTest methods into smaller, more composable methods. The
driving goal of this change is to improve the interaction between
MultiTest and the interactive runner to remove bugs in the interactive
mode, but I hope it also provides a readability improvement and possibly
remove some bugs in MultiTest.

Replaces the home-grown thread pool implementation that was jammed into
MultiTest previously and just use concurrent.futures package for
multithreading testcases that belong to an execution group.


ryan-collingham:
Need to fix up tests for python 2 (they're passing on python 3).

ryan-collingham:
To guide review, the methods with the most substantive changes are the dry_run(), run_tests(), _run_suite() and _run_testcase() methods. The overarching design has been to ensure that all of these return an appropriate report object instead of mutating some passed-in report object. By making these functions "purer" by reducing side-effects, they will be more useful to re-use for interactive runner. For the _run_suite() method, I also split out the running of serial and parallel testcases into sub methods to more clearly separate the two methods of running.

Apart from that, I also reorganised the order of methods slightly, so that public methods are at the top of the class and "private" methods are at the bottom, and I also applied `black` to reformat the file. Sorry that this has ended up making the diff a bit larger, but overall I think it improves the readability quite a bit.

testplan/testing/multitest/base.py
```diff
@@ -1,72 +1,68 @@
-"""Multitest main test execution framework."""
+"""MultiTest test execution framework."""
+from __future__ import unicode_literals
+from __future__ import print_function
+from __future__ import division
+from __future__ import absolute_import
```
morsedav: I get OCD with this. My OCD is satiated with this:

```py
from __future__ import (
    unicode_literals, print_function, division, absolute_import
)
```

morsedav: No change needed. I'll just manage my OCD better.

ryan-collingham: It's funny, I get OCD when multiple imports are grouped on one line - though I totally get it's just a personal preference.

These lines were added by the `pasteurize` tool (https://python-future.org/quickstart.html#to-convert-existing-python-3-code), I just left them as-is.

tests/functional/testplan/testing/multitest/test_suite_decorators.py
```diff
@@ -100,7 +100,7 @@ def test_basic_multitest():
         assert len(mt_entry.entries) == 2  # 2 Suites
         for st_entry in mt_entry.entries:
             assert isinstance(st_entry, TestGroupReport)
-            assert len(st_entry.entries) == 3  # 3 Testcases
```
ryan-collingham: FYI the reason this is changed... previously if a setup method was defined that didn't take a `result` parameter, no report entry would be generated in the report. I made a small tweak to generate a report object anyway, since the setup could still raise an Exception and it makes sense to store details of that Exception inside the setup report. Plus for consistency, I think it's simpler if a setup method always generates some kind of report. Let me know if you have any questions or concerns.

testplan/testing/multitest/base.py
```diff
@@ -310,432 +275,580 @@ def dry_run(self, status=None):
...
+    def _thread_pool_size(self):
+        """
+        :return: the size of thread pool to use, based on configured limits
+        """
+        if self.cfg.thread_pool_size > 0:
+            return min(
+                self.cfg.thread_pool_size, self.cfg.max_thread_pool_size
+            )
+        else:
+            return max(self.cfg.max_thread_pool_size // 2, 2)
```
morsedav: Seems we could be less conservative here, no? Like if `thread_pool_size <= 0` then give at least 5, since it seems that's the minimum default thread pool size in the [doc](https://docs.python.org/3.7/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor):

> ... If max_workers is None or not given, it will default to the number of processors on the machine, multiplied by 5 ...

tests/unit/testplan/testing/multitest/test_multitest.py
```diff
@@ -0,0 +1,281 @@
...
+        name="Mtest", suites=[], **MTEST_DEFAULT_PARAMS
+    )
+    assert mtest.runpath is None
+    assert mtest._runpath is None
+    mtest.run()
+    assert mtest.runpath == path.default_runpath(mtest)
+    assert mtest._runpath == path.default_runpath(mtest)
+
+    # runpath in local cfg
+    custom = os.path.join("", "var", "tmp", "custom")
```
morsedav: Nitpick: This could just be /var/tmp/custom since Windows won't have this anyway.

morsedav:
Looks cleaner and has more tests than the original 👍

ryan-collingham:
> Looks cleaner and has more tests than the original 👍

Great, thanks!

testplan/testing/multitest/base.py
ryan-collingham: Oops sorry I missed this comment before merging... I agree, I'll increase the default value.

tests/unit/testplan/testing/multitest/test_multitest.py
ryan-collingham: I think this test still runs on Windows. I don't think it particular matters whether this path actually exists or not, I think it's just testing that it can be set as an attribute. For now I think safest to leave as is.

PR #357, opened 2020-01-24
Blacken the source files

ryan-collingham:
Apply the code-formatter `black` (https://github.com/psf/black) to all source files, and add a check into the travis CI to ensure files stay blackend. Final part of the dusting.


scripts/utils/crlf_check.sh
```diff
@@ -0,0 +1,13 @@
+#!/bin/bash
+# Look for any CRLF line terminators in source code. Outputs the name of any
+# files containing CRLF and returns 1 if any are found, otherwise returns 0
+# if no CRLF is found.
+find . -path ./testplan/web_ui/testing/node_modules -prune -o -not -type d -exec file "{}" ";" | grep CRLF
```
morsedav: At some point in the future we should make this a pre-commit git hook.

testplan/runners/pools/remote.py
```diff
@@ -468,62 +525,79 @@ def _prepare_remote(self):
...
+            self._remote_testplan_runpath,
+            "--remote-pool-type",
+            self.cfg.pool_type,
+            "--remote-pool-size",
+            str(self.cfg.workers),
+            "--testplan",
+            self._testplan_import_path.remote,
+            "--sys-path-file",
+            self._write_syspath(),
+        ]
```
morsedav: Unfortunate that we'll lose these types of formattings, i.e. the unformatted bit on the left is more readable in this case. THough overall its a fine tradeoff for consistent style throughout.

examples/Interactive/Frameworks/Java/src/testplan/HttpRequests.java
```diff
@@ -1,87 +1,87 @@
...
-            System.out.println("\nPost request failure: " + url);
-            System.out.println("Post data : " + data);
-            System.out.println("Response Code : " + responseCode);
-            System.out.println("POST Response Message : " + con.getResponseMessage());
-            System.out.println("POST Response map: " + responseMap);
-        }
-        return responseMap;
-    }
-}
+/**
```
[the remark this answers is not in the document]
morsedav: Never mind it was the [CRLF checking script](https://github.com/Morgan-Stanley/testplan/blob/c89f65220c13d02d46befa4bc1098df8fa16ca2d/scripts/utils/crlf_check.sh) that changed the java files.

PR #363, opened 2020-02-07
Redesign how interactive runner runs tests

ryan-collingham:
Implement new design for the interaction between the interactive runner
and individual MultiTests to run tests:

* Add new method to MultiTest to run testscases iteratively and yield
  individual testcase reports as they are run. The interactive runner
  handles merging testcase reports into its main report tree. This way,
  setup and teardown methods are only run once per suite and return
  appropriate report entries that can be rendered in the UI.
* Remove separate MultiTestIRunner class that is no longer required.
* Remove run buttons from special "suite related" report entries such as
  testsuite setup/teardown. These methods are run automatically by the
  backend as required and it does not really make sense for them to be
  manually run on their own.
* Small aesthetic tweak to UI to use a spinning icon instead of a bar
  loader when running tests.

### DIFF
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -157,10 +157,8 @@ class TestRunnerIHandler(entity.Entity):
 
         all_tests = self.all_tests(runner_uid)
 
-        return self._create_testplan_report(
+        for test_uid, real_runner_uid in all_tests:
             self.run_test(test_uid, real_runner_uid)
-            for test_uid, real_runner_uid in all_tests
-        )
 
     def run_test(self, test_uid, runner_uid=None, await_results=True):
         """
@@ -179,15 +177,9 @@ class TestRunnerIHandler(entity.Entity):
             return self._run_async(self.run_test, test_uid, runner_uid)
 
         test = self.test(test_uid, runner_uid=runner_uid)
-        irunner = test.cfg.interactive_runner(test)
-        test_run_generator = irunner.run()
 
         with self._auto_start_stop_environment(test_uid, runner_uid):
-            return self._merge_test_reports(
-                self._run_all_test_operations(test_run_generator),
-                test_uid,
-                runner_uid,
-            )
+            self._merge_testcase_reports(test.run_testcases_iter())
 
     def run_test_suite(
         self, test_uid, suite_uid, runner_uid=None, await_results=True
@@ -211,20 +203,11 @@ class TestRunnerIHandler(entity.Entity):
             )
 
         test = self.test(test_uid, runner_uid=runner_uid)
-        irunner = test.cfg.interactive_runner(test)
-        test_run_generator = irunner.run(suite=suite_uid)
 
         with self._auto_start_stop_environment(test_uid, runner_uid):
-            test_report = self._merge_test_reports(
-                self._run_all_test_operations(test_run_generator),
-                test_uid,
-                runner_uid,
+            self._merge_testcase_reports(
+                test.run_testcases_iter(testsuite_pattern=suite_uid)
             )
-
-        if test_report:
-            return test_report[suite_uid]
-        else:
-            return None
 
     def run_test_case(
         self,
@@ -254,69 +237,13 @@ class TestRunnerIHandler(entity.Entity):
             )
 
         test = self.test(test_uid, runner_uid=runner_uid)
-        irunner = test.cfg.interactive_runner(test)
-        test_run_generator = irunner.run(suite=suite_uid, case=case_uid)
 
         with self._auto_start_stop_environment(test_uid, runner_uid):
-            test_report = self._merge_test_reports(
-                self._run_all_test_operations(test_run_generator),
-                test_uid,
-                runner_uid,
+            self._merge_testcase_reports(
+                test.run_testcases_iter(
+                    testsuite_pattern=suite_uid, testcase_pattern=case_uid
+                )
             )
-
-        if test_report:
-            return test_report[suite_uid][case_uid]
-        else:
-            return None
-
-    def run_parametrized_test_case(
-        self,
-        test_uid,
-        suite_uid,
-        case_uid,
-        param_uid,
-        runner_uid=None,
-        await_results=True,
-    ):
-        """
-        Run a single parametrized testcase.
-
-        :param test_uid: UID of the test that owns the testcase.
-        :param suite_uid: UID of the suite that owns the testcase.
-        :param param_uid: UID of the parametrized testcase to run.
-        :param runner_uid: UID of a specific test runner, or None to use the
-            default local runner.
-        :param await_results: Whether to block until the testcase is finished,
-            defaults to True.
-        :return: If await_results is True, returns a testcase report.
-            Otherwise, returns a future which will yield a testcase report when
-            ready.
-        """
-        if not await_results:
-            return self._run_async(
-                self.run_parametrized_test_case,
-                test_uid,
-                suite_uid,
-                case_uid,
-                param_uid,
-                runner_uid,
-            )
-
-        test = self.test(test_uid, runner_uid=runner_uid)
-        irunner = test.cfg.interactive_runner(test)
-        test_run_generator = irunner.run(suite=suite_uid, case=param_uid)
-
-        with self._auto_start_stop_environment(test_uid, runner_uid):
-            test_report = self._merge_test_reports(
-                self._run_all_test_operations(test_run_generator),
-                test_uid,
-                runner_uid,
-            )
-
-        if test_report:
-            return test_report[suite_uid][case_uid][param_uid]
-        else:
-            return None
 
     def test(self, test_uid, runner_uid=None):
         """
@@ -352,19 +279,19 @@ class TestRunnerIHandler(entity.Entity):
             of each resource start operation, otherwise returns a list of async
             result objects.
         """
-        test = self.test(test_uid, runner_uid=runner_uid)
-        test_irunner = test.cfg.interactive_runner(test)
-
-        if await_results:
-            return self._start_environment(
-                test_uid, test_irunner.start_resources()
-            )
-        else:
+        if not await_results:
             return self._run_async(
-                self._start_environment,
-                test_uid,
-                test_irunner.start_resources(),
+                self.start_test_resources, test_uid, runner_uid
             )
+
+        with self.report_mutex:
+            self.report[test_uid].env_status = entity.ResourceStatus.STARTING
+
+        test = self.test(test_uid, runner_uid=runner_uid)
+        test.resources.start()
+
+        with self.report_mutex:
+            self.report[test_uid].env_status = entity.ResourceStatus.STARTED
 
     def stop_test_resources(
         self, test_uid, runner_uid=None, await_results=True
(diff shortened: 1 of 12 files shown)


Pyifan:
This PR might need some discussion. The purpose of RunnableIRunner class family (MultiTestIRunner) provides a consistent interface between TestRunnerIHandler and Runnable (could be MultiTest, GTest and other different type of testing) class. If we remove this layer entirely, then looks like the logic of TestRunnerIHandler is hardwired with MultiTest directly. This might not be a good thing when we need to make interactive mode working for all other tests. Shall we come up with an all-way-rounded design first?

ryan-collingham:
> Not that I love the original design, but I believe the purpose of RunnableIRunner class family (MultiTestIRunner) provides a consistent interface between TestRunnerIHandler and Runnable (could be MultiTest, GTest and other different type of testing) class. If we remove this layer entirely, then looks like the logic of TestRunnerIHandler is hardwired with MultiTest directly. This might not be a good thing when we need to make interactive mode working for all other tests. Shall we come up with an all-way-rounded new design first?

So, the new interface between InteractiveRunner and any Test is based on two methods:

- dry_run() to generate an empty initial report skeleton
- run_testcases_iter() to run testcases iteratively and yield testcase reports that the interactive runner can merge into its report tree

All Test runners already have a dry_run() method. All we need to do is implement run_testcases_iter() for another Test runner and it can work with interactive mode. Maybe I can add a NotImplemented version of run_testcases_iter() into the Test base class.

ryan-collingham:
The travis build is blocked by an unrelated install problem... let me investigate.

testplan/runnable/interactive/base.py
```diff
@@ -157,10 +157,8 @@ def run_all_tests(self, runner_uid=None, await_results=True):
 
         all_tests = self.all_tests(runner_uid)
 
-        return self._create_testplan_report(
+        for test_uid, real_runner_uid in all_tests:
             self.run_test(test_uid, real_runner_uid)
```
morsedav: If you don't want to return a list of [future](https://github.com/Morgan-Stanley/testplan/blob/4804ec125dab9adc3f8d056c4d27ad29faec273b/testplan/runnable/interactive/base.py#L734) perhaps `yield self.run_test(test_uid, real_runner_uid)`? 

I'm not stoked on ignoring the return value if my understanding is correct:
- A [concurrent.futures.Future](https://docs.python.org/3/library/concurrent.futures.html?highlight=concurrent%20futures#concurrent.futures.Future) is returned from [testplan.runnable.interactive.base.TestRunnerIHandler._run_async](https://github.com/ryan-collingham/testplan/blob/4804ec125dab9adc3f8d056c4d27ad29faec273b/testplan/runnable/interactive/base.py#L726)
 - This `Future` is _maybe_ returned from [testplan.runnable.interactive.base.TestRunnerIHandler.run_test](https://github.com/ryan-collingham/testplan/blob/4804ec125dab9adc3f8d056c4d27ad29faec273b/testplan/runnable/interactive/base.py#L163) (which I have other comments about, see below)
- If `run_test` _does_ return this `Future`, it's ignored here.

testplan/runnable/interactive/base.py
```diff
@@ -856,32 +740,18 @@ def _log_async_exceptions(self, future):
...
+                )
 
-        for report in test_reports:
-            initial_report.merge(report)
+                parent_entry = self.report
+                for uid in parent_uids:
+                    parent_entry = parent_entry[uid]
 
-        return initial_report
+                parent_entry[report.uid] = report
```
morsedav: Why not `return` or `yield` the report after this line?

testplan/runnable/interactive/base.py
```diff
@@ -157,10 +157,8 @@ def run_all_tests(self, runner_uid=None, await_results=True):
 
         all_tests = self.all_tests(runner_uid)
 
-        return self._create_testplan_report(
+        for test_uid, real_runner_uid in all_tests:
             self.run_test(test_uid, real_runner_uid)
-            for test_uid, real_runner_uid in all_tests
-        )
 
     def run_test(self, test_uid, runner_uid=None, await_results=True):
```
morsedav: What's the reason for changing this to return either `Future<None>` or `None` rather than `Future<testplan.report.TestReport>` or `testplan.report.TestReport`?

Seems that if you returned the report object from `_merge_testcase_reports` and then returned its result in the `with` statement you could get the same signature as the old method with little effort.

Also I'm not jazzed on this one function having two possible return types but that's what the previous function has and I have seen it in the wild (but in *javascript* code) so I'm not horribly fussed about that.

morsedav:
Just FYI I'm going to mostly yield to @Pyifan's judgement on this one for the high-level impacts on the rest of the project since she'll be able to project better than I the ramifications of these kinda big changes.

That being said, I like the removal of well-intentioned but not-obviously-useful artifacts like [RunnableIRunner](https://github.com/Morgan-Stanley/testplan/pull/363/files#diff-a795ee4fc4235bd7118b7cfb91f192a6L591) / [TestIRunner](https://github.com/Morgan-Stanley/testplan/pull/363/files#diff-00338c67fec4ee430b4d3ed54f029c0dL42) / [MultitestIRunner](https://github.com/Morgan-Stanley/testplan/pull/363/files#diff-df53b985de90c945d5ec2afe61e9d8d4L44). But again I'd like @Pyifan to rule on whether or not these have a non-obvious use somewhere.

testplan/runnable/interactive/base.py
ryan-collingham: No, you've misunderstood I think.

If run_all_tests is called with await_results=False, then a single future object is returned (which represents the eventual result of running all the tests)

If await_results=True (the default), then nothing is returned - the report object is updated as a side-effect. run_test() will _not_ return a future object, since it is implicitly passed await_results=True (as the default value for that param). So there is never any future object to ignore.

Does that make sense?

testplan/runnable/interactive/base.py
ryan-collingham: what benefit would that give?

Yielding a list of testcase reports probably isn't very useful, if they all belong to different MultiTests / testsuites

We _could_ return the root self.report object here. However that would be misleading IMO, as it will include the results of all testcases that have been run, not just those run in the current run.

Overall, I think that returning a report _and also_ mutating the self.report as a side-effect is a potentially dangerous pattern. By returning a report, it sorta gives the impression of being a pure function that runs tests and returns the results of only those tests as a single report tree - but that's not what is actually happening. If there is no return type, it is obvious that there must be some side-effect, which in this case is mutating the report tree at self.report.

testplan/runnable/interactive/base.py
ryan-collingham: Yeah so previously my idea was to magically combine the test reports for only the tests that ran and returning them. But in the end I decided this wasn't really worth the effort or additional computation required to construct such reports, which most of the time will be ignored (at least when running with the web UI driving the API).

ryan-collingham: oh and having an optional return type is pretty common, in a strongly typed language you would say the return type is `Optional<Future<None>>`. In fact, python3 typing has an Optional, so this would directly translate to `Optional[Future]` (not sure if the Future object takes a template param for typing purposes).

testplan/runnable/interactive/base.py
morsedav: All perfectly clear.

[The previous version](https://github.com/Morgan-Stanley/testplan/blob/15abdad4670808c6054e385c6e628f16b80197f6/testplan/runnable/interactive/base.py#L160) returns a report object or a Future.

[Your version](https://github.com/Morgan-Stanley/testplan/blob/4804ec125dab9adc3f8d056c4d27ad29faec273b/testplan/runnable/interactive/base.py#L160) returns nothing or a Future that resolves to None.

The question posed is, why not just return the report or a Future that resolves to the report.

testplan/runnable/interactive/base.py
morsedav: It is a cost-free way to minimize API breakage.

ryan-collingham: There is no API breakage - this is brand new code I added just a week or so ago, nothing apart from the unit test I modified is using it.

testplan/runnable/interactive/base.py
morsedav: There's no reason that I can see to break the API. The old method never returns None.

morsedav: This is all relatively new code (this method added two months ago) and its usage throughout testplan is confined only to the interactive runner, which no other part of testplan depends on and which no client is using. So, my concerns about API breakage are moot.

testplan/runnable/interactive/base.py
morsedav: This is all relatively new code (this method added two months ago) and its usage throughout testplan is confined only to the interactive runner, which no other part of testplan depends on and which no client is using. So, my concerns about API breakage are moot.

testplan/runnable/interactive/base.py
morsedav: ^ this is new information to me and nullifies my concerns about API breakage.

morsedav:
@Pyifan I've changed my thoughts a bit on this. I was thinking that this code being changed must be intertwined throughout testplan and be very old (we're modifying `base.py` after all!) but in fact the oldest changed code is two months old and its usage is confined to interactive mode which no clients are using and no other part of testplan depends on.

The removed *IRunner classes are older but are not used anywhere and don't appear to be essential so I'm fine with those being gone.

Thus I'm inclined to approve this since its (a) not breaking anything (b) is something we need, and (c) it exists now today. I'm going to wait on approving til the morning to see if perhaps you could convince me its flawed in some crucial way else I'll approve it tomorrow.

Pyifan:
> > Not that I love the original design, but I believe the purpose of RunnableIRunner class family (MultiTestIRunner) provides a consistent interface between TestRunnerIHandler and Runnable (could be MultiTest, GTest and other different type of testing) class. If we remove this layer entirely, then looks like the logic of TestRunnerIHandler is hardwired with MultiTest directly. This might not be a good thing when we need to make interactive mode working for all other tests. Shall we come up with an all-way-rounded new design first?
> 
> So, the new interface between InteractiveRunner and any Test is based on two methods:
> 
> * dry_run() to generate an empty initial report skeleton
> * run_testcases_iter() to run testcases iteratively and yield testcase reports that the interactive runner can merge into its report tree
> 
> All Test runners already have a dry_run() method. All we need to do is implement run_testcases_iter() for another Test runner and it can work with interactive mode. Maybe I can add a NotImplemented version of run_testcases_iter() into the Test base class.

I think the change might work well for Multitest. But my major concern is it might not be feasible to implement such dry_run() method for other test types, also not all test types would allow fine control at testcase level. Also when we shift the report ownership from the runnable to interactive runner, what will the impact be. If some exception is thrown 

> > Not that I love the original design, but I believe the purpose of RunnableIRunner class family (MultiTestIRunner) provides a consistent interface between TestRunnerIHandler and Runnable (could be MultiTest, GTest and other different type of testing) class. If we remove this layer entirely, then looks like the logic of TestRunnerIHandler is hardwired with MultiTest directly. This might not be a good thing when we need to make interactive mode working for all other tests. Shall we come up with an all-way-rounded new design first?
> 
> So, the new interface between InteractiveRunner and any Test is based on two methods:
> 
> * dry_run() to generate an empty initial report skeleton
> * run_testcases_iter() to run testcases iteratively and yield testcase reports that the interactive runner can merge into its report tree
> 
> All Test runners already have a dry_run() method. All we need to do is implement run_testcases_iter() for another Test runner and it can work with interactive mode. Maybe I can add a NotImplemented version of run_testcases_iter() into the Test base class.

I've only seem MultiTest has a dry_run() implementation. I wonder if it is feasible for other test types to implement the same. Also would other test types all allow fine control down to testcase level (and yield testcase report) - some of the test frameworks we integrate just run as a whole and convert into a test level report - I think that is the reason why originally MultiTest interchanges test level report with interactive runner.

Even for MultiTest, the current model feels a bit simplistic to me. For example, if resource.start() throw exception, who should be responsible to log that exception in report. Or if user specified before_start/after_start/after_stop hooks, are we going to evoke them under interactive mode, and who takes care of that. I don't think we want to stuff interactive runner with all these runnable-specific details in future.

Pyifan:
> @Pyifan I've changed my thoughts a bit on this. I was thinking that this code being changed must be intertwined throughout testplan and be very old (we're modifying `base.py` after all!) but in fact the oldest changed code is two months old and its usage is confined to interactive mode which no clients are using and no other part of testplan depends on.
> 
> The removed *IRunner classes are older but are not used anywhere and don't appear to be essential so I'm fine with those being gone.
> 
> Thus I'm inclined to approve this since its (a) not breaking anything (b) is something we need, and (c) it exists now today. I'm going to wait on approving til the morning to see if perhaps you could convince me its flawed in some crucial way else I'll approve it tomorrow.

Lol! I mean it's fine if we start simple, but how can we tell what we 'thought' is complexity is not because other people has really think ahead.

PR #365, opened 2020-02-11
PyUnit fixups

ryan-collingham:
* Fix the default status to be PASSED instead of unknown when there are
  no failures.
* Add PyUnit and PyTest examples to docs

### DIFF
--- a/testplan/testing/pyunit.py
+++ b/testplan/testing/pyunit.py
@@ -4,6 +4,7 @@ from testplan.testing import base as testing
 from testplan.report import testing as report_testing
 from testplan.testing.multitest.entries import assertions
 from testplan.testing.multitest.entries import schemas
+from testplan.testing.multitest.entries import base as base_entries
 
 import unittest
 
@@ -16,7 +17,7 @@ class PyUnitConfig(testing.TestConfig):
 
     @classmethod
     def get_options(cls):
-        return {"suites": [unittest.suite.TestSuite]}
+        return {"suite": unittest.suite.TestSuite}
 
 
 class PyUnit(testing.Test):
@@ -35,6 +36,7 @@ class PyUnit(testing.Test):
     """
 
     CONFIG = PyUnitConfig
+    _TESTCASE_NAME = "PyUnit test results"
 
     def main_batch_steps(self):
         """Specify the test steps: run the tests, then log the results."""
@@ -43,22 +45,16 @@ class PyUnit(testing.Test):
 
     def run_tests(self):
         """Run PyUnit and wait for it to terminate."""
-        for pyunit_suite in self.cfg.suites:
-            suite_report = self._run_pyunit_suite(pyunit_suite)
-            self.result.report.append(suite_report)
-
-    def _run_pyunit_suite(self, pyunit_suite):
-        """
-        Run a single PyUnit testsuite and return the results as a suite report.
-        """
         suite_result = unittest.TestResult()
-        pyunit_suite.run(suite_result)
+        self.cfg.suite.run(suite_result)
 
         # Since we can't reliably inspect the individual testcases of a PyUnit
         # suite, we put all results into a single "testcase" report. This
         # will only list failures and errors and not give detail on individual
         # assertions like with MultiTest.
-        testcase_report = report_testing.TestCaseReport(name="PyUnit results")
+        testcase_report = report_testing.TestCaseReport(
+            name=self._TESTCASE_NAME
+        )
 
         for call, error in suite_result.errors:
             assertion_obj = assertions.RawAssertion(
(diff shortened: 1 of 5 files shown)


testplan/testing/pyunit.py
```diff
@@ -25,7 +25,7 @@ class PyUnit(testing.Test):
 
     :param name: Test instance name. Also used as uid.
     :type name: ``str``
-    :param suite: Description of test instance.
+    :param suite: PyUnit testsuite
```
ryan-collingham: oops, need to fix docstring

tests/unit/testplan/testing/test_pyunit.py
```diff
@@ -0,0 +1,40 @@
...
+def test_run_tests():
+    test_runner = pyunit.PyUnit(
+        name="My PyUnit",
+        description="PyUnit example testcase",
+        suites=[unittest.makeSuite(Passing)],
+        **PYUNIT_DEFAULT_PARAMS
+    )
+    result = test_runner.run()
+    report = result.report
+    assert report.passed
```
ryan-collingham: also need to add more testcases

testplan/testing/pyunit.py
```diff
@@ -36,42 +36,56 @@ class PyUnit(testing.Test):
...
                 description=str(call), content=str(error).strip(), passed=False
             )
             testcase_report.append(
                 schemas.base.registry.serialize(assertion_obj)
             )
-            self.result.report.entries.append(testcase_report)
+
+        # In case of no failures or errors we need to explicitly mark the
+        # testsuite as passed.
+        testcase_report.pass_if_empty()
```
Pyifan: I cannot seem to find the definition of this function.

In MultiTest, this is achieved by:
testcase_report.runtime_status = RuntimeStatus.FINISHED

But setting RuntimeStatus to FINISHED, report._status is set to PASS. We may want to implement the same thing for pyunit test as well, will be needed for interactive mode anyway.

Pyifan:
I see that the hierarchy mapping between pyunit and test report has changed. 

Originally

PyUnit <==> TestGroupReport (equivalent to Multitest)
    - it contains just 1 pyunit testsuite
pyunit testsuite <==> TestGroupReport (testsuite level)
    - pyunit testcase level (class definition) is flattened
pyunit call (function definition) <==> TestCaseReport

I think it make sense to allow multiple testsuites in PyUnit. But I don't quite understand why we need to change the mapping of pyunit call to assertion. I think the new mapping is like this:

PyUnit <==> TestGroupReport (equivalent to Multitest)
Here we add a dummy testsuite level 
pyunit testsuite <==> TestCaseReport
    - pyunit testcase level (class definition) is flattened
pyunit call (function definition) <==> Assertions

testplan/testing/pyunit.py
```diff
@@ -36,42 +36,56 @@ class PyUnit(testing.Test):
...
+
+        # Store the testcase reoport inside a testsuite report, and return it.
+        suite_report = report_testing.TestGroupReport(
```
Pyifan: Here we add a dummy testsuite level report, without much useful infomation.

testplan/testing/pyunit.py
```diff
@@ -36,42 +36,56 @@ class PyUnit(testing.Test):
...
-        )
-
     def main_batch_steps(self):
         """Specify the test steps: run the tests, then log the results."""
         self._add_step(self.run_tests)
         self._add_step(self.log_test_results)
 
     def run_tests(self):
         """Run PyUnit and wait for it to terminate."""
+        for pyunit_suite in self.cfg.suites:
```
Pyifan: when we have multiple suites, all of them will use self.cfg.name as the suite name, I think that would be a problem.

testplan/testing/pyunit.py
ryan-collingham: It is defined here: https://github.com/Morgan-Stanley/testplan/blob/master/testplan/report/testing/base.py#L974

Actually, we don't need to worry about setting runtime_status from a Test runner because that is handled by the interactive handler (and it is not used in batch mode).

I don't really like implicitly changing the report status to passed when updating the runtime status, I think that pass_if_empty() is more explicit.

ryan-collingham:
> I see that the hierarchy mapping between pyunit and test report has changed.
> 
> Originally
> 
> PyUnit <==> TestGroupReport (equivalent to Multitest)
> - it contains just 1 pyunit testsuite
> pyunit testsuite <==> TestGroupReport (testsuite level)
> - pyunit testcase level (class definition) is flattened
> pyunit call (function definition) <==> TestCaseReport
> 
> I think it make sense to allow multiple testsuites in PyUnit. But I don't quite understand why we need to change the mapping of pyunit call to assertion. I think the new mapping is like this:
> 
> PyUnit <==> TestGroupReport (equivalent to Multitest)
> - then we add dummy testsuite level reports for each pyunit testsuite
> pyunit testsuite <==> TestCaseReport
> - again pyunit testcase level (class definition) is flattened
> pyunit call (function definition) <==> Assertions
> 
> Did I understand it correct?

Yes there is a small (deliberate) change to the report sub-structure. The idea is that the testcase structure should always be the same, regardless of what passed/failed. This is an invariant expected by the interactive runner, and also by Insight. So the previous approach of dynamically appending new testcase entries based on what passed/failed would break interactive mode for PyUnit.

testplan/testing/pyunit.py
ryan-collingham: right, it is required to satisfy the Test -> testsuite -> testcase hierarchy expected in other places.

testplan/testing/pyunit.py
ryan-collingham: Good spot! I think maybe this needs some rethink with adding multiple suites.

testplan/testing/pyunit.py
ryan-collingham: I don't like adding in "dummy" report entries either, maybe I can change this.

ryan-collingham:
I've reverted the breaking API change for now, I think it's simpler to leave the behaviour as-is.

tests/unit/testplan/testing/test_pyunit.py
```diff
@@ -0,0 +1,88 @@
...
+    """
+
+    def test_asserts(self):
+        """Test with basic assertions."""
+        self.assertTrue(True)
+        self.assertFalse(False)
+
+
+@pytest.mark.skip("Example PyUnit testsuite")
+class Failing(unittest.TestCase):
```
morsedav: We're skipping these so pytest doesn't pick up on the `test_` prefix correct?

ryan-collingham: Right, otherwise PyTest would thing these were testcases it should run and then fail.

PR #366, opened 2020-02-12
Bump black version and update script

ryan-collingham:
Use latest version of black available internally. Update the utility
script to be able to run in both "conversion" and "checking" modes.


scripts/utils/black_wrapper.py
```diff
@@ -0,0 +1,60 @@
+#!/usr/bin/env python
```
morsedav: Why can't the command just be put in the train / travis config files? I'd rather not start / perpetuate a trend of making wrappers that wrap wrappers that wrap ...., makes it hard to figure out where a setting is actually really being set, especially for new people accustomed to out-of-the-box configurations.

scripts/utils/black_check.py
```diff
@@ -1,30 +0,0 @@
-#!/usr/bin/env python
-"""Run the black check on python 3, do nothing on python 2."""
```
morsedav: Also, understood that this wrapper existed in the original commit, I missed it when reviewing the 1000s of files changed. I would've raised the same objection wrt my other comment about wrappers.

scripts/utils/black_wrapper.py
ryan-collingham: So broadly I agree about not wanting to add pointless wrappers. The reason for having this wrapper originally was because Travis will run the check with both python 2 and 3 interpreters; this is a problem because black requires python 3 to run. So the wrapper handles this by checking if it is being run with a Python 2 interpreter, and if so does nothing to pass the check.

I'm no expert on Travis, so maybe there is a better way to add a check that only runs in the Python 3 environment. If you know such a way please let me know!

ryan-collingham: Another thing is that if we don't have a wrapper script and rely on people to run black directly, most likely they will forget to (say) exclude the vendored dependencies or node_modules or not set the correct line-length.

morsedav: Legit, but:
1. Yes give me a day to see if there's a non-wrapper way to ensure python3 in travis
2. --include / exclude [can be specified](https://github.com/psf/black/blob/383857273c79ad0bbe12f63716b4825de526a522/README.md#pyprojecttoml) in a [pyproject.toml](https://snarky.ca/clarifying-pep-518/) which is the [future](https://podbay.fm/podcast/1029487211/e/1541445300) and is meant to replace setup.cfg files, so we'd be well off adding this file for other reasons too.

ryan-collingham:
Resolved separately so closing.

PR #368, opened 2020-02-13
Default raise_on_timeout to True for timing.wait and friends

ryan-collingham:
From user feedback: it is surprising that raise_on_timeout defaults to False. I can't see any good reason it needs to, so let's change the default to True.

### DIFF
--- a/testplan/common/utils/timing.py
+++ b/testplan/common/utils/timing.py
@@ -129,7 +129,7 @@ def timeout(seconds, err_msg="Timeout after {} seconds."):
     return timeout_decorator
 
 
-def wait(predicate, timeout, interval=0.05, raise_on_timeout=False):
+def wait(predicate, timeout, interval=0.05, raise_on_timeout=True):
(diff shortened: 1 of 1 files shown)


morsedav:
seems dangerous...

testplan/common/utils/timing.py
```diff
@@ -196,13 +196,25 @@ def retry_until_timeout(
...
+    :type args: ``Optional[Iterable[Any]]``
+    :param kwargs: Keyword args to pass to ``item``
+    :type kwargs: ``Optional[Dict[str, Any]]``
+    :param interval: time to wait between successive call attempts, in seconds.
+    :type interval: ``int``
+    :param raise_on_timeout: Whether to raise a TimeoutException on timeout,
+        defaults to True.
     :return: Result of item.
-    :rtype: ``object``
+    :rtype: ``Any``
```
morsedav: actual types 👍

PR #369, opened 2020-02-13
Update black version + store black config in pyproject.toml

morsedav:
So this is one way to accomplish the goals of #366 without a wrapper, with some tradeoffs.

The upshot IMHO of this way is that it uses existing tools and introduces the pyproject.toml file into the project which ought to allow us to setup more sophisticated builds in the future. More info in these easy-to-read PEPs [here](https://www.python.org/dev/peps/pep-0517/) and [here](https://www.python.org/dev/peps/pep-0518/). And with a small amount of additional effort we could move the dependencies from both requirements files into the pyproject.toml.

One thing I'm not in love with here is specifying the version of `black` in both `.travis.yml` and `requirements.txt`. I'm pretty sure both can be consolidated into the pyproject.toml but I didn't know anything about pyproject.toml until a few hours ago so I'm not sure how to do this right now. The inelegant solution right now is what you see before you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,8 +10,36 @@ install:
   - python install-testplan-ui --verbose --dev
 
 script:
+  # Travis uses pyenv to manage a virtualenv / venv whose 'bin' dir is placed
+  # at the front of PATH. The version of Python depends on which version in the
+  # 'python' array is currently being tested.
+  #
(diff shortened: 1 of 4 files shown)


morsedav:
Note that the tests will fail because one file isn't blackened. This actually kinda proves that it works though that file will need to be blackened before anything's merged.

PR #374, opened 2020-02-21
WIP: Make the MultiTest before_start and after_start hooks default to…

morsedav:
… PASSED rather than UNKNOWN


morsedav:
This is in "draft" status since it needs a unit test where a multitest testplan uses one of these hooks and passes.

dobragab:
I encountered the same bug, it would be nice if it was fixed.

ryan-collingham:
Closing this WIP change for now as I have fixed the same issue in https://github.com/Morgan-Stanley/testplan/pull/386

PR #376, opened 2020-02-24
Set parent_uids correctly in interactive report

ryan-collingham:
When merging testcase entries into the report tree, ensure that the
parent_uids attribute is set correctly. There have been some strange UI
bugs because of this not being correct.

Also, when validating a PUT request, check that the UID of the updated
entry definitely matches with the current entry - changing a UID is not
allowed.

### DIFF
--- a/tests/unit/testplan/runnable/interactive/test_api.py
+++ b/tests/unit/testplan/runnable/interactive/test_api.py
@@ -27,14 +27,22 @@ def example_report():
                 uid="MTest1",
                 category=report.ReportCategories.MULTITEST,
                 env_status=entity.ResourceStatus.STOPPED,
+                parent_uids=["Interactive API Test"],
                 entries=[
                     report.TestGroupReport(
                         name="Suite1",
                         uid="MT1Suite1",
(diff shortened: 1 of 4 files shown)


tests/unit/testplan/runnable/interactive/test_api.py
```diff
@@ -476,6 +525,30 @@ def test_put(self, api_env):
...
+        """Test that 400 BAD REQUEST is returned for invalid PUT data."""
+        client, ihandler = api_env
+        api_url = (
+            "/api/v1/interactive/report/tests/MTest1/suites/MT1Suite1/"
+            "testcases/MT1S1TC2/parametrizations/MT1S1TC2_0"
+        )
+
+        # JSON body is required.
+        rsp = client.put(api_url)
+        assert rsp.status == "400 BAD REQUEST"
```
yuxuan-ms: why not check rsp.status_code?

ryan-collingham: No particular reason, I can change it.

PR #386, opened 2020-02-28
Fixes for running the ProcessPool example

ryan-collingham:
* Handle running pre/post stop/start callbacks interactively
* Mark after_start and friends' reports as passed if empty
* Skip unstable procpool UT on Windows
* Schedule tasks to local runner in interactive mode
* Remove handling of multiple runners from interactive handler
* Only log API access info at DEBUG level


ryan-collingham:
This PR fixes https://github.com/Morgan-Stanley/testplan/pull/386 and supersedes https://github.com/Morgan-Stanley/testplan/pull/386

PR #390, opened 2020-03-09
Fixes for python3.8

ryan-collingham:
- Use thread.is_alive() spelling to avoid deprecation warning
- cgi.escape is removed in Python3.8 so try to use html.escape in preference (introduced in Py3.2). Fall back to cgi.escape if not available (i.e. on Py2)

### DIFF
--- a/testplan/exporters/testing/pdf/renderers/entries/assertions.py
+++ b/testplan/exporters/testing/pdf/renderers/entries/assertions.py
@@ -1,5 +1,12 @@
 import re
-from cgi import escape
+import functools
+
+try:
+    from html import escape as _orig_escape
+
+    escape = functools.partial(_orig_escape, quote=False)
+except ImportError:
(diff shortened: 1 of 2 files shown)


testplan/exporters/testing/pdf/renderers/entries/assertions.py
```diff
@@ -1,5 +1,9 @@
 import re
-from cgi import escape
+
+try:
+    from html import escape
+except ImportError:
+    from cgi import escape
```
morsedav: What about this instead:

```py
try:
    from html import escape as _orig_escape
    from functools import partial
    escape = partial(_orig_escape, quote=False)
except ImportError:
    from cgi import escape
```

Reason being that we don't have any tests for how the PDF looks, and escaping quotes could have unforeseen consequences for how the PDF ends up looking.

PR #391, opened 2020-03-09
Setup env for adding proper routing

morsedav:
These are the small-ish changes I'll need before the larger changes I'm making to BatchReport.js can be merged.

### DIFF
--- a/testplan/web_ui/testing/public/index.html
+++ b/testplan/web_ui/testing/public/index.html
@@ -4,7 +4,6 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="theme-color" content="#000000">
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
     <link rel="shortcut icon" href="%PUBLIC_URL%/testplan-logo.png">
     <link href="//netdna.bootstrapcdn.com/font-awesome/4.1.0/css/font-awesome.min.css" rel="stylesheet">
     <title>Testplan</title>
--- a/testplan/web_ui/testing/src/Common/Home.jsx
+++ b/testplan/web_ui/testing/src/Common/Home.jsx
@@ -1,6 +1,10 @@
 import React from "react";
 import { Redirect } from "react-router-dom";
 
+const NoReport = () => (
+  <h1 style={{ color: "red" }}>NO TEST REPORT TO RENDER</h1>
+);
+
 /**
  * When NODE_ENV === 'development' we can choose an actual report to render
  * by setting the environment variable REACT_APP_REPORT_GUID_OVERRIDE
@@ -18,6 +22,7 @@ const DevHome = () => {
   } else {
     return (
       <>
+        <NoReport/>
         <h2 style={{ color: "green" }}>
           Set these environment variables before compiling:
           <ul>
@@ -38,6 +43,7 @@ const DevHome = () => {
 
 const ProdHome = () => (
     <>
+      <NoReport/>
       <h2>Navigate to the test report link printed at the end of your testplan run.</h2>
     </>
 );
@@ -46,7 +52,6 @@ const ProdHome = () => (
 export default function Home() {
   return (
     <>
-      <h1 style={{ color: "red" }}>NO TEST REPORT TO RENDER</h1>
       {process.env.NODE_ENV === "development" ? <DevHome /> : <ProdHome />}
     </>
   );
--- a/testplan/web_ui/testing/src/Common/Home.jsx
+++ b/testplan/web_ui/testing/src/Common/Home.jsx
@@ -1,0 +1,53 @@
+import React from "react";
+import { Redirect } from "react-router-dom";
+
+/**
+ * When NODE_ENV === 'development' we can choose an actual report to render
+ * by setting the environment variable REACT_APP_REPORT_GUID_OVERRIDE
+ * to an actual report GUID.
+ *
+ * Note that process.env.REACT_APP_* vars are elided during compilation. See
+ * node_modules/react-scripts/config/env.js:73.
+ */
+const DevHome = () => {
+  const REPORT_GUID_OVERRIDE = process.env.REACT_APP_REPORT_GUID_OVERRIDE;
+  if(typeof REPORT_GUID_OVERRIDE !== "undefined") {
+    const dest = `/testplan/${REPORT_GUID_OVERRIDE}`;
+    console.log(`REACT_APP_REPORT_GUID_OVERRIDE="${REPORT_GUID_OVERRIDE}" so redirecting to "${dest}"`);
+    return <Redirect to={dest} />;
+  } else {
+    return (
+      <>
+        <h2 style={{ color: "green" }}>
+          Set these environment variables before compiling:
+          <ul>
+            <li>REACT_APP_REPORT_GUID_OVERRIDE - GUID of an existing report</li>
+            <li>
+              REACT_APP_COUCHDB_HOST - Full 'scheme://host:port/path' of your couchdb server,
+              e.g. 'http://couch.example.com/devel'
+            </li>
+          </ul>
+          If your couchdb server has a different origin than your development server and doesn't have
+          CORS configured, you'll need to launch chrome with the '--disable-web-security' flag to
+          allow cross-origin requests.
+        </h2>
+      </>
+    );
+  }
+};
+
+const ProdHome = () => (
+    <>
+      <h2>Navigate to the test report link printed at the end of your testplan run.</h2>
+    </>
+);
+
+// TODO: Make a homepage that is unconditionally useful
+export default function Home() {
+  return (
+    <>
+      <h1 style={{ color: "red" }}>NO TEST REPORT TO RENDER</h1>
+      {process.env.NODE_ENV === "development" ? <DevHome /> : <ProdHome />}
+    </>
+  );
+}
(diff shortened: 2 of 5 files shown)


testplan/web_ui/testing/src/Common/Home.jsx
```diff
@@ -0,0 +1,58 @@
...
+  } else {
+    return (
+      <>
+        <NoReport/>
+        <h2 style={{ color: "green" }}>
+          Set these environment variables before compiling:
+          <ul>
+            <li>REACT_APP_REPORT_GUID_OVERRIDE - GUID of an existing report</li>
+            <li>
+              REACT_APP_COUCHDB_HOST - Full 'scheme://host:port/path' of your couchdb server,
```
ryan-collingham: Bear in mind that the UI is not only used for rendering reports from couchdb - it can also be used to render reports generated locally with the --ui flag and for the interactive mode. So I don't think there should be a requirement to specify a couchdb server, it should be possible to run completely locally.

ryan-collingham: Really I wouldn't mention couchdb at all here really - the UI is designed against an API provided by the web_app.py module, which abstracts away the couchdb API.

testplan/web_ui/testing/src/Report/BatchReport.js
```diff
@@ -64,7 +64,17 @@ class BatchReport extends React.Component {
         }),
         1500);
     } else {
-      axios.get(`/api/v1/reports/${uid}`)
+      const axiosConfig = {};
+      // allow overriding where we fetch the report from in deveopment, see src/Common/Home.jsx:12
+      if(process.env.NODE_ENV === "development" && typeof process.env.REACT_APP_COUCHDB_HOST !== "undefined") {
```
ryan-collingham: It's getting a little confusing how we have the branch above checking for the special _dev report UID, and now a different branch checking for development env variables. I'm not sure we need the special _dev branch any more really. But if you want to keep it, it might be worth splitting into  a couple of separate functions to cover the special _dev page and the regular report.

testplan/web_ui/testing/src/index.js
```diff
@@ -4,29 +4,45 @@ import BatchReport from './Report/BatchReport';
...
-import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
+import { BrowserRouter, Route, Switch, Redirect } from "react-router-dom";
 
 /**
  * This single App provides multiple functions controlled via the URL path
  * accessed. We are using React-Router to control which type of report is
  * rendered and to extract the report UID from the URL when necessary.
+ *
+ * We use "strict" here to ensure that paths end with a trailing slash. This
+ * makes the UI-rendered hash routes more distinguishable from the
```
ryan-collingham: I don't quite understand this comment - can you explain it to me in more detail? In particular I don't understand what you mean by "UI-rendered hash routes" and "server-rendered regular routes".

Also the comment doesn't quite seem to match with the code below - the redirect route is the first case in the switch not the last one?

testplan/web_ui/testing/public/index.html
```diff
@@ -4,7 +4,6 @@
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     <meta name="theme-color" content="#000000">
-    <link rel="manifest" href="%PUBLIC_URL%/manifest.json">
```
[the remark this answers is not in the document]
morsedav: It throws a large annoying exception in the browser console when the page opens. [We stopped generating the manifest.json](https://github.com/Morgan-Stanley/testplan/blob/014ed0c9f8dd117d47ad0cf426286113ff78c310/testplan/web_ui/testing/src/index.js#L33) 11 months ago so this tag's been doing nothing except polluting browser consoles ever since.

testplan/web_ui/testing/src/Common/Home.jsx
morsedav: Ya I'm fine with changing the name. This is mostly meant to ease our debugging, and AFAIK there isn't a decent way to debug users' reports without manually downloading their whole json and kludging it into the app in development mode somehow.

Regarding the /interactive route and running locally, there seems to be very little contact between the BatchReport UI and the InteractiveReport UI, aside from sharing some UI components. As they exist right now they can have independent logic behing-the-scenes (though I'm pretty sure this is/was on your TODO, to deduplicate the code between the two reports).

And, while I'm thinking of it, so maybe some future person browsing this PR can remind me, we ought to publish our CouchDB JSON schema so new users can get up and running quicker. We have a log of logic that seems quite specific to how we use it at MS so we should think about either simplifying our schema so it can be inferred by new users or just publishing what we use (or - better - do both).

testplan/web_ui/testing/src/Report/BatchReport.js
morsedav: I've actually set it up now to just understand `/whatever/route?dev=true`. React Router ignores all query params by default so this way is less disruptive.

In my next push I'll merge the `if (uid === "_dev")` into `?dev=true`.

testplan/web_ui/testing/src/index.js
morsedav: Ah yeah this made more sense before I split it into its own file. And the "UI-rendered has routes" doesn't make sense without seeing the new BatchReport. 

I'll fix the description but in short, we'll have this URL:

`/testplan/path1/xyz/?a=1&b=2#/MyTestPlan/My%20Suite/my_testcase?filter=passing`

`/testplan/path1/xyz/?a=1&b=2` <- this bit is sent over the net to our API
`/MyTestPlan/My%20Suite/my_testcase?filter=passing` <- this bit controls / reflects the UI state and is not sent over the net

testplan/web_ui/testing/src/Common/Home.jsx
ryan-collingham: Cool I think changing the name and comment not to mention couchdb would be great.

The batch and interactive UIs share the central assertion pane and most of the toolbar code too. The navigation bar code is different, to accommodate the extra interactive buttons, and the top level report components are significantly different to handle polling for updates from the interactive API instead of downloading the entire batch report as one big JSON. You're right that some of the navigation code could be refactored to avoid duplication in future.

I don't think we currently have a formal schema for the report JSON?

testplan/web_ui/testing/src/index.js
ryan-collingham: So you're suggesting we store the navigation state as a # fragment? I wasn't aware that was the design. Is there any reason we can't just set the URL to something like `/reports/<UID>/My%20Suite/my_testcase?filter=passing`? Seems simpler to me.

morsedav: Seems to me this would complicate things since the browser will navigate to the portion before the hash. Try it: `window.location.pathname = '/Morgan-Stanley/testplan/pull/390'`.

This is do-able but we'd need to keep the API in sync with the routes used in the UI so it knows to ignore some, plus the roundtrip overhead of the web request.

So, simpler how?

testplan/web_ui/testing/src/Common/Home.jsx
morsedav: No I don't think we do. Appears we've just added incrementally to get what we have currently.

I'd be interested in feedback from any non-MS user who has actually successfully setup a database backend of any sort with Testplan. Our logic in many places is very tailored to the JSON shape we use at present, so anyone looking to setup a database backend would need to just guess from the code what JSON shape the app is expecting. Or monkeypatch their app to use a different JSON shape. Not new-user-friendly ATM.

ryan-collingham: The JSON structure is defined by our testplan code when serializing the report tree, it does not depend on the exact database used to store the data.

testplan/web_ui/testing/src/index.js
ryan-collingham: It's just that is how the examples I've seen in the react-router work e.g. https://reacttraining.com/react-router/web/example/url-params , e.g. you click the "Netflix" link and it redirects to /netflix. I don't think it actually reloads the page in the browser, it's just JS manipulations.

Generally I've only seen # marks in URLs used to jump to a specific section on a page, not to redirect to a new page entirely. Is there a specific example / documention you've read which suggests to do the routing using # fragments?

morsedav: This is getting a disproportionate amount of attention. https://reacttraining.com/react-router/web/guides/primary-components

ryan-collingham: Thanks for the link, I wasn't aware of the HashRouter component, I've only ever used the BrowserRouter before.

So, to my undestanding your design is to continue using the BrowserRouter at the top level to distinguish between the batch and interactive modes, and to possible add some common home page or other pages (a search page?) in future. But then the HashRouter will be used to distinguish between navigation states within each UI. Is my understanding correct? This seems like an important design decision to me so I just want to make sure we get it right.

morsedav: On the BatchReport.js side at least, yes, using HashRouter for UI-only navigation seems the way to go. I think in the end we _may_ want to the same for the InteractiveReport.js side as well, though its benefits there are less certain.

For example with the BatchReport you'll be able to get your UI into a certain state (which right now just consists of what level of the report you're at, but could conceivably include user params like hide / show passing), then you or someone else could revisit that URL and restore that UI state, which I see this as the biggest benefit. The InteractiveReport needs the local server running and can't assume the server's state hasn't changed, so a link ( == testplan report + its UI state) can't be revisited in the same way. It would make the forward-back buttons work though.

morsedav:
Good dialog here but I'll close this in favor of #399

PR #392, opened 2020-03-09
Extendable list of listers that can be accessed from commandline

kn-ms:
I have refactored the test listers to be self contained so they provide their name and description, provided a listing_registry so testplan users can add their own listers. The registry can build the required enum, so not much change in the --info parameter parsing, just use the registry created enum.

and an extra: Fix unit tests on windows

- Kill one worker expect that there are 4 worker as child of the master.
  On windows if python running from venv, the interpreter is spawning the
  real interpreter as it's child, so the process tree is one layer deeper
  than expected so the recursive children count is double as expected

- sha1sum which bundled with git is printing a leading \ before the hash on windows, so prepare the regexp for that

> sha1sum C:\Users\..\AppData\Local\Temp\pytest-of-kn\pytest-1\test_hashfile0\hash_me.txt
\4a22175c7be0588626bcd19694f6572c9808597d *C:\\Users\\..\\AppData\\Local\\Temp\\pytest-of-..\\pytest-1\\test_hashfile0\\hash_me.txt

### DIFF
--- a/doc/en/multitest.rst
+++ b/doc/en/multitest.rst
@@ -276,10 +276,9 @@ and overriding ``get_output`` method.
 An example implementation of custom test lister can be seen
 :ref:`here <example_multitest_listing_custom>`.
 
-If the lister provide, name and description it can be used to extend the ``--info`` commandline
-parameter and the lister can be selected from the commandline.
+Listers can be registered to be used with the ``--info`` commandline parameter the same way as the built in listers.
 
-The class should provide:
+The custom lister class should provide:
 
 * it's name either setting the :py:attr:`NAME <testplan.testing.listing.BaseLister.NAME>` or override the
   :py:meth:`name() <testplan.testing.listing.BaseLister.name>` method. This should be an Enum name like ``NAME_FULL``.
@@ -302,7 +301,7 @@ and it need to be registered with :py:data:`testplan.testing.listing.listing_reg
     def get_output(self, instance):
         return "Hello World"
 
-  testing_registry.add_lister(HelloWorldLister())
+  listing_registry.add_lister(HelloWorldLister())
 
   # check --info hello-world
   @test_plan()
--- a/doc/en/multitest.rst
+++ b/doc/en/multitest.rst
@@ -305,7 +305,7 @@ and it need to be registered with :py:data:`testplan.testing.listing.listing_reg
   testing_registry.add_lister(HelloWorldLister())
 
   # check --info hello-world
-  @test_plan():
+  @test_plan()
   def main(plan):
     ....
 
--- a/doc/en/multitest.rst
+++ b/doc/en/multitest.rst
@@ -276,6 +276,40 @@ and overriding ``get_output`` method.
 An example implementation of custom test lister can be seen
 :ref:`here <example_multitest_listing_custom>`.
 
+If the lister provide, name and description it can be used to extend the ``--info`` commandline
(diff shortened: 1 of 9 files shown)


tests/unit/testplan/testing/test_listing.py
```diff
@@ -0,0 +1,58 @@
...
+
+
+def test_help_text():
+    registry = ListingRegistry()
+    registry.add_lister(CountLister())
+    registry.add_lister(ExpandedNameLister())
+
+    arg_enum = registry.to_arg()
+
+    help_text = arg_enum.get_help_text(None).split(os.linesep)
```
ryan-collingham: Unless get_help_text() is returning a raw `bytes` string, this should split on `\n` - even on Windows, `str` objects are separated by `\n` not `\r\n`.

doc/en/multitest.rst
```diff
@@ -276,6 +276,40 @@ and overriding ``get_output`` method.
...
 :ref:`here <example_multitest_listing_custom>`.
 
+If the lister provide, name and description it can be used to extend the ``--info`` commandline
+parameter and the lister can be selected from the commandline.
+
+The class should provide:
+
+* it's name either setting the :py:attr:`NAME <testplan.testing.listing.BaseLister.NAME>` or override the
+  :py:meth:`name() <testplan.testing.listing.BaseLister.name>` method. This should be an Enum name like ``NAME_FULL``.
+  The name will be used to derive the commandline param which is the kebab-case version of the name.
```
kelliott55: Typo: kebab?

doc/en/multitest.rst
```diff
@@ -276,6 +276,40 @@ and overriding ``get_output`` method.
...
+
+  class HelloWorldLister(BaseLister):
+
+    NAME = "HELLO_WORLD"
+    DESCRIPTION = "This lister print Hello World for each multitest"
+
+    def get_output(self, instance):
+        return "Hello World"
+
+  testing_registry.add_lister(HelloWorldLister())
```
kelliott55: Bug, should be listing_registry

doc/en/multitest.rst
```diff
@@ -276,6 +276,40 @@ and overriding ``get_output`` method.
 An example implementation of custom test lister can be seen
 :ref:`here <example_multitest_listing_custom>`.
 
+If the lister provide, name and description it can be used to extend the ``--info`` commandline
+parameter and the lister can be selected from the commandline.
```
kelliott55: This sentence is a little tricky to understand. Can you improve the wording?
"If the lister provide, ... "

kn-ms:
> Thank you for the PR. For our information what is your use-case behind needing the new functionality please?

I'm planning to implement a lister which can export information about the testplan(testsuite/testcases) in json/xml. I'm planning to run these testplans with this kind of metadata generation lister, for many projects and build a rich test repository, that can be used for various things (including reporting/review/test_selection)

doc/en/multitest.rst
kn-ms: do you mean 'kebab version' instead of 'kebab-case version'?

doc/en/multitest.rst
kn-ms: is this looking better?

> Listers can be registered to be used with the ``--info`` commandline parameter the same way as the built in listers.
> 
> The custom lister class should provide:

PR #393, opened 2020-03-12
Support GTest runner in interactive mode

ryan-collingham:
Add required methods to ProcessTestRunner base class to support running
interactively. Slightly modify report structure so that a ProcessCheck
suite and testcase are always included regardless of pass/fail. Render
RawAssertion content in UI.

### DIFF
--- a/testplan/testing/base.py
+++ b/testplan/testing/base.py
@@ -786,16 +786,17 @@ class ProcessRunnerTest(Test):
 
     def run_testcases_iter(self, testsuite_pattern="*", testcase_pattern="*"):
         """
-        For a Test to be run interactively, it must implement this method.
+        Runs testcases as defined by the given filter patterns and yields
+        testcase reports. A single testcase report is made for general checks
+        of the test process, including checking the exit code and loggin stdout
+        and stderr of the process. Then, testcase reports are generated from
+        the output of the test process.
 
-        It is expected to run tests iteratively and yield a tuple containing
-        a testcase report and the list of parent UIDs required to merge the
-        testcase report into the main report tree.
-
-        If it is not possible or very inefficient to run individual testcases
-        in an iteratie manner, this method may instead run all the testcases
-        in a batch and then return an iterator for the testcase reports and
-        parent UIDs.
+        For efficiency, we run all testcases in a single subprocess rather than
+        running each testcase in a seperate process. This reduces the total
+        time taken to run all testcases, however it will mean that testcase
+        reports will not be generated until all testcases have finished
+        running.
 
         :param testsuite_pattern: Filter pattern for testsuite level.
         :type testsuite_pattern: ``str``
@@ -838,6 +839,6 @@ class ProcessRunnerTest(Test):
     def test_command_filter(self, testsuite_pattern, testcase_pattern):
         """
         Return the base test command with additional filtering to run a
-        specific set of testcases.
+        specific set of testcases. To be implemented by concrete subclasses.
         """
         raise NotImplementedError
--- a/testplan/testing/base.py
+++ b/testplan/testing/base.py
@@ -3,6 +3,7 @@ import os
 import sys
 import subprocess
 import six
+import tempfile
 
 from lxml import objectify
 from schema import Or, Use, And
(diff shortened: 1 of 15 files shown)


testplan/testing/base.py
```diff
@@ -628,74 +657,84 @@ def process_test_data(self, test_data):
...
-        if self._test_process_killed or not self._test_has_run:
-            self.result.report.append(self.get_process_failure_report())
-            return
-
-        if len(self.result.report):
-            raise ValueError(
-                "Cannot update test report,"
-                " it already has children: {}".format(self.result.report)
-            )
+        with open(self.stdout) as stdout, open(self.stderr) as stderr:
```
morsedav: Could `stdout` / `stderr` not be a filename? Like, could it ever be a `io.StringIO`?

ryan-collingham: No I don't think so. See their definition: https://github.com/Morgan-Stanley/testplan/blob/b2725e2f2761e4991558faa50d375032cbf03ba0/testplan/testing/base.py#L444-L450

testplan/testing/base.py
```diff
@@ -722,3 +761,83 @@ def aborting(self):
...
+        )
+        testsuite_report.append(testcase_report)
+        report.append(testsuite_report)
+
+        return result
+
+    def run_testcases_iter(self, testsuite_pattern="*", testcase_pattern="*"):
+        """
+        For a Test to be run interactively, it must implement this method.
+
```
morsedav: I have misgivings about this. The docstring suggests this should be decorated `@abc.abstractmethod`, which means the class needs to use metaclass `abc.ABCMeta`, which may not be something you want to do in this preexisting class.

But since overriding this method is required for running `ProcessRunnerTest`-derived tests in interactive mode, how about making this more obvious by moving it to a new class: 

```py
class InteractiveProcessRunnerTest(ProcessRunnerTest, metaclass=abc.ABCMeta):
    @abc.abstractmethod
    def run_testcases_iter(self, testsuite_pattern="*", testcase_pattern="*"):
        ...
```

That, or write this in such a way that the implementor of a class doesn't need to know to override this method (i.e. it will "just work" for whatever subclass derives from it).

Else I can imagine a future test-writer deriving from `ProcessRunnerTest` > trying to run it in interactive mode > it doesn't work > test-writer does not read the docstrings of all `ProcessRunnerTest` methods > test-writer stays frustrated and confused.

ryan-collingham: To be clear, run_testcases_iter is an abstract method defined in the Test base class, this here is the concrete implementation. Maybe I should update the docstring to make that clearer.

I agree we should refactor Test and other abstract base classes to use the ABC meta-class, but that is not really in the scope of what this PR is trying to address.

I don't personally see the advantage of creating yet another base class specifically for interactive usage.

morsedav: I see, yes the docstring makes it seem like it must be overridden by any derived classes.

And agreed we way too many base classes as it is, but if something different must be done in order for the test to be run interactively I think it's worth it to make a new class, i.e. the benefits of making it obvious that there's additional work to be done by the implementor outweighs the burden of dealing with another base class.

morsedav: Since the goal I'm trying to get at is to make it obvious to the implementor that they need to do something extra in order to get their test to run interactively, a alternative to a new class could me an error message (either on the Python side or in the UI) that says something like "Cannot run interactively - implement the method `run_testcases_iter` in <insert name of class missing `run_testcases_iter`>. For an example implementation, see `testplan.testing.base.ProcessRunnerTest.run_testcases_iter`."

That is, if it absolutely must be overridden by a derived class.

ryan-collingham: I agree in general that we should use the abc module to mark abstract classes and methods. However that applies to many many base classes, not just this ProcessTestRunner class. So though I think the change you're proposing is outside the scope of this PR, I think we should raise an issue to address it in future to all base classes that have abstract methods, not just this particular one.

testplan/testing/base.py
```diff
@@ -838,6 +839,6 @@ def run_testcases_iter(self, testsuite_pattern="*", testcase_pattern="*"):
     def test_command_filter(self, testsuite_pattern, testcase_pattern):
         """
         Return the base test command with additional filtering to run a
-        specific set of testcases.
+        specific set of testcases. To be implemented by concrete subclasses.
         """
         raise NotImplementedError
```
morsedav: This is in line with what's present in other places in this codebase, but just for the record methods like this ought to be marked `abstractmethod` since normal IDEs will highlight to the user that they need to implement the method.

That however will require a broad refactor of several base classes and so is outside the scope of this PR.

testplan/testing/base.py
ryan-collingham: Here you go, I've raised an issue to address: https://github.com/Morgan-Stanley/testplan/issues/394

PR #398, opened 2020-03-19
Add flag parameter to result.log() and result.fail() to include message in email exporter

Pyifan:
For email exporter, we want to allow user to selectively include some detailed information for each testcase. The proposed solution is to include result.fail(flag="email") and result.log(flag="email") assertions as they are both single-message assertions with no further detail and well fit email exporter usage

Also added is a compound assertions result.conditional_log() as a shortcut for user to write logic such as:

if condition:
    result.log()
else:
    result.fail()

### DIFF
--- a/examples/Assertions/Basic/test_plan.py
+++ b/examples/Assertions/Basic/test_plan.py
@@ -60,8 +60,7 @@ class SampleSuite(object):
         result.true("foo" == "foo", description="Boolean Truthiness check")
         result.false(5 < 2, description="Boolean Falseness check")
 
-        result.passing("This is a passing assertion")
-        result.failing("This is an explicit failure.")
+        result.fail("This is an explicit failure.")
 
         # Membership checks
         result.contain("foo", "foobar", description="Passing membership")
--- a/examples/Assertions/Basic/test_plan.py
+++ b/examples/Assertions/Basic/test_plan.py
@@ -60,7 +60,8 @@ class SampleSuite(object):
         result.true("foo" == "foo", description="Boolean Truthiness check")
         result.false(5 < 2, description="Boolean Falseness check")
 
-        result.fail("This is an explicit failure.")
+        result.passing("This is a passing assertion")
+        result.failing("This is an explicit failure.")
 
         # Membership checks
         result.contain("foo", "foobar", description="Passing membership")
--- a/testplan/runnable/base.py
+++ b/testplan/runnable/base.py
@@ -214,7 +214,7 @@ class TestRunnerResult(RunnableResult):
 
 
 class TestRunner(Runnable):
-    r"""
+    """
     Adds tests to test
     :py:class:`executor <testplan.runners.base.Executor>` resources
     and invoke report
--- a/testplan/runnable/base.py
+++ b/testplan/runnable/base.py
@@ -214,7 +214,7 @@ class TestRunnerResult(RunnableResult):
 
 
 class TestRunner(Runnable):
-    """
+    r"""
     Adds tests to test
     :py:class:`executor <testplan.runners.base.Executor>` resources
     and invoke report
--- a/testplan/testing/multitest/result.py
+++ b/testplan/testing/multitest/result.py
@@ -93,7 +93,7 @@ def _bind_entry(entry, result_obj):
     ``entries`` list.
     """
     # Second element is the caller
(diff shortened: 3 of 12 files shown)


testplan/runnable/base.py
```diff
@@ -214,7 +214,7 @@ def success(self):
 
 
 class TestRunner(Runnable):
-    r"""
+    """
```
ryan-collingham: Why have you made this change? Please revert - the `r"""` prefix is required so that the literal `\`s in the docstring aren't treated as escape sequences, which trigger warnings on Python 3.6+.

testplan/testing/multitest/result.py
```diff
@@ -1404,10 +1404,15 @@ def log(self, message, description=None):
...
         :param description: Text description for the assertion.
         :type description: ``str``
+        :param category: Custom category for the assertion, set to 'email' to
+            include message in email exporter.
+        :type category: ``str``
         :return: ``True``
         :rtype: ``bool``
         """
-        entry = base.Log(message=message, description=description)
+        entry = base.Log(message=message,
```
ryan-collingham: You forgot to run black!

If you are using PyCharm, you can set it up as an external tool to run on every save, it works pretty well: https://black.readthedocs.io/en/stable/editor_integration.html . There are integrations for other editors listed on that page too. Alternatively, you can use the pre-commit hook I added to the internal repo - just copy script/pre-commit to your .git/hooks directory in the root of the OSS repo.

examples/Assertions/Basic/test_plan.py
```diff
@@ -60,7 +60,8 @@ def test_basic_assertions(self, env, result):
         result.true("foo" == "foo", description="Boolean Truthiness check")
         result.false(5 < 2, description="Boolean Falseness check")
 
-        result.fail("This is an explicit failure.")
+        result.passing("This is a passing assertion")
+        result.failing("This is an explicit failure.")
```
ryan-collingham: I'm not a big fan of this rename from fail -> failing. It's not really consistent with the rest of the existing assertions - most other assertions are verbs (equal, match, log etc.) not nouns. I think that result.fail() reads better too. Was there a specific reason you made this change or just personal preference?

testplan/testing/multitest/result.py
```diff
@@ -1404,27 +1404,47 @@ def log(self, message, description=None, category=None):
...
-        :param category: Custom category for the assertion, set to 'email' to
-            include message in email exporter.
+        :return: ``True``
+        :rtype: ``bool``
+        """
+        entry = base.Log(message=message, description=description)
+        _bind_entry(entry, self)
+        return entry
+
+    def passing(self, description, category=None):
```
ryan-collingham: I don't really see how this new `result.passing` is different from `result.log`? I don't think we need it, it's confusing.

testplan/testing/multitest/result.py
```diff
@@ -1437,6 +1457,51 @@ def fail(self, description, category=None):
...
+        :param failing_description: Description if condition evaluates to False.
+        :type failing_description: ``str``
+        :param category: Custom category that will be used for summarization.
+        :type category: ``str``
+        :return: ``True``
+        :rtype: ``bool``
+        """
+        if condition:
+            if passing_description:
+                return self.passing(passing_description, category=category)
```
ryan-collingham: could just as easily call `result.log` here.

examples/Assertions/Basic/test_plan.py
ryan-collingham: Another thing, this rename will be a breaking change for any existing users of `result.fail`. Is it really worth breaking user's tests for?

testplan/testing/multitest/result.py
```diff
@@ -1404,103 +1404,92 @@ def log(self, message, description=None):
...
-        :type category: ``str``
-        :return: ``True``
-        :rtype: ``bool``
-        """
-        entry = assertions.Pass(description, category=category)
-        _bind_entry(entry, self)
-        return entry
-
-    def failing(self, description, category=None):
+    def fail(self, description, category=None, flag=None):
```
ryan-collingham: It's a little confusing how we now have both "category" and "flag" parameters here, but I guess the docstring explains the difference so it's not that big a deal.

Issue #402, opened 2020-03-20
Interactive reset button should stop environments


ryan-collingham:
Currently, the interactive mode has a reset button which resets the results of all testcases. However, it does not yet stop any started environments, which may be unexpected.

PR #403, opened 2020-03-20
Reset environment state before testcase state

ryan-collingham:
Also add some extra debug logging + error reporting in UI.

Fixes https://github.com/Morgan-Stanley/testplan/issues/402


ryan-collingham:
Not sure what's going on with the Travis build exactly - it's failing to run the npm build. Works fine for me locally. I'll take another look on monday.

PR #404, opened 2020-03-24
PyTest tests can be run in interactive UI

ryan-collingham:
Add support for PyTest in interactive mode by implementing the
required methods on the Pytest test runner class. Add unit test.

### DIFF
--- a/tests/unit/testplan/testing/test_pytest.py
+++ b/tests/unit/testplan/testing/test_pytest.py
@@ -9,99 +9,26 @@ from testplan.testing import py_test as pytest_runner
 import testplan.report
 from testplan import defaults
 
+from tests.unit.testplan.testing import pytest_expected_data
+
 
 @pytest.fixture
-def pytest_test_inst():
+def pytest_test_inst(repo_root_path):
     """Return a PyTest test instance, with the example tests as its target."""
-    # This file is at tests/unit/testplan/testing/test_pytest.py, we want to run
-    # the pytest example at examples/PyTest/pytest_tests.py.
-    root_path = os.path.join(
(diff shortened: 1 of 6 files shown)


tests/unit/testplan/testing/test_pytest.py
```diff
@@ -0,0 +1,213 @@
...
+        root_path, "examples", "PyTest", "pytest_tests.py"
+    )
+    return pytest_runner.PyTest(
+        name="pytest example",
+        target=example_path,
+        stdout_style=defaults.STDOUT_STYLE,
+    )
+
+
+_EXPECTED_DRY_RUN_REPORT = testplan.report.TestGroupReport(
```
morsedav: I have a tiny non-blocking preference that `_EXPECTED_DRY_RUN_REPORT` live in a different file so readers' concentration isn't broken when studying this file top to bottom.

ryan-collingham: seems reasonable, I'll move it out.

tests/unit/testplan/testing/test_pytest.py
```diff
@@ -0,0 +1,213 @@
...
+from testplan import defaults
+
+
+@pytest.fixture
+def pytest_test_inst():
+    """Return a PyTest test instance, with the example tests as its target."""
+    # This file is at tests/unit/testplan/testing/test_pytest.py, we want to run
+    # the pytest example at examples/PyTest/pytest_tests.py.
+    root_path = os.path.join(
+        os.path.dirname(__file__), *(os.pardir for _ in range(4))
```
morsedav: Clever, but why build the relative path? And this will break if this file's moved to a higher / lower directory, yes?

PR #406, opened 2020-03-26
Display execution time for each entry in navigator

raoyitao:
* On web UI the left pane is used for navigating among multitests,
  suites and testcases. Adding an option to display the execution
  time of each entry.

### DIFF
--- a/testplan/web_ui/testing/src/Nav/NavBreadcrumbs.js
+++ b/testplan/web_ui/testing/src/Nav/NavBreadcrumbs.js
@@ -37,7 +37,10 @@ return (
         status={entry.status}
         type={entry.category}
         caseCountPassed={entry.counter.passed}
-        caseCountFailed={entry.counter.failed} />
+        caseCountFailed={entry.counter.failed}
+        executionTime={null}
+        displayTime={false}
+      />
     </div>
   </li>
 );});
--- a/testplan/web_ui/testing/src/Nav/NavEntry.js
+++ b/testplan/web_ui/testing/src/Nav/NavEntry.js
@@ -22,6 +22,15 @@ import {
  */
 const NavEntry = (props) => {
   const badgeStyle = `${STATUS_CATEGORY[props.status]}Badge`;
+  const executionTime = (
+    props.displayTime && props.executionTime ? (
+      <i className={css(styles.entryIcon)} title='Execution time'>
+        <span className={css(styles[STATUS_CATEGORY[props.status]])}>
+          {props.executionTime}s
+        </span>
+      </i>
+    ) : null
+  );
   return (
     <div
       className='d-flex justify-content-between align-items-center'
@@ -34,6 +43,7 @@ const NavEntry = (props) => {
         {props.name}
       </div>
       <div className={css(styles.entryIcons)}>
+        {executionTime}
         <i className={css(styles.entryIcon)} title='passed/failed testcases'>
           <span className={css(styles.passed)}>{props.caseCountPassed}</span>
           /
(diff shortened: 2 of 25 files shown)


testplan/web_ui/testing/src/Nav/NavEntry.js
```diff
@@ -22,6 +22,16 @@ import {
  */
 const NavEntry = (props) => {
   const badgeStyle = `${STATUS_CATEGORY[props.status]}Badge`;
+  const executionTime = (
+    props.displayTime && props.executionTime ? (
+      <i className={css(styles.entryIcon)} title='Execution time'>
+        <span className={css(styles[STATUS_CATEGORY[props.status]])}>
+          ({((new Date(props.executionTime.end)).getTime()
```
ryan-collingham: I think it would be neater to calculate the timedelta separately and store in a variable before building the JSX content.

testplan/web_ui/testing/src/Report/InteractiveReport.js
```diff
@@ -497,6 +497,7 @@ class InteractiveReport extends React.Component {
           updateFilterFunc={noop}
           updateEmptyDisplayFunc={noop}
           updateTagsDisplayFunc={noop}
+          updateTimeDisplayFunc={noop}
```
ryan-collingham: Can we make sure that the button to toggle time display is not displayed in the interactive UI?

Ideally, I think the time display toggle would be provided as a one of the extraButtons passed from the BatchReport (see code immediately below this comment for example of extraButtons usage). That way, only buttons that are common to both batch and interactive report UIs are stored in the toolbar component itself.

testplan/web_ui/testing/src/Toolbar/Toolbar.js
```diff
@@ -282,6 +310,7 @@ class Toolbar extends Component {
             {this.filterButton(toolbarStyle)}
             {this.printButton()}
             {this.tagsButton()}
+            {this.timeButton()}
```
ryan-collingham: As mentioned above, let's move this timeButton out of the toolbar and make it an extraButton passed in by the BatchReport only - it doesn't make sense to have a button that does nothing in interactive UI.

testplan/web_ui/testing/src/Nav/NavBreadcrumbs.js
```diff
@@ -37,7 +37,10 @@ return (
         status={entry.status}
         type={entry.category}
         caseCountPassed={entry.counter.passed}
-        caseCountFailed={entry.counter.failed} />
+        caseCountFailed={entry.counter.failed}
+        executionTime={null}
```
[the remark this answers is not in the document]
raoyitao: yes in breadcrumbs navigation tab the this info should not be displayed, or the UI looks ugly.

testplan/web_ui/testing/src/Nav/NavEntry.js
raoyitao: I moved it to the outside in a function call.

testplan/web_ui/testing/src/Report/InteractiveReport.js
raoyitao: I remove it from interactive UI, place it in extraButtons.

testplan/web_ui/testing/src/Toolbar/Toolbar.js
raoyitao: I remove it from interactive UI, place it in extraButtons.

PR #410, opened 2020-03-27
Testcase logger

kn-ms:
Two mixins to test suite that enable 

- logging in the suite through testplan namespace, captured to testplan log
- logs can be auto captured in result as log assertion or attachments
- can capture  with context manager per testcase or for all case in a suite with minimal boilerplate

see example:  Multitest/Logging

### DIFF
--- a/doc/en/multitest.rst
+++ b/doc/en/multitest.rst
@@ -1331,7 +1331,7 @@ Python standard logging infrastructure can be used for logging, however testplan
 
 See also the :ref:`a downloadable example <example_multitest_logging>`.
 
-:py:class:`LogCaptureMixin <testplan.testing.multitest.logging.LogCaptureMixin>` when inherited provide a ``self.logger`` which will log to the normal testplan log. Furthermore the mixin provide a context manager :py:meth:`capture_log(result) <testplan.testing.multitest.logging.LogCaptureMixin.capture_log>` which can be used to automatically capture logs happening in teh context and attaching it to the result.
+:py:class:`LogCaptureMixin <testplan.testing.multitest.logging.LogCaptureMixin>` when inherited provide a ``self.logger`` which will log to the normal testplan log. Furthermore the mixin provide a context manager :py:meth:`capture_log(result) <testplan.testing.multitest.logging.LogCaptureMixin.capture_log>` which can be used to automatically capture logs happening in the context and attaching it to the result.
 
 .. code-block:: python
 
--- a/doc/en/multitest.rst
+++ b/doc/en/multitest.rst
@@ -1323,3 +1323,54 @@ If a test is expect to fail all the time, you can also use the `strict=True` the
     @xfail(reason='api changes', strict=True)
     def fail_testcase(self, env, result):
         ...
+
+Logging
+-------
+
+Python standard logging infrastructure can be used for logging, however testplan provide mixins to use a conveniently configured logger from testcases.
+
+See also the :ref:`a downloadable example <example_multitest_logging>`.
+
+:py:class:`LogCaptureMixin <testplan.testing.multitest.logging.LogCaptureMixin>` when inherited provide a ``self.logger`` which will log to the normal testplan log. Furthermore the mixin provide a context manager :py:meth:`capture_log(result) <testplan.testing.multitest.logging.LogCaptureMixin.capture_log>` which can be used to automatically capture logs happening in teh context and attaching it to the result.
+
+.. code-block:: python
+
+    @testsuite
+    class LoggingSuite(LogCaptureMixin):
+
+        @testcase
+        def testsuite_level(self, env, result):
+            with self.capture_log(
+                result
+            ) as logger:  # as convenience the logger is returned but is is really the same as self.logger
+                logger.info("Hello")
+                self.logger.info("Logged as well")
+
+The code above will capture the two log line and inject it into the result. The capture can be configured to capture log to a file and attach to the result. It also possible to capture the base testplan logger, or even the root logger during the execution of the context. If the default formatting is not good enough it can be changed for the report. For all these options see :py:meth:`LogCaptureMixin.capture_log <testplan.testing.multitest.logging.LogCaptureMixin.capture_log>`
+
+:py:class:`AutoLogCaptureMixin <testplan.testing.multitest.logging.AutoLogCaptureMixin>` when inherited it automatically capture and insert logs to the result for every testcase.
+
+.. code-block:: python
+
+    @testsuite
+    class AutoLoggingSuite(AutoLogCaptureMixin):
+        """
+        AutoLogCaptureMixin will automatically add captured log at the end of all testcase
+        """
+
+        @testcase
+        def case(self, env, result):
+            self.logger.info("Hello")
+
+        @testcase
+        def case2(self, env, result):
+            self.logger.info("Do it for all the testcases")
+
+The capture can be tweaked to set up ``self.log_capture_config`` during construction time, very similar to the :py:meth:`LogCaptureMixin.capture_log <testplan.testing.multitest.logging.LogCaptureMixin.capture_log>`
+
+.. code-block:: python
+
+    def __init__(self):
+        super(AutoLoggingSuiteThatAttach, self).__init__()
+        self.log_capture_config.attach_log = True
+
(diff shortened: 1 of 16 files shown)


Pyifan:
> (This is WIP, I'm still working on the documentation but would like to hear some feedback on the concept)
> 
> Two mixins to test suite that enable
> 
> * logging in the suite through testplan namespace, captured to testplan log
> * logs can be auto captured in result as log assertion or attachments
> * can capture  with context manager per testcase or for all case in a suite with minimal boilerplate
> 
> see example: Multitest/Logging

Hi Krisztian,

I like the idea a lot! I have seen testplan user creating a log file himself to capture some runtime info that doesn't belong to the report but still could be useful. It's a good idea to allow them to log directly to testplan.log from testsuite/testcase. Maybe we want to attach the entire log file in the report as well?

doc/en/multitest.rst
```diff
@@ -1323,3 +1323,54 @@ If a test is expect to fail all the time, you can also use the `strict=True` the
...
+
+Logging
+-------
+
+Python standard logging infrastructure can be used for logging, however testplan provide mixins to use a conveniently configured logger from testcases.
+
+See also the :ref:`a downloadable example <example_multitest_logging>`.
+
+:py:class:`LogCaptureMixin <testplan.testing.multitest.logging.LogCaptureMixin>` when inherited provide a ``self.logger`` which will log to the normal testplan log. Furthermore the mixin provide a context manager :py:meth:`capture_log(result) <testplan.testing.multitest.logging.LogCaptureMixin.capture_log>` which can be used to automatically capture logs happening in teh context and attaching it to the result.
+
```
Pyifan, line 1335: typo teh-> the

testplan/testing/multitest/base.py
```diff
@@ -804,7 +804,7 @@ def _run_suite_related(self, testsuite, method):
         return method_report
 
     def _run_case_related(self, method, testcase, case_result):
-        interface.check_signature(method, ["name", "self", "env", "result"])
```
Pyifan, line 807: i think the pre_testcase funcs we are using in our tests/examples do has a signature of name, self, env, result.

testplan/testing/multitest/base.py
```diff
@@ -804,7 +804,7 @@ def _run_suite_related(self, testsuite, method):
...
+        interface.check_signature(method, ["self", "name", "env", "result"])
         method(testcase.__name__, self.resources, case_result)
```
Pyifan, line 808: but we didn't pass self here, should be a problem.

testplan/testing/multitest/logging.py
```diff
@@ -0,0 +1,166 @@
...
+        self.capture_level = CaptureLevel.TESTSUITE
+        self.attach_log = False
+        self.format = _LOGFILE_FORMAT
+
+
+class LogCaptureMixin(Loggable):
+    """ Mixin to add easy logging support to any @multitest.testsuite
+
+    """
+
```
Pyifan, line 57: hmm, i thought black would complain about this docstring...

tests/unit/testplan/web_ui/testing/test_jest.py
```diff
@@ -46,4 +46,4 @@ def test_testplan_ui():
 )
 def test_eslint():
     """Run eslint over the UI source code."""
-    subprocess.check_call(["npm", "run", "lint"], cwd=TESTPLAN_UI_DIR)
```
Pyifan, line 49: I wonder why is this change?

testplan/testing/multitest/logging.py
kn-ms: I will fix this to a proper one liner

testplan/testing/multitest/base.py
kn-ms: _run_case_related is always used with the per_testcase and post_testcase methods, they are always looked up from the testsuite instence
```python
        pre_testcase = getattr(testsuite, "pre_testcase", None)
        post_testcase = getattr(testsuite, "post_testcase", None)
```
which is a bound method, so running it will implicitly get the instance (aka: self) as the first argument. so no need to pass.

Please note, that this is when the testsuite class has a `pre_testcase` or `post_testcase` method, which is not much documented and is different from the `@pre_tescase` `@post_testcase` decorator for a testsuite, which expect not bound functions, that taking `"name", "self", "env", "result"` 

```python
   @wraps(testcase_method)
    def testcase_with_pre(*args, **kwargs):
        """
        Testcase with prelude
        """
        for prelude in preludes:
            prelude(
                testcase_method.__name__,
                *args,
                **_get_kwargs(kwargs, testcase_method)
            )
        return testcase_method(*args, **kwargs)
```

None of the unittests testing the member function way of doing this, that is why the above signature check never found.

testplan/testing/multitest/base.py
kn-ms: Only the decorator `@pre_testcase` and `@post_testcase` is used in tests, I do not see examples having this yet.

doc/en/multitest.rst
kn-ms: Will fix this

tests/unit/testplan/web_ui/testing/test_jest.py
kn-ms: Without `shell=True` on my windows machine the testcase does not pick up PATH and do not found `npm`, however when `shell=True` the behaviour of the popen changes a bit, best describe in [this article](https://medium.com/python-pandemonium/a-trap-of-shell-true-in-the-subprocess-module-6db7fc66cdfd).  This looks to be a sweet spot for windows and linux, as well as it is in sync with the test_testplan_ui method just above.

ryanc414: I saw the same error running on Windows. I didn't think to change it because it is just as easy to run the linter directly as `npm run lint`; this pytest wrapper is mostly here just to ensure that the linter is run as part of the Travis build.

It looks like the error occurs because, while on Linux `npm` is the name of the executable, on Windows the executable is `npm.CMD` and `npm` is just a kind of shell command for convenience. So when not setting shell=True, python cannot find the `npm` executable (because it doesn't exist). I found a good SO question which explains it better here: https://stackoverflow.com/questions/3022013/windows-cant-find-the-file-on-subprocess-call/43580381

testplan/testing/multitest/logging.py
```diff
@@ -0,0 +1,166 @@
...
+    def select_logger(self, capture_level):
+        return capture_level(self)
+
+
+class AutoLogCaptureMixin(LogCaptureMixin):
+    def __init__(self):
+        super(AutoLogCaptureMixin, self).__init__()
+        self._state = None
+
```
Pyifan, line 159: I found that there are some discrepancy in the current code about how pre_testcase shall be defined. One way is via @pre_testcase decorator - that's way in examples and tests. But then defining pre_testcase method in testsuite also kind of works, but in a slightly different way. We will need to consolidate this, but this shouldn't block your PR.

testplan/testing/multitest/base.py
Pyifan: I see, i think your change is correct and no wonder it didn't break any test - because def pre_testcase way of defining pre_testcase is never used/tested.

Pyifan:
I think overall this looks really good! (Except for I confused myself with this pre_testcase thing for 
 a while). I just have one quick thought, as user needs to inherit from LogCaptureMixin to enable this feature, but user might have his own class hierarchy while defining testsuites, is it possible to support syntax like:

@testsuite(mixins=[LogCaptureMixin, ...])
def AlphaSuite(BaseSuite):
      blah...

What do you think?

Again, this is a great addition to testplan, thank you for the contribution!

kn-ms:
> I think overall this looks really good! (Except for I confused myself with this pre_testcase thing for
> a while). I just have one quick thought, as user needs to inherit from LogCaptureMixin to enable this feature, but user might have his own class hierarchy while defining testsuites, is it possible to support syntax like:
> 
> @testsuite(mixins=[LogCaptureMixin, ...])
> def AlphaSuite(BaseSuite):
> blah...
> 
> How do you feel about this idea?
> 
> Again, this is a great addition to testplan, thank you for the contribution!

I do not think this will be a problem with their class hierarchy they just can use

```python
@testsuite
def AlphaSuite(BaseSuite, LogCaptureMixin):
    ...
```
Which is better from ide point of view, as it can pick up the function and logger member, and I found it much more clear to see what is happening.

kn-ms:
Yifan, can you merge this PR, I guess it needs the travis to rerun as github seems missed the status.

morsedav:
I'm not seeing any running builds in Travis so I'll force start one.

PR #413, opened 2020-04-01
slugify runpath

dcm:


testplan/common/entity/base.py
```diff
@@ -512,12 +513,17 @@ def generate_runpath(self):
         if self.parent and self.parent.runpath:
             return os.path.join(self.parent.runpath, self.uid())
```
Pyifan, line 514: I remember for some derived classes, the uid() could just be its name which is a free style string. Do you meant to apply slugify on it, because slugify is imported but not used.

kelliott55:
@dcm What are we doing with this PR?

PR #416, opened 2020-04-03
Add npm-force-resolutions.

yuxuan-ms:


morsedav:
This doesn't do anything at the moment, and the instructions say you need to add `"preinstall": "npx npm-force-resolutions"` to the `"scripts"` of `package.json`. However, when I do that, I keep getting this error:

```
[BABEL] C:\MyWorkspace\testplan\web_ui\testing\src\Report\BatchReport\components\__tests__\DocumentationButton.test.jsx: Could not find plugin "proposal-numeric-separator". Ensure there is an entry in ./available-plugins.js for it. (While processing: "C:\\MyWorkspace\\testplan\\web_ui\\testing\\node_modules\\babel-preset-react-app\\index.js$0")

      at getPlugin (node_modules/@babel/preset-env/lib/index.js:67:11)
      at Array.from.map.pluginName (node_modules/@babel/preset-env/lib/index.js:258:62)
          at Array.map (<anonymous>)
      at _default (node_modules/@babel/preset-env/lib/index.js:258:43)
      at node_modules/@babel/helper-plugin-utils/lib/index.js:19:12
      at node_modules/@babel/core/lib/config/full.js:199:14
      at Function.<anonymous> (node_modules/@babel/core/lib/gensync-utils/async.js:26:3)
      at evaluateSync (node_modules/gensync/index.js:244:28)

```

I haven't done mch digging into it at this point.

morsedav:
Actually this isn't too great.

So, if you're installing on a brand new workspace, you need to install without `"preinstall": "npm-force-resolutions"` first, then add this to your package.json, then install again, then remember to remove it again so it's not committed. See [here](https://github.com/rogeriochaves/npm-force-resolutions/issues/10#issuecomment-593875949).

I think if having `"resolutions"` is critical (which it may well be since we're not using the public NPM internall and thus are on a take-what-you-can-get wrt package versions) then we should just switch to yarn.

PR #418, opened 2020-04-16
Fix testplan package copy logic for internal test failure

Pyifan:


testplan/runners/pools/remote.py
```diff
@@ -222,9 +222,7 @@ def _copy_testplan_package(self):
             # add trailing / to _testplan_import_path.local
             # this will copy everything under import path to to testplan_lib
             self._transfer_data(
-                source=os.path.join(
-                    self._testplan_import_path.local, os.path.sep
-                ),
+                source=os.path.join(self._testplan_import_path.local, ""),
```
morsedav, line 225: I didn't know that 👍

PR #422, opened 2020-04-28
Add bytes regex support.

yuxuan-ms:


testplan/common/utils/match.py
```diff
@@ -26,9 +26,23 @@ def match_regexps_in_file(logpath, log_extracts, return_unmatched=False):
 
     extracts_status = [False for _ in log_extracts]
 
-    with open(logpath, "r") as log:
+    # If log_extracts contain bytes regex, will convert all log_extracts to
+    # bytes regex.
+    if not all([isinstance(x.pattern, str) for x in log_extracts]):
```
morsedav, line 31: Here too we could maybe save a few clock cycles and remove ambiguity:

```
if not six.PY2 and not all([isinstance(x.pattern, six.text_type) for x in log_extracts]):
    ...
```

testplan/testing/multitest/driver/base.py
```diff
@@ -217,7 +217,11 @@ def extract_values(self):
                 logpath=outfile, log_extracts=regexps, return_unmatched=True
             )
             unmatched.extend(file_unmatched)
-            self.extracts.update(file_extracts)
+            for k, v in file_extracts.items():
+                if isinstance(v, bytes):
+                    self.extracts[k] = v.decode()
```
morsedav, line 222: Similarly wrt my other comments maybe this is better:

```
if isinstance(v, six.binary_type):
    self.extracts[k] = v.decode("utf_8")
```

testplan/vendor/tempita/__init__.py
```diff
@@ -1172,6 +1172,7 @@ def get_token(pos=False):
             start_pos = end_pos = None
             parts = []
             while 1:
+                # pylint: disable=unbalanced-tuple-unpacking
                 tok_type, tok_string, s, e = get_token(True)
```
morsedav, line 1176: I know you didn't add this but we're confident that we'll never hit [this StopIteration](https://github.com/Morgan-Stanley/testplan/blob/668ca2d1d9ed25e9679f25952e55fce77b1a7399/testplan/vendor/tempita/__init__.py#L1134) exception? It seems like [this generator](https://github.com/Morgan-Stanley/testplan/blob/668ca2d1d9ed25e9679f25952e55fce77b1a7399/testplan/vendor/tempita/__init__.py#L1124) will run out eventually, thus throwing StopIteration, thus returning only two elements, thus throwing an exception here since we're expecting at least 4 elements.

PR #428, opened 2020-05-06
Make `get_default_exporters` a static method of runner class

raoyitao:
* the function `get_default_exporters` is only used by `TestRunner`
  and its sub classes. Make it a static method of runner class so
  the sub classes can overwrite it and accept new config options
  which is related to command line arguments.

### DIFF
--- a/testplan/runnable/base.py
+++ b/testplan/runnable/base.py
@@ -37,27 +37,6 @@ from testplan.testing import listing, filtering, ordering, tagging
 from testplan.testing.base import TestResult
 
 
-def get_default_exporters(config):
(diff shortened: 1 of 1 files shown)


testplan/runnable/base.py
```diff
@@ -331,6 +310,29 @@ def exporters(self):
                 exporter.parent = self
         return self._exporters
 
+    @staticmethod
+    def get_default_exporters(config):
```
Pyifan, line 314: This doesn't make a lot of sense to me: a static method but taking self.cfg as a parameter.

PR #429, opened 2020-05-06
Remove deprecated argument used in interactive mode

raoyitao:
* Arguments `interactive` and `port` should be totally removed from
  runner class because a single argument `interactive_port` is used.
* Correct document and examples.
* Fix typo.

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -634,10 +634,10 @@ class Runnable(Entity):
     :py:class:`~testplan.common.entity.base.Resource` objects
     that can be started/stopped and utilized by the steps defined.
 
-    :param interactive: Enable interactive execution mode.
-    :type interactive: ``bool``
-    :param interactive_no_block: Do not block on run() on interactive mode.
-    :type interactive_no_block: ``bool``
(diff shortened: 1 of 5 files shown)


testplan/common/entity/base.py
```diff
@@ -634,10 +634,10 @@ class Runnable(Entity):
...
     that can be started/stopped and utilized by the steps defined.
 
-    :param interactive: Enable interactive execution mode.
-    :type interactive: ``bool``
-    :param interactive_no_block: Do not block on run() on interactive mode.
-    :type interactive_no_block: ``bool``
+    :param interactive_port: Enable interactive execution mode on a port.
+    :type interactive_port: ``int`` or ``NoneType``
+    :param interactive_block: Block on run() on interactive mode.
+    :type interactive_block: ``bool``
```
Pyifan, line 640: Is it OK to just revert this parameter? Is it used anywhere?

raoyitao: Our testcase uses this argument, I'm not very clear with this argument, but seems it can have some effect, I think just leave it here now.

PR #430, opened 2020-05-10
add binary data validation & typechecks

dcm:


testplan/common/utils/testing.py
```diff
@@ -131,6 +140,23 @@ def suppress_warnings(func):
     return context_wrapper(warnings_suppressed)
 
 
+def _render_mismatch(orig_act, act, orig_exp, exp, full_path):
```
morsedav, line 143: FFR this was moved out to prevent creating a brand new function on each recursion. The optimization was too easy to make so I made it.

testplan/common/utils/testing.py
```diff
@@ -140,26 +166,6 @@ def check_iterable(
...
-        ).format(
-            full_path=full_path,
-            linesep=os.linesep,
-            expected=pprint.pformat(exp, indent=2),
-            actual=pprint.pformat(act, indent=2),
-            exp_data=pprint.pformat(_orig_exp, indent=2),
-            act_data=pprint.pformat(_orig_act, indent=2),
-        )
-
-    msg = render_mismatch(act=actual, exp=expected, full_path=curr_path)
```
morsedav, line 161: This doesn't need to be invoked on every single recursion. Another easy optimization to move this to where it _might_ be used.

testplan/report/testing/schemas.py
```diff
@@ -75,111 +77,60 @@ def _deserialize(self, value, attr, data):
 
 
 class EntriesField(fields.Field):
```
morsedav, line 79: At this point it's clear that this class needs to be made much simpler or much more complicated...

tests/unit/testplan/report/test_testing.py
```diff
@@ -342,14 +343,33 @@ def dummy_test_plan_report():
...
+    res.equal(
+        b"\x00\xb1\xC1",
+        b"\x00\xB2\xc2",
+        "mix-caps multi byte"
+    )
+    res.equal(
+        [b"\x00\xb1\xC1"],
+        {b"\x00\xB2\xc2"},
+        "mix-caps multi byte mutable containers"
+    )
```
morsedav, line 362: At present these containers are stringified before we can access the bytestrings in the serializer classes

PR #431, opened 2020-05-12
Add support to automatically re-run failed task

raoyitao:
* Implement task rerun feature, it has default implementation (check
  if test pass) but can also be decided by end users.
* Add a `rerun` argument to `Task`, when a task running in `Pool`
  fails, testplan will automatically re-assign it to a worker.
* Intermediate task report should be appended to the end.
* Pool has a member `_task_retries_limit`, it is used to restrict
  the total times a task is re-assigned due to some error (e.g. worker
  died), while property `rerun` of a task is used to restrict the total
  it can reassigned and run if `should_rerun` returns True.
* Some simple interfaces for derived Pool classes in the future.
* New testcase for rerun feature.

### DIFF
--- a/testplan/runners/pools/base.py
+++ b/testplan/runners/pools/base.py
@@ -1,13 +1,14 @@
 """Worker pool executor base classes."""
-import numbers
 import os
+import numbers
 import threading
 import time
 import datetime
+import uuid
 import pprint
 import traceback
 
-from schema import Or, And
+from schema import Or, And, Use
 
 from testplan.common.config import ConfigOption, validate_func
 from testplan.common import entity
@@ -15,6 +16,7 @@ from testplan.common.utils.thread import interruptible_join
 from testplan.common.utils.strings import Color
 from testplan.common.utils.timing import wait_until_predicate
 from testplan.runners.base import Executor, ExecutorConfig
+from testplan.report import ReportCategories
 
 from .communication import Message
 from .connection import QueueClient, QueueServer
@@ -193,11 +195,37 @@ class Worker(entity.Resource):
         return "{}[{}]".format(self.__class__.__name__, self.cfg.index)
 
 
-def default_check_reschedule(pool, task_result):
+def default_check_rerun(pool, task_result):
     """
-    Determines if a task should be rescheduled based on the task result info.
+    Determines if a task needs to be rerun based on the task result info.
+
+    :param task_result: Task execution result.
+    :type task_result:
+        :py:class:`~testplan.runners.pools.tasks.base.TaskResult`
+    :return: True if Task should be re-run else False.
+    :rtype: ``bool``
     """
+    if task_result.task and task_result.task.rerun > 0:
+        result = task_result.result
+        if not task_result.status or (
+            result and not (result.run and result.report.passed)
+        ):
+            return True
     return False
+
+
+def validate_custom_func(check_rerun):
+    """
+    Validate the callable which determines if a task should be rerun.
+    It must accept pool object (``pool``) and task result (``task_result``).
+
+    :param check_rerun: Custom callable for task rerun or reschedule.
+    :type check_rerun: ``callable``
+    """
+    if validate_func("pool", "task_result")(check_rerun):
+        return check_rerun
+    else:
+        raise ValueError("Not a valid function for checking rerun")
 
 
 class PoolConfig(ExecutorConfig):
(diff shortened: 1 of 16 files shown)


testplan/runners/pools/base.py
```diff
@@ -447,7 +473,9 @@ def _handle_taskresults(self, worker, request, response):
                 "De-assign {} from {}".format(task_result.task, worker)
             )
 
-            if self.should_reschedule(self, task_result):
+            if self.should_reschedule(self, task_result) or self.should_rerun(
+                task_result
+            ):
                 if self.task_assign_cnt[uid] >= self.cfg.task_retries_limit:
```
Pyifan, line 479: retry: if worker dies (no result returned, detected by monitor), reschedule the task to pool - subject to pool.task_retries_limit. retry is currently enabled by default.

rerun: if task result is fail and user specifies so, reschedule the task to pool - subject to task.max_rerun. rerun can be controlled by user.

I think we shall keep these 2 and their limits separate

testplan/runners/pools/base.py
```diff
@@ -447,7 +473,9 @@ def _handle_taskresults(self, worker, request, response):
                 "De-assign {} from {}".format(task_result.task, worker)
             )
 
-            if self.should_reschedule(self, task_result):
+            if self.should_reschedule(self, task_result) or self.should_rerun(
```
Pyifan, line 476: I think the rerun logic could be implemented as an default implementation of should_reschedule. User can override if desired.

I also suggest that we use term 'rerun' in here, we can save 'reschedule' for assigning task from one pool to another.

testplan/runners/pools/base.py
```diff
@@ -259,25 +310,29 @@ def __init__(
...
         restart_count=3,
+        max_active_loop_sleep=5,
+        should_rerun=default_check_rerun,
         **options
     ):
         options.update(self.filter_locals(locals()))
         super(Pool, self).__init__(**options)
         self.unassigned = []  # unassigned tasks
         self.task_assign_cnt = {}  # uid: times_assigned
+        # Method `should_reschedule` should be implemented internally while
```
Pyifan, line 323: Let's remove the should_reschedule code path for now, it doesn't add any value but could be quite confusing for ppl who is looking at this code. We can easily add it back if we need such in future.

testplan/runners/pools/base.py
```diff
@@ -441,16 +535,25 @@ def _handle_taskresults(self, worker, request, response):
...
             uid = task_result.task.uid()
             worker.assigned.remove(uid)
-            if worker not in self._workers_last_result:
-                self._workers_last_result[worker] = time.time()
+            self._workers_last_result.setdefault(worker, time.time())
             self.logger.test_info(
                 "De-assign {} from {}".format(task_result.task, worker)
             )
 
             if self.should_reschedule(self, task_result):
```
Pyifan, line 543: Same as above, I think this code branch can be removed for now.

testplan/runners/pools/base.py
```diff
@@ -441,16 +535,25 @@ def _handle_taskresults(self, worker, request, response):
...
+                self.logger.test_info(
+                    "Will reschedule %(task)s due to `should_reschedule` "
+                    "cfg option of %(pool)s",
+                    {"task": task_result.task, "pool": self,},
+                )
+                self.unassigned.append(uid)
+                continue
+
+            if self.should_rerun(self, task_result):
+                if self.task_assign_cnt[uid] > self.cfg.task_retries_limit:
```
Pyifan, line 553: should_rerun already checked against task's rerun limit, the task_retries_limited is checked in taskpull handler.

In a treadmill pool, if a task is scheduled out to a worker, and in the worker's process pool, it has already retried 3 time (but worker died every time with no result returned), and finally the task is returned back to treadmill, treadmill need to reset its retry number and rerun this task again. At least this is how its handled in ets/python.

testplan/runners/pools/base.py
```diff
@@ -671,6 +771,23 @@ def _discard_pending_tasks(self):
...
+        test_report = task_result.result.report
+        uid = task_result.task.uid()
+        if uid not in self.task_assign_cnt:
+            return
+
+        iteration = self.task_assign_cnt[uid] - 1
+        postfix = " => Run {}".format(iteration)
+        test_report.name = "{}{}".format(test_report.name, postfix)
+        test_report.uid = "{}{}".format(test_report.uid, postfix)
+        test_report.category = ReportCategories.MULTITEST_RERUN
```
Pyifan, line 785: Task is not necessarily a Multitest.

testplan/runners/pools/base.py
```diff
@@ -259,25 +292,27 @@ def __init__(
...
         **options
     ):
         options.update(self.filter_locals(locals()))
         super(Pool, self).__init__(**options)
         self.unassigned = []  # unassigned tasks
-        self.task_assign_cnt = {}  # uid: times_assigned
-        self.should_reschedule = default_check_reschedule
+        self._task_retries_cnt = {}  # uid: times_reassigned_without_result
+        self._task_retries_limit = 2
+        self._task_rerun_limit = 3
```
Pyifan, line 305: Let's control rerun limit only at Task. I prefer not having yet another limit in Pool to control the same behavior

tests/functional/testplan/runners/pools/test_task_rerun.py
```diff
@@ -0,0 +1,298 @@
+"""TODO."""
```
Pyifan, line 1: Remaining TODO.

PR #439, opened 2020-05-25
Re-organize document structure of OUTPUT chapter

raoyitao:
* Re-organize the document structure of OUTPUT chapter, for the ease of
  extending the document with appended content.
* Fix some small issues in document.


doc/en/exporters.rst
```diff
@@ -0,0 +1,308 @@
+Exporters
+=========
+
+At the end of a test run Testplan creates a
+:py:class:`report <testplan.report.testing.base.TestReport>` object, which is
```
Pyifan, line 5: I suggest use :py:class:`<~testplan.report.testing.base.TestReport>` instead?

PR #440, opened 2020-05-25
enum34 is not compatible with newer python

cpages:
If enum34 is installed with newer python versions (>3.3) pip then fails on successive runs with:
`  AttributeError: module 'enum' has no attribute 'IntFlag'`


kelliott55:
Thank you for the PR!
Please could you review the [CONTRIBUTING](https://github.com/Morgan-Stanley/testplan/blob/master/CONTRIBUTING.rst) file - we'll need a DCO completing before we can merge.

PR #450, opened 2020-06-23
Cppunit test integreated

raoyitao:
* Integrate Cppunit test into Testplan.
* Replace argument `driver` with `binary` in `ProcessRunnerTest` and
  the derived classes from it.
* Add examples and testcases (executable built from CPP source).


testplan/testing/cpp/cppunit.py
```diff
@@ -0,0 +1,363 @@
...
+            cmd.extend([self.cfg.file_output_flag, self.report_path])
+        return cmd
+
+    def list_command(self):
+        if self.cfg.listing_flag:
+            return [self.cfg.driver, self.cfg.listing_flag]
+        else:
+            return super(Cppunit, self).list_command()
+
+    def read_test_data(self, stdout=None, stderr=None):
```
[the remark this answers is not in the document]
raoyitao: the only reason to pass this handler is that int the outer function of read_test_data, the stdout and stderr have been open as handlers. so I don't want to open them twice before they are closed.

testplan/testing/base.py
```diff
@@ -400,8 +400,8 @@ class ProcessRunnerTest(Test):
     Test report will be populated by parsing the generated report output file
     (report.xml file by default.)
 
-    :param driver: Path the to application binary.
-    :type driver: ``str``
+    :param executable: Path the to application binary or script.
```
Pyifan, line 403: typo

doc/en/drivers_basic.rst
```diff
@@ -122,7 +122,7 @@ started test process. Have a look at the following code:
 
     plan.add(GTest(
         name='My GTest',
-        driver=BINARY_PATH,
+        executable=BINARY_PATH,
```
Pyifan, line 125: Calling the parameter 'executable', but the meta var and documentation uses 'binary' most often - this is a bit inconsistent. I would suggest that we call this parameter 'binary' for simplicity. We use 'binary' most often when we need to specify an executable for drivers in testplan2. Consistency makes easy for user.

testplan/testing/cpp/gtest.py
```diff
@@ -63,8 +62,8 @@ class GTest(ProcessRunnerTest):
 
     :param name: Test instance name. Also used as uid.
     :type name: ``str``
-    :param driver: Path the to application binary.
-    :type driver: ``str``
+    :param executable: Path the to application binary or script.
```
Pyifan, line 65: typo

testplan/testing/cpp/cppunit.py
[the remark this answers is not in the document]
Pyifan: In that case, maybe we can refactor the caller code.

doc/en/drivers_basic.rst
raoyitao: use `binary` instead

PR #471, opened 2020-07-31
Some fixes/changes required to run all pytest with Testplan ProcessPool.

Pyifan:
This PR has a few changes:
1) a few bug fixes for pytests to be run in process pool
2) introduction of runpath & mockplan fixture to pytest root, not that they are function scoped fixtures
3) introduction of TestplanMock class for testing, where mockplan fixture is insufficient
4) interactive mode tests are disabled until we have time working on that area again


testplan/base.py
```diff
@@ -430,3 +431,12 @@ def test_plan_inner_inner():
...
 
 test_plan = Testplan.main_wrapper
+
+
+class TestplanMock(Testplan):
+    """A mock Testplan class for testing purpose"""
+
+    def __init__(self, *args, **kwargs):
+        kwargs["abort_signals"] = []  # mock testplan could run in threads
+        kwargs.setdefault("runpath", tempfile.mkdtemp())
```
raoyitao, line 441: seems this temp dir will not automatically removed, if the test fails and we run it again and again, them more and more temp directories will be made and a lot of disk space might be occupied. Is it better get a fixed temp directory thus the test will use the same space again. or remove the temp dir in some post steps (e.g. TestplanMock can accept an augument to decide if do clear job after test, by default it is `True`) ?

testplan/runners/pools/child.py
```diff
@@ -63,26 +63,11 @@ def metadata(self):
...
-        """Start the heartbeat manager thread."""
-        heartbeat = threading.Thread(target=self.heartbeat_thread)
-        heartbeat.daemon = True
-        heartbeat.start()
-
     def _child_pool(self):
         # Local thread pool will not cleanup the previous layer runpath.
         self._pool = self._pool_type(
             name="Pool_{}".format(self._metadata["pid"]),
+            # restart_count=0,
```
raoyitao, line 70: what's the aim of this annotated line ?

testplan/runners/pools/child.py
```diff
@@ -194,13 +179,14 @@ def worker_loop(self):
             return
 
         with self._child_pool():
-            if self._pool_cfg.worker_heartbeat:
-                self.heartbeat_setup()
             message = Message(**self.metadata)
             next_possible_request = time.time()
+            next_heartbeat = time.time()
```
raoyitao, line 184: here, set next_heartbeat = 0 to avoid unnecessary computation.

testplan/runners/pools/child.py
```diff
@@ -217,7 +203,7 @@ def worker_loop(self):
                                 time.time() - hb_resp.data,
                             )
                         )
-                    self._to_heartbeat = self._pool_cfg.worker_heartbeat
+                    next_heartbeat = now + self._pool_cfg.worker_heartbeat
```
raoyitao, line 206: now = time.time() is gotten at the first line in while loop. if it takes some time to execute line 190-205, e.g. worker respond slowly due to high CPU consuming, then `now + self._pool_cfg.worker_heartbeat` could  be smaller than the current time.time(). After getting the result of heartbeat check, no need to check it again. So I think it could be better written:

while True:
    if if self._pool_cfg.worker_heartbeat and time.time() > next_heartbeat:
        ....
        next_heartbeat = time.time() + self._pool_cfg.worker_heartbeat

testplan/base.py
Pyifan: It seems we cannot have it all. The most important thing is to make sure runpath doesn't collide to ensure test stability. And I prefer that we keep the test output for easy investigation. So the best compromise is to write to /tmp and let system clean it up. I'm wrapping pytest fixture tmp_path that creates a very nice tmp dir with testcase information.

testplan/runners/pools/child.py
Pyifan: Just for debugging, I can remove it.

testplan/runners/pools/child.py
Pyifan: It does trigger heartbeat in the first iteration. I'm not a great fan of initializing next_heartbeat to 0 though.

PR #472, opened 2020-08-03
allow non-string headers in Tablelog

dcm:


testplan/common/utils/table.py
```diff
@@ -30,14 +33,40 @@ def _check_table(self, table):
...
+                if isinstance(cel, six.string_types):
+                    header_row_mut_normd.append(cel)
+                elif isinstance(cel, six.integer_types):
+                    header_row_mut_normd.append("{}".format(cel))
+                elif isinstance(cel, float):
+                    header_row_mut_normd.append("{}".format(cel))
+                else:
+                    raise TypeError(
+                        (
+                            "Table headers must all be strings - "
```
kn-ms, line 61: The error is not in sync with behavior, now int and float is ok as well

kn-ms (dismissed):

kn-ms:
Why limiting headers to be str, int, float, can it be just like anything having __str__ ?

Pyifan:
Do we really need to limit the meta data row to string only? If not, we can just save the data as is without converting to str (and introducing copy and lose of reference etc).

Pyifan (approved):

Pyifan (approved):

PR #485, opened 2020-08-30
Support custom testcase/testsuite name

raoyitao:
* User can define the names of testcase/testsuite in @testcase and
  @testsuite decorator which are human readable. Instead of class
  name or function name, these strings will be displayed on UI or
  exported PDF. Also they are used for pattern matching while
  filtering testcases, and set as parent uids for a report entity.
* No testcases having the same function name can be defined in one
  test suite, Testplan will check it and raise. User defined names
  should also be unique. If the function name of generated testcase
  conflict with others, Testplan could altomatically solve it.
* No duplicate test suite name allowed, otherwise at the step of
  "generating report" exception will be raised.
* User can define a naming function for parametrized testcases, or
  the default one is used, generating a function name like this:
  "<original_name> {arg_name:value, ...}". Raise if testcase names
  are not valid (e.g. name too long), then users have to pass their
  own `name_func`.
* "::" is the delimiter for listed cases to conform to a more generic
  format, e.g. "Multitest::Suite::TestCase". To list test with
  "--tests" argument, both ":" and "::" can be used as delimiter.
* CFG option `reset_report_uid` is added to `TestRunner` so that
  it is not necessary to make UUID as unique identifier for test
  report, without UUID set, the original class name or function name
  can be used as unique id for test report.
* Aoid creating test report twice while executing Multitest instance.
* All false value including "None", empty string, number 0 can not be
  set as uid of test report, otherwise will be replaced by uuid4(),
  change the testcase where number 0 is passed to report constructor
  as uid.
* Update related testcases.

### DIFF
--- a/testplan/testing/multitest/parametrization.py
+++ b/testplan/testing/multitest/parametrization.py
@@ -6,8 +6,6 @@ import itertools
 import re
 import warnings
 
-import six
-
 from testplan.common.utils.convert import make_tuple
 from testplan.common.utils import callable as callable_utils
 from testplan.testing import tagging
@@ -20,9 +18,12 @@ from testplan.testing import tagging
 PYTHON_VARIABLE_PATTERN = r"^(?![\d])\w+$"
 PYTHON_VARIABLE_REGEX = re.compile(PYTHON_VARIABLE_PATTERN)
 
-# Python attribute names can be of unlimited length.
-# But the UI will not be pleasant when we end up with really long names.
+# Python attribute names can be of unlimited length but we set a limit here
 MAX_METHOD_NAME_LENGTH = 255
+# Name of testcase (usually used for display) cannot be too long,
+# Because the UI will not be pleasant when we end up with really long names
+MAX_TESTSUITE_NAME_LENGTH = 80
+MAX_TESTCASE_NAME_LENGTH = 80
 
 
 class ParametrizationError(ValueError):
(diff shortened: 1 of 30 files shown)


testplan/testing/multitest/parametrization.py
```diff
@@ -372,9 +321,54 @@ def default_name_func(func_name, kwargs):
...
 
 
+def default_name_func(func_name, kwargs):
+    """
+    Readable testcase name generator for parametrized testcases.
+
+    >>> import collections
+    >>> default_name_func('Test Method',
+                          collections.OrderedDict(('foo', 5), ('bar', 10)))
+    'Test Method {foo:5, bar:10}'
```
Pyifan, line 331: I suggest using = instead of :

testplan/testing/multitest/parametrization.py
```diff
@@ -372,9 +321,54 @@ def default_name_func(func_name, kwargs):
...
+            )
+        else:
+            return str(s)
+
+    arg_strings = ["{arg}:{{{arg}}}".format(arg=arg) for arg in kwargs]
+    template = "{func_name} {{" + ", ".join(arg_strings) + "}}"
+    return template.format(
+        func_name=func_name, **{k: escape_char(v) for k, v in kwargs.items()}
+    )
+
```
Pyifan, line 366: func name<a='abc', b=(1, 2, 3)>

tests/unit/testplan/testing/test_filtering.py
```diff
@@ -275,17 +275,22 @@ def test_filter_parametrized_cases(self):
...
-    def test_filter_initialization_error(self):
+    def test_filter_initialization(self):
         """
-            Pattern filter should raise error if
-            pattern depth exceeds MAX_LEVEL
+        Symbol ":" or "::" can be used as delimiter and MAX_LEVEL parts
+        can be generated for matching.
         """
-        with pytest.raises(ValueError):
-            filtering.Pattern("foo:bar:baz:bat")
```
Pyifan, line 288: I would suggest that we restore this pattern length check.

Pyifan (approved):

PR #488, opened 2020-09-01
Names of split json report are recorded directly in main report

raoyitao:
* In main report 2 fields `structure_file` and `assertions_file` are
  added so there's no need to search them in `attachments`.
* version info added.

### DIFF
--- a/examples/Test Output/Exporters/JSON/json_to_pdf.py
+++ b/examples/Test Output/Exporters/JSON/json_to_pdf.py
@@ -15,8 +15,17 @@ def main(source, target):
 
     with open(source) as source_file:
         data = json.loads(source_file.read())
-        if data.get("split"):
-            raise RuntimeError("Cannot process JSON report that was split")
+        if (
+            data.get("version", 1) >= 2
+            or data.get("structure_file")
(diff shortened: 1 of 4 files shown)


yuxuan-ms (dismissed):

examples/Test Output/Exporters/JSON/json_to_pdf.py
```diff
@@ -15,7 +15,7 @@ def main(source, target):
 
     with open(source) as source_file:
         data = json.loads(source_file.read())
-        if data.get("split"):
+        if data.get("structure_file") or data.get("assertions_file"):
```
yuxuan-ms, line 18: if data.get("version", 1) < 2:

examples/Test Output/Exporters/JSON/json_to_pdf.py
```diff
@@ -15,7 +15,7 @@ def main(source, target):
...
             raise RuntimeError("Cannot process JSON report that was split")
```
Pyifan, line 19: I think we can point user to your utility script that merges split json files

Pyifan (approved):

PR #500, opened 2020-09-15
Add result.markdown

yuxuan-ms:


testplan/web_ui/testing/src/AssertionPane/AssertionTypes/MarkdownAssertion.js
```diff
@@ -0,0 +1,33 @@
...
+  return (
+    <SyntaxHighlighter language={props.language} style={coy}>
+      {props.value}
+    </SyntaxHighlighter>
+  );
+}
+
+export default function MarkdownAssertion(props) {
+  return (
+    <>
```
kn-ms, line 17: As there is only a markdown node here, is this empty node needed?

Pyifan (changes requested):
As a new assertion, let's add it to documentation and assertion example.

Pyifan (approved):

PR #501, opened 2020-09-18
UI does not display top-level errmsg correctly

raoyitao:
* If the top most testplan has errors logs (for example: timeout), as
  clicking the lable of testplan in nav bar, the error message can be
  displayed, however, as the report is loaded the error message is not
  automatically displayed, that's not a correct behavior.

    -- Should fix the incorrect behavior
    -- No "description" field is kept for top level report, it is not
       passed as an argument while initializing report, should fix it.
    -- Update JS testcase.

### DIFF
--- a/testplan/report/testing/schemas.py
+++ b/testplan/report/testing/schemas.py
@@ -260,6 +260,7 @@ class TestReportSchema(Schema):
 
     timer = TimerField(required=True)
     name = fields.String(required=True)
+    description = fields.String(allow_none=True)
     uid = fields.String(required=True)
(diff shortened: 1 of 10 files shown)


Pyifan (changes requested):
Hi Yitao, nice change! Could you add a testcase to cover the top level error msg scenario please?

raoyitao:
> Hi Yitao, nice change! Could you add a testcase to cover the top level error msg scenario please?

Test case added

Pyifan (approved):

PR #504, opened 2020-09-23
Change pre_start order in driver start-up logic

raoyitao:
* User defined `pre_start` should be called by driver's `pre_start`
  after making runpath.
* Similar changes applied to `post_start`, `pre_stop` and `post_stop`.
* update testcase.

### DIFF
--- a/testplan/testing/multitest/driver/base.py
+++ b/testplan/testing/multitest/driver/base.py
@@ -142,18 +142,18 @@ class Driver(Resource):
     def start(self):
         """Start the driver."""
         self.status.change(self.STATUS.STARTING)
+        self.pre_start()
         if self.cfg.pre_start:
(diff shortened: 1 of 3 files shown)


Pyifan (changes requested):
As we discussed, I think it would be easier for maintenance if we call pre_start() then call cfg.pre_start in Driver's start() method. The current implementation put a string on driver writer to either call super's pre_start or handle cfg.pre_start himself.

raoyitao:
> As we discussed, I think it would be easier for maintenance if we call pre_start() then call cfg.pre_start in Driver's start() method. The current implementation put a string on driver writer to either call super's pre_start or handle cfg.pre_start himself.

Changed

Pyifan (approved):

PR #512, opened 2020-10-06
Standardize internal table representation

dcm:
testplan.common.utils.table.TableEntry is maintained internally as a list of lists


testplan/common/utils/table.py
```diff
@@ -32,58 +32,21 @@ def _std_table_from_list_of_lists(tbl):
...
-                    curr_row[tgt_col_idx] = cell_val
-                    checking_row_idx = -1
-                else:
-                    checking_row_idx += 1
-    return [tbl_headers] + tbl_data
+    dict_of_lists = {}
+    for dict_elem in tbl:
+        for hdr, val in six.iteritems(dict_elem):
+            dict_of_lists.setdefault(hdr, [])
+            dict_of_lists[hdr].append(val)
```
kn-ms, line 48: you do not need this extra lookup, the return value of setdefault is the one you want to append to.

testplan/common/utils/table.py
```diff
@@ -32,58 +32,21 @@ def _std_table_from_list_of_lists(tbl):
...
-                # else we'll need to check the same on the next row
-                if curr_row[tgt_col_idx] == _TP_BLANK_CELL:
-                    curr_row[tgt_col_idx] = cell_val
-                    checking_row_idx = -1
-                else:
-                    checking_row_idx += 1
-    return [tbl_headers] + tbl_data
+    dict_of_lists = {}
+    for dict_elem in tbl:
+        for hdr, val in six.iteritems(dict_elem):
```
kn-ms, line 46: This does not sounds to be right when dictionaries in the list are non uniform (having different key sets).

Suggestion:
- compute the union of key sets from the dictionaries
- do whatever you do currently to add to the lists
- for the difference of the key set and the union, add None's or something that represent missing value

Or alternatively just reject those datasets to keep things simple, and get the user to provide nice data.

testplan/web_ui/testing/src/Common/utils.js
```diff
@@ -148,8 +148,8 @@ export const flattened = (strArr, ...tagsArr) => {
 export const toPlainObjectIn = (obj, enumerableOnly = false) => {
   const objectify = o => {
     return Object.fromEntries(Object.entries(o)
-      .filter(([_, v]) => _.isPlainObject(v.value))
-      .filter(([_, v]) => !enumerableOnly || v.enumerable)
+      .filter(([__, v]) => _.isPlainObject(v.value))
```
dcm, line 151: `_` is a global variable for `lodash`. Using it as a placeholder for an unused parameter made the placeholder take precedence over the global variable. Hence, this change.

testplan/common/utils/table.py
```diff
@@ -1,40 +1,114 @@
...
-                return table
-        raise TypeError(
-            (
-                "`table` must be a list of"
-                " lists or list of dicts, got:\n{}".format(table)
-            )
-        )
+        self._tbl_list_of_dict = None
+        self._tbl_list_of_list = None
+        if isinstance(table, Container):
```
Pyifan, line 86: I think this complicates things - Container can be a lot of things. The table we accept shall be a list / tuple at top level, period.

testplan/common/utils/table.py
```diff
@@ -1,40 +1,114 @@
...
-    Represents a table. Internally represented either
-    as a ``list`` of ``list`` or a ``list`` of ``dict``.
+    Represents a table. Internally represented as ``List[List[Any]]``.
     """
 
     def __init__(self, table):
-        self.table = self._validate_table(table or [])
-
-    @staticmethod
-    def _validate_table(table):
```
Pyifan, line 21: Actually this original implementation that validates user input looks just fine to me. Why are we changing it?

With that said, one thing I may want to add is - if user passes a list of list, we may want all sub-list to be of uniform length.

testplan/common/utils/table.py
```diff
@@ -1,40 +1,114 @@
...
+
+def _is_nonstr_sequence(obj):
+    return isinstance(obj, Sequence) and not isinstance(obj, six.string_types)
+
+
+def _table_from_list_of_lists(list_of_lists):
+    # type: (Sequence[Sequence[Any]]) -> List[List[Any]]
+    std_tbl = []
+    # all that needs to happen here is that each nested list must be padded to
+    # the length of the longest list
```
Pyifan, line 24: I would suggest we be more strict on user input. If user pass in a list of list, all sub-lists shall be of uniform length. It's a guess to pad at the end anyway - so better not do that at all.

testplan/common/utils/table.py
Pyifan: The consolidate_columns method of TableEntry does the union of keys. Also it defines placeholder as "ABSENT" for empty cells. We only need one placeholder either "ABSENT" or "_TP_BLANK_CELL", need to consolidate.

Pyifan (changes requested):
I didn't finish reviewing all files. The original user request for this change is they would like to be able to display a header only table if they only provided the meta data in table.log. For example:

result.table.log([["A", "B", "C"]].

It does not seem to me we are addressing this. TableLog assertion -> get_table -> TableEntry.as_list_of_dict(). So we are still saving table as list of dict in report and cannot properly present the scenario above.

Pyifan:
Close the PR.

PR #513, opened 2020-10-12
Add result.log_html and result.log_code

yuxuan-ms:


testplan/testing/multitest/result.py
```diff
@@ -1442,6 +1442,42 @@ def markdown(self, message, description=None, escape=True):
...
+        :param code: HTML code string. Tag <script> will not be executed.
+        :type code: ``str``
+        :param description: Text description for the assertion.
+        :type description: ``str``
+        :return: ``True``
+        :rtype: ``bool``
+        """
+        return self.markdown(code, description=description, escape=False)
+
+    def codelog(self, code, language="python", description="Code log"):
```
kn-ms, line 1459: This can be better if would be implemented just on top of SyntaxHighlighter instead of markdown as an intermediary. For this level of code log, I would go for a tip/suggestion/example next to the markdown in the documentation. 

With a full blown implementation on top of SyntaxHighlighter, many fine tune possibility can be put into the hand of the user.

testplan/testing/multitest/result.py
```diff
@@ -1442,6 +1442,42 @@ def markdown(self, message, description=None, escape=True):
...
+        """
+        Create a markdown message entry which only contains code snippet, can
+        be used for providing additional context related to test steps.
+
+        :param code: The source code string.
+        :type code: ``str``
+        :param language: The language of source code. e.g. json, xml, python,
+            java. Defaults to python.
```
Pyifan, line 1467: Can we provide more examples:
c++/cpp
js/javascript
shell?

Pyifan (approved):

Pyifan (approved):

PR #515, opened 2020-10-14
Improve the feature of testcase/suite name customization

raoyitao:
* Allow user to give testsuite & testcase a custom name
  - @testcase(name=<str>, name_func=<callable>)
  - @testsuite(name=<str or callable>)
* Consisten testcase/testsuite name func signature:
  - def case_name_fn(func_name, kwargs)
  - def suite_name_fn(cls_name, suite)
* If user explicitly makes "name_func=None" instead of the default one,
  parameterized testcases will get renamed with integer suffix appended
  at the end (no argument showed).
* If name of test suite or testcase is too long (e.g >255 char),
  give a soft warning.
* No duplicate test suite names are allowed for now.
* Fix documentation errors.

### DIFF
--- a/examples/Multitest/Parametrization/test_plan.py
+++ b/examples/Multitest/Parametrization/test_plan.py
@@ -77,10 +77,9 @@ class SimpleTest(object):
 # This way we can come up with more readable testcase
 # method names on the test reports.
 
-# If we didn't use a custom name function, we'd end up with method names
-# like `func_raises_error__0`, `func_raises_error__1`
-# But instead the custom function will give
-# us names like `func_raises_error__ValueError`
+# If we didn't use a custom name function, we'd end up with method name
+# like `func_raises_error <func=.., error=...>`, but instead, the custom
+# function will give us names like `func_raises_error__ValueError`.
 
 
 def custom_error_name_func(func_name, kwargs):
--- a/testplan/report/testing/base.py
+++ b/testplan/report/testing/base.py
@@ -207,10 +207,10 @@ class BaseReportGroup(ReportGroup):
 
     exception_logger = ExceptionLogger
 
-    def __init__(self, *args, **kwargs):
+    def __init__(self, name, **kwargs):
         self.meta = kwargs.pop("meta", {})
         self.status_reason = kwargs.pop("status_reason", None)
-        super(BaseReportGroup, self).__init__(*args, **kwargs)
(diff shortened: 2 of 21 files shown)


examples/Multitest/Parametrization/test_plan.py
```diff
@@ -78,9 +78,8 @@ def is_even(self, env, result, value):
...
 
 # If we didn't use a custom name function, we'd end up with method names
-# like `func_raises_error__0`, `func_raises_error__1`
-# But instead the custom function will give
-# us names like `func_raises_error__ValueError`
+# like `func_raises_error <func=.., error=...>`, but instead the custom
+# function will give us names like `func_raises_error__ValueError`.
 
 
 def custom_error_name_func(func_name, kwargs):
```
Pyifan, line 85: disregard `func` argument? but I don't see a func argument

testplan/report/testing/base.py
```diff
@@ -590,9 +591,10 @@ def __init__(
         extra_attributes=None,
         fix_spec_path=None,
         env_status=None,
+        *args,
```
Pyifan, line 594: I suggest that we remove *args for this class family

testplan/testing/base.py
```diff
@@ -129,9 +128,11 @@ class Test(Runnable):
     filter_levels = [filtering.FilterLevel.TEST]
 
     def __init__(self, **options):
-        validate_display_name(
-            options.get("name"), MAX_MULTITEST_NAME_LENGTH, "MultiTest name"
-        )
+        test_name = options.get("name", "")
+        if isinstance(test_name, str) and ":" in test_name:
```
Pyifan, line 132: why are we checking test_name is str here? We have already validated that in cfg schema?

testplan/testing/multitest/parametrization.py
```diff
@@ -230,24 +226,23 @@ def _generated(self, env, result):
 
 def _check_name_func(name_func):
     """
-    Make sure ``name_func`` is a callable that takes
-    ``func_name``, ``kwargs`` arguments.
+    Make sure ``name_func`` is `Nonr` or a callable that
```
Pyifan, line 229: None

Pyifan (changes requested):

Pyifan (approved):

PR #517, opened 2020-10-16
Add save report button in the interactive mode

yuxuan-ms:


testplan/runnable/interactive/http.py
```diff
@@ -429,6 +444,45 @@ def put(self, test_uid, suite_uid, testcase_uid, param_uid):
...
+    class ExportReport(flask_restplus.Resource):
+        """
+        Interactive export endpoint. There is an API for exporting root
+        report object.
+        """
+
+        def get(self):
+            available_exporters = []
+            with ihandler.report_mutex:
+                for exporter in ihandler.parent.exporters:
```
kn-ms, line 457: I would push this logic to the interactive handler. The http handler should not know about that handler's internal, like ti has a parent, not talking about assumptions of aprent internal. 

The same apply to the history, and the logic calling the exporters itself.

testplan/web_ui/testing/src/Toolbar/InteractiveButtons.js
```diff
@@ -46,4 +65,204 @@ const ResetButton = (props) => {
   }
 };
 
-export {ResetButton};
+
+const showHistory = (historyExporters) => {
+  if (Array.isArray(historyExporters)) {
```
kn-ms, line 70: This can be written as:

if (Array.isArray(historyExporters) && historyExporters.length > 0)

to spare an indent

testplan/web_ui/testing/src/Toolbar/InteractiveButtons.js
```diff
@@ -46,4 +65,204 @@ const ResetButton = (props) => {
   }
 };
 
-export {ResetButton};
+
+const showHistory = (historyExporters) => {
```
kn-ms, line 69: more of a getHistoryTable than showHistory as that sounds like you return a flag if show or not show

testplan/exporters/testing/pdf/__init__.py
```diff
@@ -144,6 +145,11 @@ def create_pdf(source, config):
             reportlab_data.extend(row_data.content)
             reportlab_styles.extend(row_data.style)
 
+    pdf_path = config.pdf_path
+    if os.path.exists(pdf_path):
```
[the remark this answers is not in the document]
yuxuan-ms: How to determinate the current runnable in the exporter class?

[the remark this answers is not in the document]
yuxuan-ms: > This might be valid in the case when you save from interactive, as you get a link/button for that. When one just run it from commandline I prefer the override behaviour, than generationg something I need to search for.

How to determinate the current runnable in the exporter class?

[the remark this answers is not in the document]
kn-ms: not necessary in the exporter class, might be in the export function which could take a parameter describing the nature of export.

testplan/web_ui/testing/src/Toolbar/InteractiveButtons.js
```diff
@@ -46,4 +65,228 @@ const ResetButton = (props) => {
   }
 };
 
-export {ResetButton};
+
+const getHistoryTable = (historyExporters) => {
+  if (Array.isArray(historyExporters) && historyExporters.length > 0) {
+    if (historyExporters.length > 0) {
```
kn-ms, line 71: This if is not needed, as you put this check in the above line

testplan/exporters/testing/pdf/__init__.py
```diff
@@ -104,7 +105,7 @@ def add_count_suffix(directory, path, count=0):
     return add_count_suffix(config.report_dir, path)
 
 
-def create_pdf(source, config):
+def create_pdf(source, config, override=True):
```
kn-ms, line 108: would not be better to call it overwrite instead of override?

testplan/common/exporters/__init__.py
```diff
@@ -43,6 +43,8 @@ class BaseExporter(Configurable):
 
     def __init__(self, **options):
         self._cfg = self.CONFIG(**options)
+        self.export_path = None
+        self.name = self.__class__.__name__
```
Pyifan, line 47: Would it be more tp2 style if we define a "name" config option in ExporterConfig?

testplan/exporters/testing/pdf/__init__.py
```diff
@@ -213,18 +220,32 @@ class PDFExporter(Exporter):
 
     CONFIG = PDFExporterConfig
 
-    def export(self, source):
-
-        pdf_path = self.cfg.pdf_path
+    def __init__(self, **options):
+        super(PDFExporter, self).__init__(**options)
+        self.name = "PDF exporter"
```
Pyifan, line 225: I think we could expose name config option to user, and give it a default if user didn't specify.

testplan/exporters/testing/pdf/__init__.py
```diff
@@ -213,18 +220,32 @@ class PDFExporter(Exporter):
...
-                pathname2url(os.path.abspath(pdf_path))
-            )
+            try:
+                if self.cfg.interactive_port is None:
+                    override = True
+                else:
+                    override = False
+            except AttributeError:
+                override = True
+            pdf_path = create_pdf(source, self.cfg, override=override)
```
Pyifan, line 236: I suggest that we move the check of cfg.interactive_port to create_pdf(), so that we can save the override parameter and there we are checking if the file already exists anyway.

testplan/runnable/interactive/http.py
```diff
@@ -429,6 +445,57 @@ def put(self, test_uid, suite_uid, testcase_uid, param_uid):
...
+                    if exporter.name in save_exports.get("exporters", []):
+                        export_result = {
+                            "time": time.time(),
+                            "name": exporter.name,
+                            "uid": str(uuid.uuid4()),
+                        }
+                        try:
+                            exporter.export(ihandler.report)
+                            export_result["success"] = True
+                            export_result["message"] = exporter.export_path
```
Pyifan, line 479: I think the intro of exporter.export_path and its usage in code is quite unnatural - can we change export() to return the exported report path? I think it will make the code look neater

Pyifan (changes requested):

Pyifan (approved):

PR #518, opened 2020-10-19
Nav list Filtering

kn-ms:
- Nav filter  first cut
- fix tags badge in named tag
- honor last user selection if available after filter
-remove generated parser from git
-free text search
-small ui changes
-rework grammar and the filters as well
-get build working
-use pegjs@dev that has es6 importable output
-parser compile to prebuild to make it automatic
-Popover with help, and feedback on  parse error
-fix long operators, search is case insensitive
-filterbox placeholders to beta and interactive
-remove not used help button
-Update existing unittests
-unittest and minor fixes


kn-ms:
There is a known failing test. The Popup is clashing with deep render in unittest. I will figure out what the issue, is, but the main part of the code review can be started. It's quiet big anyway

testplan/web_ui/testing/src/Toolbar/FilterBoxPlaceholder.js
```diff
@@ -0,0 +1,17 @@
+import React from 'react';
+
+const FilterBoxPlaceholder = () => <div><center>
```
yuxuan-ms, line 3: <center> tag is not supported in HTML5.

testplan/web_ui/testing/src/Toolbar/__tests__/__snapshots__/Toolbar.test.js.snap
```diff
@@ -10,6 +10,11 @@ exports[`Toolbar inserts extra buttons into the toolbar 1`] = `
   >
     <div
       className="filterBox_1kzbnzj"
+      style={
+        Object {
+          "width": undefined,
```
yuxuan-ms, line 15: why is the with of filterBox is undefined?

testplan/web_ui/testing/src/Report/__tests__/__snapshots__/BatchReport.test.js.snap
```diff
@@ -2229,264 +2229,7 @@ exports[`BatchReport shallow renders the correct HTML structure when report load
...
-        "timer": Object {
-          "run": Object {
-            "end": "2018-10-15T14:30:11.296158+00:00",
-            "start": "2018-10-15T14:30:10.998071+00:00",
-          },
-        },
-        "uid": "520a92e4-325e-4077-93e6-55d7091a3f83",
-      }
-    }
+    report={null}
```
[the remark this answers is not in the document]
kn-ms: That is because now we not simply passing the raw report down, but the filtered one, which get computed if the report is set properly. This testcase just set the state, but not in the right way. I replace it with setState to have it as expected

testplan/web_ui/testing/src/Toolbar/FilterBox.js
```diff
@@ -1,26 +1,225 @@
...
+import SearchFieldParser from "../Parser/SearchFieldParser";
+import { RED } from "../Common/defaults";
+import { Popover, PopoverHeader, PopoverBody, Table } from "reactstrap";
 
 /**
  * Filter box, enter filter expressions to filter entries from the Nav
  * component.
  */
 class FilterBox extends Component {
+  constructor(props) {
```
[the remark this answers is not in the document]
kn-ms: I prefer to have that in a separate pull request, this is big enough to review in this form

testplan/web_ui/testing/src/Toolbar/FilterBoxPlaceholder.js
kn-ms: right, will replace with flex layout

testplan/web_ui/testing/src/Toolbar/__tests__/__snapshots__/Toolbar.test.js.snap
kn-ms: it is not new, just the width used to be a prop of FilterBox component, and when it was undefined it was not rendered to the dom. Now in style it is shown.  Oh and the rootcause of that was that it was not there in the default props used for the test. I will add it.

yuxuan-ms (approved):

PR #523, opened 2020-10-24
Remove duplicated description on assertion pane

raoyitao:
* Only display the description of the current report on assertion pane
  but ignore the descriptions of the parent reports.
* For each NavEntry display the report description as tooltip rather
  than the report name.
* UI layout adjustment.
* update testcases.

### DIFF
--- a/testplan/web_ui/testing/src/Report/reportUtils.js
+++ b/testplan/web_ui/testing/src/Report/reportUtils.js
@@ -216,7 +216,7 @@ const GetCenterPane = (
   selectedEntries
 ) => {
   const logs = state.logs || [];
-  const selectedDescription = selectedEntries.map((element) => {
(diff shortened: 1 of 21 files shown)


testplan/web_ui/testing/src/Report/reportUtils.js
```diff
@@ -216,7 +216,7 @@ const GetCenterPane = (
   selectedEntries
 ) => {
   const logs = state.logs || [];
-  const selectedDescription = selectedEntries.map((element) => {
+  const selectedDescription = selectedEntries.slice(-1).map((element) => {
```
Pyifan, line 219: Can we display description of Testplan/Multitest/Testsuite in assertion panel if they are the bottom-most selection (no testcase selected yet)

Pyifan (changes requested):

Pyifan (approved):

PR #530, opened 2020-11-05
Json report is split by `UID` of report object instead of `name`

raoyitao:
* `uid` shoud be a unique value through out the test report, so when
  a test report is split into smaller ones, the `uid` shoube used as
  reference, however name might be same (currently not allowed maybe
  the rule changes in the future).

### DIFF
--- a/testplan/exporters/testing/json/__init__.py
+++ b/testplan/exporters/testing/json/__init__.py
@@ -138,17 +138,14 @@ class JSONExporter(Exporter):
             """Remove assertions from report and place them in a dictionary."""
             for entry in entries:
                 if entry.get("category") == ReportCategories.TESTCASE:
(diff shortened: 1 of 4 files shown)


Pyifan (changes requested):
For the assertion part, now that we are linking by a unique uid, we do not need the hierarchy info in the assertion file anymore. I suggest that we trim down the file with necessary info only.

raoyitao:
> For the assertion part, now that we are linking by a unique uid, we do not need the hierarchy info in the assertion file anymore. I suggest that we trim down the file with necessary info only.

updated both JSONExporter and webUI, now the dictionary of assertions is simple and takes less space.

Pyifan (approved):

PR #532, opened 2020-11-09
post_testcase exception should be logged in report

raoyitao:
* If a testcase raises the exception is logged, the following epilogues
    should still be executed, but if any of them raises, then the error
    is output by `TESTPLAN_LOGGER`, or the testcase is executed nicely
    then the first exception raised in epilogues will be logged in test
    report. But if the testcase itself raises, exceptions raised from
    epilogues should not overwrite it.

### DIFF
--- a/testplan/testing/multitest/suite.py
+++ b/testplan/testing/multitest/suite.py
@@ -11,6 +11,7 @@ import warnings
 import six
 
 from testplan import defaults
+from testplan.common.utils.logger import TESTPLAN_LOGGER
 from testplan.common.utils import callable as callable_utils
 from testplan.common.utils import interface
 from testplan.common.utils import strings
@@ -499,9 +500,9 @@ def _validate_testcase(func):
         elif not func.name:
             raise ValueError("Testcase name cannot be an empty string")
 
-    except Exception as ex:
+    except Exception as exc:
         _reset_globals()
-        raise ex
+        raise exc
 
     func.name = six.ensure_str(func.name)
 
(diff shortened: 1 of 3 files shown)


testplan/testing/multitest/suite.py
```diff
@@ -805,16 +806,27 @@ def testcase_with_pre(*args, **kwargs):
     return testcase_with_pre
 
 
-def run_epilogues(epilogues, name, *args, **kwargs):
+def _run_epilogues(epilogues, name, raise_exc=False, *args, **kwargs):
```
Pyifan, line 809: This is a bit too many if-else. How about we let _run_epilogues 1) log the exception 2) return the exceptions if any? Then the caller can decide if re-raise?

tests/functional/testplan/testing/multitest/test_suite_decorators.py
```diff
@@ -26,22 +27,24 @@
 
 def pre1(name, self, env, result):
     result.equal(2, 2)
-    result.contain("testcase", name)
+    result.contain("case", name)
```
Pyifan, line 30: Was this assertion originally failing?

Pyifan (changes requested):

testplan/testing/multitest/suite.py
raoyitao: updated. `_run_epilogues` returns all possible exception during running post-testcase functions. if the testcase itself does not raise, then the caller decides to raise the first exception found in epilogues.

tests/functional/testplan/testing/multitest/test_suite_decorators.py
raoyitao: It does not harm because this testcase is not used to check report status, so no matter if the case/assertions is passed or failed.

Pyifan (approved):

PR #533, opened 2020-11-10
check for type dict instead of typing.Mapping in binary deserialization.

kn-ms:
This should be fine, as marshmallow return objects as dicts
and will work when 2.7 typing backport is not available

### DIFF
--- a/testplan/report/testing/schemas.py
+++ b/testplan/report/testing/schemas.py
@@ -3,17 +3,7 @@
 import functools
 import json
 
-from typing import Mapping
-
-import six
 from boltons.iterutils import remap, is_scalar
-
-# pylint: disable=no-name-in-module,import-error
-if six.PY2:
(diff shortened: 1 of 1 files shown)


testplan/report/testing/schemas.py
```diff
@@ -3,18 +3,8 @@
 import functools
 import json
 
-from typing import Mapping
```
yuxuan-ms, line 6: Why not:
```
try:
    from collections.abc import Mapping, Iterable
except ImportError:
    from collections import Mapping, Iterable
```

raoyitao (approved):

testplan/report/testing/schemas.py
kn-ms: We know that what we got it is a dict so no need to do match for abc

Pyifan (approved):

PR #536, opened 2020-11-12
Fix console output error when generating PDF.

yuxuan-ms:


testplan/exporters/testing/pdf/__init__.py
```diff
@@ -232,17 +232,18 @@ def __init__(self, name="PDF exporter", **options):
 
     def export(self, source):
         if len(source):
-            pdf_path = os.path.abspath(create_pdf(source, self.cfg))
+            pdf_path = create_pdf(source, self.cfg)
+            abs_pdf_path = os.path.abspath(pdf_path)
             if pdf_path == self.cfg.pdf_path:
-                self.logger.exporter_info("PDF generated at %s", pdf_path)
+                self.logger.exporter_info("PDF generated at %s", abs_pdf_path)
```
Pyifan, line 238: Would it make better sense if we make create_pdf a member of pdf exporter? And we can log these info in create_pdf where we already checked if we override or rename etc.

Pyifan (approved):

PR #540, opened 2020-11-19
pass runtime info to tests with env.multitest.runtime_info

kn-ms:
In case of parametrized testcases the testcase name could be usefull for debugging, logging purpose
The current change delivering a framework to be able to provide more runtime info later to testcase
the testcase name is accessible as env.multitest.runtime_info.testcase.name

## Solution description

## Checklist:
- [X] Test
- [X] Documentation (API)
- [X] MS info leakage check

### DIFF
--- a/doc/en/multitest.rst
+++ b/doc/en/multitest.rst
@@ -94,7 +94,7 @@ A MultiTest instance can be constructed from the following parameters:
   would typically be defined, as well as drivers for the interfaces that are
   required for interacting with it, such as network connections.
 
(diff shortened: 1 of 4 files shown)


Pyifan (commented):
Hi Kris,

I like this solution much better! I would suggest that we rename MultiTestRuntimeInfoEnvWrapper class, it's a bit long and Env is embedded so I missed it the first time I read through the PR. Maybe call EnvWithRuntimeInfo or something similar?

Pyifan (approved):

PR #543, opened 2020-11-25
Attachment assertion improved design.

kn-ms:
The preview of txt file attachements are not performing well on large files

## Solution description
- use the syntax highlight library with better settings
- do not enable fully display long files
- not even download big files
- with a new ui design
- apply new design to image and normal download representation

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/web_ui/testing/src/AssertionPane/AssertionTypes/AttachmentAssertions.js
```diff
@@ -16,59 +44,66 @@ import TextAttachment from './TextAttachment.js';
...
         />
       );
 
-    case 'jpeg':
-    case 'jpg':
-    case 'bmp':
-    case 'png':
-        return getImageContent(getPath, assertion.description);
-
+    case "jpeg":
```
yuxuan-ms, line 71: I'm not sure. Can we check content-type if it is image/*?

testplan/web_ui/testing/src/Common/fakeReport.js
```diff
@@ -358,6 +358,50 @@ var fakeReportAssertions = {
...
+                            "status_override": null,
+                            "runtime_status": "finished",
+                            "uid": "897ee608-ac20-4c14-bc44-b22969929716",
+                            "entries": [
+                              {
+                                "hash": "ead907197415758fb6a15295336bd51da891a88a",
+                                "line_no": 44,
+                                "category": "DEFAULT",
+                                "meta_type": "entry",
+                                "source_path": "C:\\Users\\KN\\AppData\\Local\\Temp\\tmpc355typr.txt",
```
yuxuan-ms, line 386: set the source_path to "SOURCE_PATH\\tmpc355typr.txt"

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/AttachmentAssertions.js
kn-ms: At this point we do not have the file itself, so to go based on content type we need the backend to set content type. Maybe the result attach can set that from user input. If user do not specify type in the input do semthing similar logic as above just on the backend side. This sounds a good move to me as a second phase, how about open an issuo on this second step?

Pyifan (approved):

PR #548, opened 2020-12-10
Fix an issue that report of Multitest-part could be merged unexpectedly

raoyitao:
* When reports of 2 Multitests have the same name they should not be
  merged unless they are split parts of a Multitest object. However
  the report of Multitest part can unexpected be merged with report of
  another Multitest object, That is certainly wrong. Fix it.
* Code optimization for `TestRunner._merge_reports` with clearer
  error log.
* Add a testcase.
* A simple `uuid4` utility to generate unique id for testplan entities.
* Remove unecessary methods of `LogMatcher`.


testplan/runnable/base.py
```diff
@@ -601,50 +602,60 @@ def _create_result(self):
...
             if report.part and self.cfg.merge_scheduled_parts:
-                # Save the report temporarily and later will merge it
-                test_rep_lookup.setdefault(report.uid, []).append(
-                    (test_results[uid].run, report)
-                )
                 try:
                     test_report.get_by_uid(report.uid)
-                    continue
                 except KeyError:
+                    # Save the report temporarily and later will merge it
```
kn-ms, line 611: This Try/expect/else is definetly a if else written in an uggly way ()handling conditional with exception is not performant, as creating teh stack trace and all the things for an exception is huge

if report.id not in test_report:
...
else:
...

testplan/runnable/base.py
```diff
@@ -669,30 +680,35 @@ def _merge_reports(self, test_report_lookup):
         merge_result = True
 
         for uid, reports in test_report_lookup.items():
-            count = reports[0][1].part[1]  # How many parts scheduled
+            parts = [(rep.part[0], rep.part[1]) for _, rep in reports]
```
kn-ms, line 683: for the validity check this part of the code is traversing the reports/parts list at least six times, I admint that the error message is more specific than it was in previous implementation, but does it worth it?. It might be able to done in one for loop, collecting information, and the some iff making validation, if this detail is needded.

kn-ms (approved):
I approvet this request as I think the logic is good, my two comments are more in code cleannes and performance (though it might have just negligible impact with the report sizes we working). I do not want to hold of this functionality, so I'm ok to geting this merged and a followup issue created, if you do not want to fix this now.

testplan/common/report/base.py
```diff
@@ -192,7 +192,7 @@ def reset_uid(self, uid=None):
         Reset uid of the report, it can be useful when need to generate
         a standard UUID instead of the current one.
         """
-        self.uid = uuid.uuid4() if uid is None else uid
+        self.uid = str(uuid.uuid4()) if uid is None else uid
```
Pyifan, line 195: uid or str(uuid.uuid4())

testplan/runnable/base.py
```diff
@@ -601,50 +602,60 @@ def _create_result(self):
...
+                            " in {rep}".format(uid=report.uid, rep=test_report)
+                        )
+
             elif report.part:
                 report.name = "{} - part({}/{})".format(
                     report.name, report.part[0] + 1, report.part[1]
                 )
-                # Change report uid to avoid conflict during appending
-                report.uid = uuid.uuid4()
+                report.uid = str(uuid.uuid4())  # Change uid to avoid conflict
```
Pyifan, line 650: set to report.name? but this is to be changed soon anyway

testplan/common/report/base.py
```diff
@@ -53,7 +53,7 @@ def __init__(
         self.name = name
         self.description = description
 
-        self.uid = uid or uuid.uuid4()
+        self.uid = uid or str(uuid.uuid4())
```
Pyifan, line 56: I don't like it that we need to add str() wherever we use uuid4(). Can we possibly remove it or extract it to a common util function?

Pyifan (changes requested):

testplan/runnable/base.py
raoyitao: Yes, I'll make a little change to it, making the purpose of code clearer, having less loop.

testplan/common/report/base.py
raoyitao: sure

testplan/runnable/base.py
raoyitao: if report.id in test_report.entry_uids:
  ...
else:
  ...

Use if-esle instead of try-except

testplan/runnable/base.py
```diff
@@ -669,30 +676,43 @@ def _merge_reports(self, test_report_lookup):
...
+            for _, report in reports:
+                if (
+                    report.part[0] < 0
+                    or report.part[0] >= report.part[1]
+                    or report.part[1] != count
+                ):
+                    invalid_param = True
+                    break
+                if report.part[0] in part_indexes:
+                    duplicate_index = True
```
kn-ms, line 693: you can break out from the loop here as well

Pyifan (approved):

PR #550, opened 2020-12-15
Testplan core: bug fix and functional improvement

raoyitao:
* Should check the uid of real test target even if they are wrapped
  in a `Task` object as much as possible to prevent duplicate tests
  being schedule and then exception raised during creating results.
* Fix a bug that `local_runner` cannot execute a callable object.
* Fix a bug that in interactive mode a callable object that returns
  a test entity cannot be correctly added into TestRunner.
* Fix a bug that in interactive mode Multitest-parts cannot work.
* Modify uid of Multitest parts automatically to avoid conflict, and
  set it back as test entity's name for merge sibling reports.
* If fails to merge reports of MultiTest parts then error message will
  be logged and then append sibling reports at the end of plan report.
* Be cautious about the return value of step `_create_result` because
  `run` attribute in returned result of each test entity can be a
  non-boolean value.
* Fix a small issue that when unhandled exception raised during adding
  tests into plan, all tests should abort but function has no `abort`
  method, the extra error log is confusing.
* Fix a potential bug that if scheduled task return None then program
  can enter an infinite recursion.

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -462,7 +462,7 @@ class Entity(logger.Loggable):
         timeout = wait_timeout or self.cfg.abort_wait_timeout
         try:
             self.logger.debug("Aborting {}".format(entity))
-            entity.abort()
+            entity.abort()  # Here entity can be a function and will raise
             self.logger.debug("Aborted {}".format(entity))
         except Exception as exc:
             self.logger.error(traceback.format_exc())
@@ -617,6 +617,7 @@ class RunnableResult(object):
 
     def __init__(self):
         self.step_results = OrderedDict()
+        self.run = False
 
     def __repr__(self):
         return "{}[{}]".format(self.__class__.__name__, self.__dict__)
--- a/testplan/runnable/base.py
+++ b/testplan/runnable/base.py
@@ -4,6 +4,7 @@ import os
 import random
 import time
 import datetime
+import warnings
 import webbrowser
 from collections import OrderedDict
 
@@ -11,7 +12,6 @@ import pytz
 from schema import Or, And, Use
 
 from testplan import defaults
-from testplan.common.utils import logger
 from testplan.common.config import ConfigOption
 from testplan.common.entity import (
     Entity,
@@ -22,8 +22,9 @@ from testplan.common.entity import (
 )
 from testplan.common.exporters import BaseExporter, ExporterResult
 from testplan.common.report import MergeError
-from testplan.common.utils.path import default_runpath
+from testplan.common.utils import logger
 from testplan.common.utils import strings
+from testplan.common.utils.path import default_runpath
 from testplan.exporters import testing as test_exporters
 from testplan.report import (
     TestReport,
@@ -286,6 +287,7 @@ class TestRunner(Runnable):
     def __init__(self, **options):
         super(TestRunner, self).__init__(**options)
         self._tests = OrderedDict()  # uid to resource
+        self._part_instance_names = set()  # name of multitest part
         self._result.test_report = TestReport(
             name=self.cfg.name,
             description=self.cfg.description,
@@ -383,7 +385,7 @@ class TestRunner(Runnable):
             resource, uid=uid or getattr(resource, "uid", strings.uuid4)()
         )
 
-    def schedule(self, task=None, resource=None, uid=None, **options):
+    def schedule(self, task=None, resource=None, **options):
         """
         Schedules a serializable
         :py:class:`~testplan.runners.pools.tasks.base.Task` in a task runner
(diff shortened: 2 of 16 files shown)


testplan/runnable/base.py
```diff
@@ -393,106 +394,130 @@ def schedule(self, task=None, resource=None, uid=None, **options):
...
+            raise RuntimeError(
+                'Resource "{}" does not exist.'.format(resource)
             )
-            return uid
 
-        if isinstance(runnable, Entity):
-            runnable.cfg.parent = self.cfg
+        # The runnable added into TestRunner can be: 1> a real test entity
+        # 2> a task wraps a test entity 3> a callble returns a test entity
+        if isinstance(runnable, Runnable):
```
Pyifan, line 426: I see that this part of code is dup with what we are doing in _verify_test_target. Can we consolidate them?

testplan/runnable/base.py
```diff
@@ -393,106 +394,130 @@ def schedule(self, task=None, resource=None, uid=None, **options):
...
             runnable.parent = self
+            runnable.cfg.parent = self.cfg
         elif isinstance(runnable, Task):
             pass
         elif callable(runnable):
-            runnable.parent_cfg = self.cfg
             runnable.parent = self
+            runnable.parent_cfg = self.cfg
```
Pyifan, line 433: i'm wondering if user pass in a callable, why does it need a parent?

testplan/runners/local.py
```diff
@@ -25,24 +25,28 @@ def __init__(self, **options):
...
-            if not runnable.cfg.parent:
-                runnable.cfg.parent = self.cfg
-        elif isinstance(target, entity.Runnable):
-            runnable = target
+        # Inspect the input type. Tasks must be materialized before
+        # they can be run.
+        if isinstance(runnable, entity.Runnable):
+            target = runnable
+        elif isinstance(runnable, tasks.Task):
+            target = runnable.materialize()
```
Pyifan, line 35: swapping target <==> runnable?

I think it makes better sense to say:
runnable = target.materialize()

testplan/runners/pools/tasks/base.py
```diff
@@ -184,41 +185,51 @@ def materialize(self, target=None):
...
+            else:
+                raise TaskMaterializationError(errmsg.format(self._target))
 
     def _string_to_target(self):
+        """Dynamically load an object from a module by target name."""
         path_inserted = False
-        if isinstance(self._path, six.string_types):
+        if (
+            isinstance(self._path, six.string_types)
+            and self._path not in sys.path
```
Pyifan, line 216: I think the point here is to insert this path at the very beginning of sys.path even if it 's already in sys.path.

testplan/runnable/base.py
```diff
@@ -393,106 +394,130 @@ def schedule(self, task=None, resource=None, uid=None, **options):
...
+            else:
+                target.cfg._options["part"] = None
+                self._tests[uid] = local_runner
+                self.resources[local_runner].add(target, uid)
+            return uid
+
+        # In batch mode uid of test target MUST be unique unless it is
+        # a part of Multitest and is wrapped in a task
+        if uid in self._tests and not (isinstance(runnable, Task) and part):
+            self.logger.error(
```
Pyifan, line 465: I think this is too mild, if user tries to add 2 runnable with the same uid, we shall error out the plan?

testplan/runnable/base.py
```diff
@@ -393,106 +394,130 @@ def schedule(self, task=None, resource=None, uid=None, **options):
...
+        uid = target.uid()
+        part = getattr(target.cfg, "part", None)
 
-        if resource not in self.resources:
-            raise RuntimeError(
-                'Resource "{}" does not exist.'.format(resource)
+        # When running interactively, add all real test targets into the local
+        # runner even if they were scheduled into a pool. It greatly simplifies
+        # the interactive runner if it only has to deal with the local runner.
+        if self.cfg.interactive_port is not None:
```
Pyifan, line 446: Can we check interacitve mode in the beginning of add? If under interactive mode, just ignore the resource param, and all the rest logic shall be just the same? I want to avoid this if self.cfg.interactive_port as much as possible.

Pyifan (changes requested):

testplan/common/entity/base.py
```diff
@@ -462,15 +462,20 @@ def _abort_entity(self, entity, wait_timeout=None):
         timeout = wait_timeout or self.cfg.abort_wait_timeout
         try:
             self.logger.debug("Aborting {}".format(entity))
-            entity.abort()
```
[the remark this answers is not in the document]
raoyitao: updated

testplan/runnable/base.py
raoyitao: yes, the code is moved into _verify_test_target(), to make `add` method more short.

testplan/runnable/base.py
raoyitao: because in LocalRunner or other user implemented Executors, the test entity returned by callable can use TestRunner as its parents, in some unit test, then LocalRunner is not added into TestRunner, then LocalRunner is set as the parent of test entity.

testplan/runners/pools/tasks/base.py
raoyitao: OK, I'll recover it back.

testplan/runners/local.py
raoyitao: That makes sense. In LocalRunner target is materialized to a runnable entity, while in TestRunner the runnable is materialized to a target, They have opposite style. So I will change the code in TestRunner, a entity/task/callable to be added is a target, and it can be materialized to a runnable object.

testplan/runnable/base.py
raoyitao: if user add 2 test entities with the same uid, then error logging. only when they are multitest-parts then TestRunner can accept it, however part should be scheduled as a task.

testplan/runnable/base.py
raoyitao: ah. not very understand about avoid `self.cfg.interactive_port`

testplan/runnable/base.py
```diff
@@ -393,106 +394,124 @@ def schedule(self, task=None, resource=None, uid=None, **options):
...
+                raise ValueError(
+                    '{} with uid "{}" already added.'.format(target, uid)
+                )
+
+        if isinstance(target, Task):
+            if part:
+                self._tests[uid] = None  # placeholder for Multitest parts
+            if target.uid() in self._tests:
+                raise ValueError("Task {} already added.".format(target))
+            uid = target.uid()
```
Pyifan, line 471: Shall we override task's uid with the uid of its content?

testplan/runnable/base.py
```diff
@@ -393,106 +394,124 @@ def schedule(self, task=None, resource=None, uid=None, **options):
...
-            if isinstance(runnable, Task):
-                runnable = runnable.materialize()
-                runnable.cfg.parent = self.cfg
-                runnable.parent = self
-            self.resources[local_runner].add(runnable, runnable.uid() or uid)
-        else:
-            self.resources[resource].add(runnable, uid)
+            if part:
+                # Modify Multitest-part instance to avoid duplicate uid
+                uid = "{} - part({}/{})".format(uid, part[0], part[1])
```
Pyifan, line 445: Also, shall we override multitest uid to use name+part{}/{}?

Each part of multitest shall have a different UID, and they could be merged by their name.

testplan/runnable/base.py
raoyitao: overwritten

testplan/runnable/base.py
raoyitao: updated

testplan/runnable/base.py
raoyitao: after discussion, the new implementation can add multitest part without wrapped in a task.

testplan/runnable/base.py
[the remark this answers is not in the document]
raoyitao: `"{} - part({}/{})".format(uid, part[0], part[1])` is moved to MultiTest::uid()

testplan/common/entity/base.py
```diff
@@ -464,13 +464,19 @@ def _abort_entity(self, entity, wait_timeout=None):
             self.logger.debug("Aborting {}".format(entity))
             entity.abort()
             self.logger.debug("Aborted {}".format(entity))
+        except AttributeError as exc:
```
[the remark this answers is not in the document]
raoyitao: removed

testplan/common/entity/base.py
```diff
@@ -464,13 +464,19 @@ def _abort_entity(self, entity, wait_timeout=None):
...
+            if "abort" in str(exc):  # Here entity might be a function
+                self.logger.debug("Aborted {}".format(entity))
         except Exception as exc:
             self.logger.error(traceback.format_exc())
             self.logger.error(
                 "Exception on aborting {} - {}".format(self, exc)
             )
         else:
-            if wait(lambda: entity.aborted is True, timeout) is False:
+            if (
```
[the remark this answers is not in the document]
raoyitao: removed

testplan/runnable/base.py
```diff
@@ -585,7 +604,9 @@ def _create_result(self):
         test_rep_lookup = {}
 
         for uid, resource in self._tests.items():
-            if not isinstance(self.resources[resource], Executor):
+            if resource is None or not isinstance(
```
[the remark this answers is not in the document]
raoyitao: removed

testplan/runners/base.py
```diff
@@ -64,7 +64,8 @@ def add(self, item, uid):
             self._input[uid] = item
             # `NoRunpathPool` adds item after calling `_prepopulate_runnables`
             # so the following step is still needed
-            self.ongoing.append(uid)
+            if uid not in self.ongoing:
```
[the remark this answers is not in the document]
raoyitao: for safety reason to keep `self._input` and `self.ongoing` synchronized.

testplan/runnable/base.py
```diff
@@ -623,16 +644,16 @@ def _create_result(self):
                     test_rep_lookup[report.uid] = [(run, report)]
                     # Create a report and then append it as a placeholder
                     if isinstance(resource_result, TaskResult):
-                        # 'target' should be an instance of MultiTest since the
+                        # `runnable` must be an instance of MultiTest since the
```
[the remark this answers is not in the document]
raoyitao: The code of create result is simplified.

testplan/runnable/base.py
```diff
@@ -647,7 +668,11 @@ def _create_result(self):
...
 
             test_report.append(report)
-            step_result = step_result and run
+            step_result = (  # `run` can be True/False/None or an exception
+                step_result
+                and not isinstance(run, Exception)
+                and run is not False
+            )
 
         step_result = self._merge_reports(test_rep_lookup) and step_result
```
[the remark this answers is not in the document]
raoyitao: This code about merging report is refactored and only one loop is needed. If failed to merge the original report will be cleared and log the error, while other sibling reports will be appended at the end.

Pyifan (approved):

PR #563, opened 2021-01-08
Standalone log file for driver which produces too much output

raoyitao:
* Multitest driver can setup file handler for logging, however, the
  names of log instances should be different, if there are 2 drivers
  with the same name but not in one Multitest, they would write
  the same content because `logging` is a global shared module
  and getLogger() by name returns the same instance.
* Since that some drivers might produce a large amount of output
  and be propagated to parent's logger, thus there could be too many
  things printed on console as well as in TestRunner's log. Thus make
  an improvement to provide a config option that Testplan can
  generate a standalone file under driver's runpath for logging and
  stop propagation.

### DIFF
--- a/testplan/testing/multitest/driver/base.py
+++ b/testplan/testing/multitest/driver/base.py
@@ -337,7 +337,11 @@ class Driver(Resource):
         formatter = logging.Formatter("%(asctime)s %(levelname)s %(message)s")
         handler = logging.FileHandler(path)
         handler.setFormatter(formatter)
-        logger = logging.getLogger("FileLogger_{}".format(self.cfg.name))
+        logger = logging.getLogger(
+            "FileLogger_{}.{}".format(self.parent.name, self.name)
+            if getattr(self, "parent", None)
+            else "FileLogger_{}".format(self.name)
+        )
         logger.addHandler(handler)
         logger.setLevel(self.logger.getEffectiveLevel())
         self.file_logger = logger
--- a/testplan/testing/multitest/driver/base.py
+++ b/testplan/testing/multitest/driver/base.py
@@ -53,10 +53,10 @@ class DriverConfig(ResourceConfig):
             "name": str,
             ConfigOption("install_files", default=None): Or(None, list),
             ConfigOption("timeout", default=60): int,
-            ConfigOption("logname", default=None): Or(None, str),
             ConfigOption("log_regexps", default=None): Or(None, list),
             ConfigOption("stdout_regexps", default=None): Or(None, list),
             ConfigOption("stderr_regexps", default=None): Or(None, list),
+            ConfigOption("file_logger", default=None): Or(None, str),
             ConfigOption("async_start", default=False): bool,
             ConfigOption("report_errors_from_logs", default=False): bool,
             ConfigOption("error_logs_max_lines", default=10): int,
@@ -90,8 +90,6 @@ class Driver(Resource):
     :type install_files: ``List[Union[str, tuple]]``
     :param timeout: Timeout duration for status condition check.
     :type timeout: ``int``
-    :param logname: Driver logfile name.
-    :type logname: ``str``
     :param log_regexps: A list of regular expressions, any named groups matched
         in the logfile will be made available through ``extracts`` attribute.
         These will be start-up conditions.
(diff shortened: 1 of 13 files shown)


Pyifan (commented):
I suggest that we add set_file_logger param to Driver class, which redirect driver's logger to a separate log file under runpath. Rather than having self.logger & self.file_logger in some drivers.

Btw, some drivers are not using "logname" config option defined in base Driver class: fixclient/server, httpclient/server and a few internal ones.  They are adding another self._logname - this might worth checking & cleaning up.

testplan/testing/multitest/driver/base.py
```diff
@@ -90,7 +91,10 @@ class Driver(Resource):
     :type install_files: ``List[Union[str, tuple]]``
     :param timeout: Timeout duration for status condition check.
     :type timeout: ``int``
-    :param logname: Driver logfile name.
+    :param logname: Driver logfile name. When a binary started by driver makes
```
Pyifan, line 94: Driver logfile base name, in which testplan will look for ``log_regexps`` as driver start-up condition. Default to (driver_name.log?) under (app_path?).

testplan/testing/multitest/driver/base.py
```diff
@@ -100,6 +104,12 @@ class Driver(Resource):
     :type stdout_regexps: ``list`` of ``_sre.SRE_Pattern``
     :param stderr_regexps: Same with log_regexps but matching stderr file.
     :type stderr_regexps: ``list`` of ``_sre.SRE_Pattern``
+    :param file_logger: A standalone file which is placed under the `runpath`
```
Pyifan, line 107: Sends driver's logging to a user specified file under `runpath`. By default, logging goes to console and top level testplan.log.

testplan/testing/multitest/driver/base.py
```diff
@@ -337,34 +348,36 @@ def _setup_file_logger(self, path):
...
-        formatter = logging.Formatter("%(asctime)s %(levelname)s %(message)s")
-        handler = logging.FileHandler(path)
-        handler.setFormatter(formatter)
-        logger = logging.getLogger("FileLogger_{}".format(self.cfg.name))
-        logger.addHandler(handler)
-        logger.setLevel(self.logger.getEffectiveLevel())
-        self.file_logger = logger
-        self.file_logger.propagate = False  # No console logs
+        if self._file_log_handler is not None:
+            raise self.logger.debug(
```
Pyifan, line 352: raise logger.debug()? why debug() instead of raise?

testplan/testing/multitest/driver/base.py
```diff
@@ -337,34 +348,36 @@ def _setup_file_logger(self, path):
...
     def _close_file_logger(self):
         """
-        Closes a logfile previously opened by _setup_file_logger() and removes
-        the logger from self.file_logger. This should be called when the file
+        Closes a handler previously opened by _setup_file_logger() and removes
+        the file handler from self.logger. This should be called when the file
         logger is done with to avoid leaking file handles - typically this
         should be called from stopping().
         """
-        if self.file_logger is None:
```
Pyifan, line 359: Why are we moving this check?

Pyifan (changes requested):

testplan/testing/multitest/driver/base.py
raoyitao: Similar code in TestRunner's `_config_file_logger` there is raise. But I think for a driver some times it will be restarted and so the no need to raise an exception but just log a message.

testplan/testing/multitest/driver/base.py
raoyitao: this `self.file_logger` is the same thing as `self.logger` and it has a handler to write to file. now the handler is `self._file_log_handler`. if `self.cfg.file_logger` is set then `self._file_log_handler` is not None, we'll close the handler.

testplan/testing/multitest/driver/base.py
raoyitao: after discussion we should raise here strictly. It's necessary to make sure that stop driver before restart and will not see this exception.

Pyifan (approved):

PR #564, opened 2021-01-12
URL Navigation

kn-ms:
The url on teh ui does not preserv navigation state

## Solution description
url routing on the ui is now the source of navigation state

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/web_ui/testing/src/Report/BatchReport.js
+++ b/testplan/web_ui/testing/src/Report/BatchReport.js
@@ -1,7 +1,7 @@
 import React from 'react';
 import {StyleSheet, css} from 'aphrodite';
 import axios from 'axios';
-import { Redirect } from 'react-router-dom';
+import PropTypes from 'prop-types';
 
 import Toolbar from '../Toolbar/Toolbar';
 import {TimeButton} from '../Toolbar/Buttons';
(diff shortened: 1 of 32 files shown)


testplan/web_ui/testing/src/Report/BatchReport.js
```diff
@@ -73,9 +73,12 @@ class BatchReport extends React.Component {
       report: processedReport,
       filteredReport,
       selectedUIDs,
-      lastManualSelectedUIDs: selectedUIDs,
       loading: false,
       logs: report.logs,
+    }, () => {
+      this.props.history.replace(
```
Pyifan, line 79: What is this history used for?

Pyifan (commented):
Like this change!

testplan/web_ui/testing/src/Report/BatchReport.js
kn-ms: history representiung the browser history, it is injected by the router. If you push or replace things in it that made the browser to actuially navigate. Push equivalent of redirect while replace is similar, just the original url is removed from history and replaced with the new one.

kn-ms:
I will close and open a new one so the migrated travis job is picked up correctly

PR #572, opened 2021-01-22
Fix several bugs in interactive mode

raoyitao:
* Make it support Multitest-parts: 1> Should correctly set Multitest UID
  in returned JSON report; 2> Multitest-part name should be correctly
  displayed (with part information) on NavEntry.
* Deal with special symbol like "slash" in Multitest name or suite
  name, need to encode the URI strings and decode them on server side.
* For error result and unstable result, the NavEntry's color is still
  black, need to add some CSS style definition for interactive mode.
* If a test suite has 'setup' or 'teardown', after executing them the
  `runtime_status` should be set to "finished", or the test suite's
  `runtime_status` is still "running" and UI shows an incorrect icon.
* Update version of "immer" library.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/testing/multitest/base.py
```diff
@@ -1033,7 +1032,7 @@ def _run_testcases_iter(self, testsuite, testcases):
         """
         pre_testcase = getattr(testsuite, "pre_testcase", None)
         post_testcase = getattr(testsuite, "post_testcase", None)
-        testsuite_uid = testsuite.name
+        testsuite_uid = testsuite.name  # Unique suite name in Multitest as UID
```
Pyifan, line 1035: I think it is confusing to assign testsuite.name to another local variable called testsuite_uid. Can we simply use testsuite.name where it is needed?

testplan/web_ui/testing/src/Nav/InteractiveNavList.js
```diff
@@ -14,7 +15,7 @@ const InteractiveNavList = (props) => {
     props,
     (entry) => (
       <InteractiveNavEntry
-        name={entry.name}
+        name={_.isEmpty(entry.part) ? entry.name : entry.uid}
```
Pyifan, line 18: I thought this is before reset_uid() so we could always use entry.uid, no?

Pyifan (changes requested):

testplan/runnable/interactive/http.py
```diff
@@ -32,6 +34,17 @@
 from .reloader import ModuleReloader
 
 
+def decode_uri_component(func):
+    """Decode URI component before arguments passed to url route handler."""
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
```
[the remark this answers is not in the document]
raoyitao: I think this is guaranteed by @api.route that when a class is decorated, its 'get' and 'set' methods are processed, they can receive an instance as args[0], and several named arguments as kwargs.

testplan/web_ui/testing/src/Nav/InteractiveNavList.js
raoyitao: seems we can use entry.uid

testplan/testing/multitest/base.py
raoyitao: I think the author want to emphasize that unique id is used as identifier when transferred through network. let me think about it again.

Pyifan (dismissed):

Pyifan (approved):

PR #574, opened 2021-01-22
Skip tests on win

kn-ms:
the desision to skip a test on windows is done in many different way, it can be consolidated
also there are many skiped testcases which are simple to fix

## Solution description
- make a consolidated partial decorator to skip on windows
- use the new decorator when something skiped on windows

- with mock reloader get the interactive unittests to work again

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


tests/functional/testplan/runnable/interactive/test_api.py
```diff
@@ -65,32 +62,39 @@ def test_parametrized(self, env, result, val):
...
-        interactive_port=0,
-        interactive_block=False,
-        exporters=[XMLExporter(xml_dir=tmpdir / "xml_exporter")],
-    )
 
-    logfile = tmpdir / "attached_log.txt"
-    logfile.write_text(
-        "This text will be written into the attached file.", encoding="utf8"
-    )
+    with patch(
```
Pyifan, line 66: Does this fix interactive mode tests in test_plan_top.py? That's great but why...? ;)

kn-ms: the pytest environment is really tricky when it comes to the special logic in reloader to build the dependency tree of imported modules, and we do not really test the reloader here, so mocking it out  we do not need to solve this for these cases. We achieve this by the constructor returning None, which means this interactive mode could not do hot reloads.

Pyifan (approved):

PR #580, opened 2021-02-09
use base64url in interactive nav for ids, and remove beta ui

kn-ms:
interactive mode using entity names as uid which is not url safe
beta ui is no more needed

## Solution description
- use base64url encoding of uids on routing, that is safe and no urlencode needed
- remove beta ui and it's dependencies

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


kn-ms:
Closing and reopening to trigger travis job to run again. For some reason I could not retrigger the travis run.

yuxuan-ms (approved):

Pyifan:
Hi Krisztian, the PR looks good to me. Please feel free to resolve and merge. Just one quick thought: Yitao put in the double encoding fix also for interactive url (that's for communication between UI and web server). Wonder if we want to consolidate to use just solution for similar use cases?

kn-ms:
> Hi Krisztian, the PR looks good to me. Please feel free to resolve and merge. Just one quick thought: Yitao put in the double encoding fix also for interactive url (that's for communication between UI and web server). Wonder if we want to consolidate to use just solution for similar use cases?

I agree on the consolidation. I see the effect of double encoding, I think a bas64url is more straightforward (though not human readable) solution and I prefer that. On the ui routing layers does some encoding when storing url in history, but not the other way around, so things might need asymmetric treating with double urlencodeing.

Pyifan (approved):

PR #583, opened 2021-02-23
Add highlight for EqualAssertion

yuxuan-ms:
Add highlight for EqualAssertion.
Also fix some eslint errors.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/entries/assertions.py
+++ b/testplan/testing/multitest/entries/assertions.py
@@ -155,6 +155,16 @@ class Equal(FuncAssertion):
     label = "=="
     func = operator.eq
 
+    def __init__(self, first, second, description=None, category=None):
+        self.type_actual = type(first).__name__
+        self.type_expected = type(second).__name__
+        super(Equal, self).__init__(
+            first=first,
+            second=second,
+            description=description,
+            category=category,
+        )
+
 
 class NotEqual(FuncAssertion):
     label = "!="
--- a/testplan/web_ui/testing/src/AssertionPane/AssertionTypes/__tests__/BasicAssertion.test.js
+++ b/testplan/web_ui/testing/src/AssertionPane/AssertionTypes/__tests__/BasicAssertion.test.js
@@ -22,6 +22,26 @@ function propsEqual() {
   };
 }
 
+function propsHighlightEqual() {
+  return {
+    assertion: {
+      "category": "DEFAULT",
+      "machine_time": "2019-02-12T17:41:42.795536+00:00",
+      "description": "compare two string",
+      "line_no": 25,
+      "label": "==",
+      "second": "This is a short thing",
+      "meta_type": "assertion",
+      "passed": false,
+      "type": "Equal",
+      "utc_time": "2019-02-12T17:41:42.795530+00:00",
+      "first": "This is a long thing with some extra at the end",
+      "type_actual": "str",
+      "type_expected": "str"
+    }
+  };
+}
+
 function propsNotEqual() {
   return {
     assertion: {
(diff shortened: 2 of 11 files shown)


testplan/testing/multitest/entries/assertions.py
```diff
@@ -155,6 +155,46 @@ class Equal(FuncAssertion):
...
+                }
+                self.highlight_expected = {
+                    "eq": (0, common_chrs_cnt),
+                    "ne": (common_chrs_cnt, len(self.second)),
+                }
+            else:
+                self.highlight_actual = {
+                    "ne": (common_chrs_cnt, len(self.first))
+                }
+                self.highlight_expected = {
```
kn-ms, line 191: This here creating a set instead of a dictionary.

Since this logic is always highlighting from 0 to b and b to end it is enough to store the common char count.

Though I feel this might be just a display details, and might not need to be in t he assertion at all.

testplan/testing/multitest/entries/assertions.py
```diff
@@ -155,6 +155,46 @@ class Equal(FuncAssertion):
...
+        self.type_expected = type(second).__name__
+        super(Equal, self).__init__(
+            first=first,
+            second=second,
+            description=description,
+            category=category,
+        )
+
+    def evaluate(self):
+        if isinstance(self.first, str) and isinstance(self.second, str):
```
kn-ms, line 171: No unittest for this logic

kn-ms (commented):
reconsider implement the diffing on the display side.

this can help to produce an even richer output https://www.npmjs.com/package/diff

yuxuan-ms:
> reconsider implement the diffing on the display side.
> 
> this can help to produce an even richer output https://www.npmjs.com/package/diff

The data in the JSON may not be the original value. So I think comparing values in python is more accurate

testplan/testing/multitest/entries/assertions.py
yuxuan-ms: Let's just save the "b" in the assertion?

kn-ms:
> > reconsider implement the diffing on the display side.
> > this can help to produce an even richer output https://www.npmjs.com/package/diff
> 
> The data in the JSON may not be the original value. So I think comparing values in python is more accurate

This sounds even worst, as you might passing highlight data not related to the data in the json, so it might be invalid

yuxuan-ms:
> > > reconsider implement the diffing on the display side.
> > > this can help to produce an even richer output https://www.npmjs.com/package/diff
> > 
> > 
> > The data in the JSON may not be the original value. So I think comparing values in python is more accurate
> 
> This sounds even worst, as you might passing highlight data not related to the data in the json, so it might be invalid

OK, let me try jsdiff

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/__tests__/BasicAssertion.test.js
```diff
@@ -22,6 +22,26 @@ function propsEqual() {
...
 
+function propsHighlightEqual() {
+  return {
+    assertion: {
+      "category": "DEFAULT",
+      "machine_time": "2019-02-12T17:41:42.795536+00:00",
+      "description": "Hel == Hello",
+      "line_no": 25,
+      "label": "==",
+      "second": "Hello",
```
kn-ms, line 33: Add a more complex example, that utilize more codepaths

first: "This is a short things",
second: "And this is a long things with some extra at the end"

kn-ms:
Hate to pester you on this, but it must have been late yesterday when I wrote the examples, can you remove the 's' from the end of things in the example sentences, that would look a bit more English

kn-ms (approved):

PR #585, opened 2021-02-24
DictMatch should support empty string or None as dict key

raoyitao:
* If empty string or `None` is used as dictionary key and try matching
  dictionary, then the assertion data in generated report is incorrect,
  should use a special ABSENT to indicate that key does not exist.
* If `None` is used as dictionary key, UI should be aware of this or
  error occurs.

### DIFF
--- a/testplan/common/utils/convert.py
+++ b/testplan/common/utils/convert.py
@@ -1,6 +1,8 @@
 """Conversion utilities."""
 import itertools
 
+from .reporting import Absent
+
 
 def make_tuple(value, convert_none=False):
     """Shortcut utility for converting a value to a tuple."""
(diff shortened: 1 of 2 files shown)


testplan/common/utils/convert.py
```diff
@@ -93,30 +95,33 @@ def expand_values(rows, level=0, ignore_key=False, key_path=None):
     """
     if key_path is None:
         key_path = []
+
     for row in rows:
-        key = row[0] if ignore_key is False else ""
-        if key:
+        key = row[0] if ignore_key is False else Absent
```
Pyifan, line 100: Add a comment that Absent means ignore?

Pyifan (approved):

PR #602, opened 2021-04-01
Fix status value lost in deserialize testcase object.

yuxuan-ms:
deserialize  JSON report will lost status if the testcase doesn't contain assertion.

## Solution description
Set status in TestcaseReportSchema post_load method

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/report/testing/schemas.py
+++ b/testplan/report/testing/schemas.py
@@ -121,7 +121,7 @@ class TestCaseReportSchema(ReportSchema):
     entries = fields.List(EntriesField())
 
     category = fields.String(dump_only=True)
-    status = fields.String(dump_only=True)
+    status = fields.String()
     runtime_status = fields.String(dump_only=True)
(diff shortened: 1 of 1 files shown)


Pyifan (approved):

testplan/report/testing/schemas.py
```diff
@@ -121,7 +121,7 @@ class TestCaseReportSchema(ReportSchema):
     entries = fields.List(EntriesField())
 
     category = fields.String(dump_only=True)
-    status = fields.String(dump_only=True)
+    status = fields.String()
```
Pyifan, line 124: Please double check if we need to load status for other report entries as well.

Pyifan (changes requested):

Pyifan (approved):

Pyifan (approved):

PR #604, opened 2021-04-01
Sort testcases without break down parametrization groups.

raoyitao:
* For example, a test suite have child [case1, param1, param2, case2],
  after shuffled this can be [case2, param2, case1, param1]. however
  the desired result should be [case2, param2, param1, case1], the
  testcases in the same param group should still be next to each other.
* this rule should also appplies to other kind or sorting, for example:
  alphanumeric sorting.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/testing/ordering.py
+++ b/testplan/testing/ordering.py
@@ -68,7 +68,7 @@ class BaseSorter(object):
     def sort_testsuites(self, testsuites):
         raise NotImplementedError
 
-    def sort_testcases(self, testcases):
+    def sort_testcases(self, testcases, param_groups=None):
         raise NotImplementedError
 
     def sorted_instances(self, instances):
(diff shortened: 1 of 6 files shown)


testplan/testing/ordering.py
```diff
@@ -65,25 +65,57 @@ def should_sort_testcases(self):
     def sort_instances(self, instances):
         raise NotImplementedError
 
-    def sort_testsuites(self, testsuites):
+    def sort_testsuites(self, instance, testsuites):
```
Pyifan, line 68: I don't see self/instance being use, do we need them?

testplan/testing/ordering.py
```diff
@@ -65,25 +65,57 @@ def should_sort_testcases(self):
...
         raise NotImplementedError
 
-    def sort_testcases(self, testcases):
+    def sort_testcases(
+        self, instance, testsuite, testcases, param_groups=None
```
Pyifan, line 72: I don't see self/instance/testsuite being use, do we need them?

Pyifan (changes requested):

Pyifan (approved):

PR #607, opened 2021-04-08
Use JSON5 to parse json which contains NaN value.

yuxuan-ms:
The JSON string  which generated by json.dumps in python will contain NaN. But chrome cannot parse it by JSON.parse.

## Solution description
Use JSON5 https://json5.org/

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/web_ui/testing/src/Common/utils.js
+++ b/testplan/web_ui/testing/src/Common/utils.js
@@ -2,6 +2,7 @@
  * Common utility functions.
  */
 import {NAV_ENTRY_DISPLAY_DATA} from "./defaults";
+import JSON5 from "json5";
 import _ from 'lodash';
 
 /**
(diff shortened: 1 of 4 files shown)


testplan/web_ui/testing/src/Common/utils.js
```diff
@@ -174,3 +175,22 @@ export const toPlainObjectIn = (obj, enumerableOnly = false) => {
...
+ */
+export const transformResponse = (data) => {
+  let result = data;
+  if (typeof data === 'string' && data.length) {
+    try {
+      result = JSON5.parse(result);
+    } catch (e) {
+      console.log("Cannot parse to json");
+      console.log(e);
+    }
```
kn-ms, line 192: Would the string be any useful for the consumer of this transformation. Instead of throwing the exception from here would not this result cryptic exception from elsewhere that need to be tracked back down to here?

yuxuan-ms: print console.trace() ?

kn-ms: More thinking to let the exception thrown and handle that appropriately where called, as it can be different based on the usecase.

testplan/web_ui/testing/src/Common/utils.js
```diff
@@ -174,3 +175,22 @@ export const toPlainObjectIn = (obj, enumerableOnly = false) => {
 export const joinURLComponent = (base, component) => {
   return `${base.replace(/\/+$/, '').trim()}/${component}`;
 };
+
+/**
+ * Parse JSON string to object by JSON5 https://json5.org/
+ * @param {string} data - The json string
+ * @returns {object}
+ */
+export const transformResponse = (data) => {
```
kn-ms, line 184: the name transformResponse is to generic, can it be called to parseToJson or something more meaningfull?

kn-ms (approved):

PR #612, opened 2021-04-13
Fix an issue that --tags cannot work well in interactive mode

raoyitao:
* In interactive mode, the server receives request from client and
  creates a pattern filter based on testsuite & testcases names, but
  this filter should be applied after the filters of test runner is
  applied, otherwise some unwanted testcases will be executed.
* Use `test_context` property instead of calling `get_test_contest`
  method as much as possible for performance.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/base.py
+++ b/testplan/testing/multitest/base.py
@@ -271,14 +271,11 @@ class MultiTest(testing_base.Test):
         else:
             return self.cfg.name
 
-    def get_test_context(self, test_filter=None):
+    def get_test_context(self):
         """
         Return filtered & sorted list of suites & testcases
         via `cfg.test_filter` & `cfg.test_sorter`.
 
-        :param test_filter: Class with test filtering logic, used for filtering
-            testcase in interactive mode while in batch mode it is ``None``.
-        :type test_filter: :py:class:`~testplan.testing.filtering.BaseFilter`
         :return: Test suites and testcases belong to them.
         :rtype: ``list`` of ``tuple``
         """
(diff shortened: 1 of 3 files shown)


testplan/testing/multitest/base.py
```diff
@@ -271,24 +271,23 @@ def uid(self):
...
-            testcase in interactive mode while in batch mode it is ``None``.
-        :type test_filter: :py:class:`~testplan.testing.filtering.BaseFilter`
         :return: Test suites and testcases belong to them.
         :rtype: ``list`` of ``tuple``
         """
         ctx = []
-        sorted_suites = self.cfg.test_sorter.sorted_testsuites(self.cfg.suites)
+        sorted_suites = (
+            self.cfg.suites
+            if getattr(self, "strict_order", False)
```
Pyifan, line 285: but multitest doesn't have strict_order attr?

testplan/testing/multitest/base.py
```diff
@@ -298,18 +297,16 @@ def get_test_context(self, test_filter=None):
...
-            # In batch mode no testcase should be removed if `strict_order` is
-            # specified, unless all of the testcases have been filtered out.
-            if (
-                test_filter is None
-                and getattr(suite, "strict_order", False)
-                and 0 < len(testcases_to_run) < len(sorted_testcases)
-            ):
+            # In batch mode if `strict_order` is specified, then either
+            # all of the testcases are filtered out, or left unchanged.
+            if getattr(suite, "strict_order", False) and 0 < len(
```
Pyifan, line 307: move strict_order check -> sorted_testcases

Pyifan (approved):

PR #620, opened 2021-04-26
Fix an issue that post-testcase doen not run if pre-testcase raises

raoyitao:
* If post-testcase method is defined, even if pre-testcase method or
  testcase itself raises an exception, all those post-testcase method
  should still run.
* @pre_testcase and @post_testcase can be defined in any order.
* Update testcase.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


Pyifan:
Before we proceed with this PR, let's decide on whether we want to keep @pre/post_testcase decorators or the suite class pre_testcase/post_testcase member function. If we go with the latter one, let's get rid of @pre/post_testcase related implementation and update the documentation accordingly.

PR #622, opened 2021-04-28
Remove compatibility related code since no need to support Python 2

raoyitao:
* No need to support Python 2, the related code for adapting Py2 & Py3
  can be removed, such as condition of sys version, Python 2 specific
  feature and all those 3rd libraries for compatibility/backporting.
  - six
  - future
  - past
  - builtins
  - nonzero
  - next
  - mock
  - queue
  - cStringIO/cPickle
  - functools32
  - io.open
  - http.server
  - zip/zip_longest
  - sys.version/sys.hexversion/sys.verion_info
  - ImportError
* Remove unused import at head of files.
* remove unused libraries from dependency.
* Use "utf-8" instead of "UTF-8", "utf8", "utf_8".


testplan/exporters/testing/pdf/renderers/base.py
```diff
@@ -8,12 +8,9 @@
...
 from reportlab.platypus import Paragraph
 
-try:
-    from html import escape as _orig_escape
+from html import escape as _orig_escape
 
-    escape = functools.partial(_orig_escape, quote=False)
-except ImportError:
-    from cgi import escape
+escape = functools.partial(_orig_escape, quote=False)
```
kn-ms, line 13: This is just used in a single place, no need to overwrite it with the partial.

testplan/exporters/testing/pdf/renderers/entries/assertions.py
```diff
@@ -1,12 +1,9 @@
...
 import functools
 
-try:
-    from html import escape as _orig_escape
+from html import escape as _orig_escape
 
-    escape = functools.partial(_orig_escape, quote=False)
-except ImportError:
-    from cgi import escape
+escape = functools.partial(_orig_escape, quote=False)
```
kn-ms, line 6: Not worth to override it for a single use

tests/unit/testplan/exporters/testing/test_sliced_paragraph.py
```diff
@@ -1,11 +1,8 @@
 import functools
 
-try:
-    from html import escape as _orig_escape
+from html import escape as _orig_escape
 
-    escape = functools.partial(_orig_escape, quote=False)
-except ImportError:
-    from cgi import escape
+escape = functools.partial(_orig_escape, quote=False)
```
kn-ms, line 5: as before no need for redefinition

kn-ms (approved):
I have 3 nice to have comments

testplan/exporters/testing/pdf/renderers/base.py
raoyitao: fixed

tests/unit/testplan/exporters/testing/test_sliced_paragraph.py
raoyitao: fixed

testplan/exporters/testing/pdf/renderers/entries/assertions.py
raoyitao: Unlike other 2 ``escape``, in this file ``escape`` is used many times, I renamed this wrapped to ``html_escape``

kn-ms (approved):

kn-ms (approved):

PR #626, opened 2021-05-04
Added new functions to LogMatcher

markokitonjics:

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


Pyifan (changes requested):
hi Marko,

PR looks good overall! It would be better if you could fill out the description and tick the check boxes.

There was a tricky issue Yitao once had with match_between on Windows, but I cannot remember the details. I will ask Yitao to review and comment on this PR next Mon. 

Mark as 'request changes' to avoid accidental merge. Thanks!

tests/unit/testplan/common/utils/test_match.py
```diff
@@ -152,6 +152,73 @@ def test_match_all(self, basic_logfile):
         assert matches[0].group(0) == "fourth"
         assert matches[1].group(0) == "fifth"
 
+    def test_not_match(self, basic_logfile):
```
raoyitao, line 155: there is already a ``test_not_match`` in this file, did I miss something ?

tests/unit/testplan/common/utils/test_match.py
```diff
@@ -152,6 +152,73 @@ def test_match_all(self, basic_logfile):
...
+        """Does the LogMatcher raise an exception when match is found."""
+        matcher = LogMatcher(log_path=basic_logfile)
+        matcher.not_match(regex=re.compile(r"bob"), timeout=0.5)
+        matcher.seek()
+        with pytest.raises(Exception):
+            matcher.not_match(regex=re.compile(r"third"), timeout=0.5)
+
+    def test_match_all(self, basic_logfile):
```
raoyitao, line 163: it seems that there is already a ``test_match_all`` in this file.

testplan/common/utils/match.py
```diff
@@ -230,3 +230,97 @@ def match_all(self, regex, timeout=5):
                 raise
 
         return matches
+
+    def match_between(self, regex, mark1, mark2):
```
raoyitao, line 234: We had got a piece of code like this, but I remembered that it cannot work well on Windows because when reading file in text mode the behavior of ``tell`` is not as expected. Kris had mentioned something:

-----------------------------------
9th Dec 2020 9:20:22 pm Notaisz, Krisztian: 

the problem is that on my windows with my python version tell and seek working with raw file position, while read (when in text mode) reading characters, and a \r\n in windos is just transformed to \n automatically by python
so if I seek just before a newline (on windows)

f.tell()  => x
f.read(1) => '\n'
f.tell() => x+2

so you read one char which is a newline represented by two char in the file which is consumed and file position increased by 2
but again, this is not defined to be like that, you should not have assumption on the result of tell it is just represent a position

https://docs.python.org/3/tutorial/inputoutput.html
 
f.tell() returns an integer giving the file object’s current position in the file represented as number of bytes from the beginning of the file when in binary mode and an opaque number when in text mode. 
well you might have the assumption that positions are comparable as it say opaque numbers
---------------------------------------

So I think that the issue of dealing with '\r\n' is still there, or point to me if I missed something?

raoyitao (commented):
See my comments

testplan/common/utils/match.py
kn-ms: The previous implementation was try to get a difference of two position, then read that amount of chars. The problem with that was that the difference are in bytes while the read is per char bases which might read 2 bytes in case of '\r\n'

This implementation does not do any difference, it just read from pos1 till the we pass pos2, the only assumption it does that the positions are comparable.

tests/unit/testplan/common/utils/test_match.py
markokitonjics: Good catch! Not sure what happened there.

Pyifan (approved):

PR #628, opened 2021-05-06
Pypi actions

kn-ms:

## Solution description

## Checklist:
- [ ] Test
- [x] Documentation (API)
- [x] MS info leakage check
- [x] For new driver: driver index page

### DIFF
--- a/tests/unit/testplan/web_ui/test_web_app.py
+++ b/tests/unit/testplan/web_ui/test_web_app.py
@@ -77,108 +77,92 @@ def _create_tmp_data_files(base_dir):
     )
 
 
-class TestStaticEndpoints(object):
-    """
-    Test the endpoints returning files from the STATIC_PATH directory.
-    """
-
-    def setup_method(self, _):
-        """Create fake static files and a test client."""
-        self.static_dir = tempfile.mkdtemp()
-        _create_tmp_static_files(self.static_dir)
-        tp_web_app.config["STATIC_PATH"] = self.static_dir
-        tp_web_app.config["TESTPLAN_REPORT_NAME"] = "report.json"
-        tp_web_app.config["TESTING"] = True
-        tp_web_app.static_folder = os.path.abspath(
-            os.path.join(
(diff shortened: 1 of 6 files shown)


tests/unit/testplan/web_ui/test_web_app.py
```diff
@@ -77,108 +77,92 @@ def _create_tmp_data_files(base_dir):
...
+    # Use correct UID, expect 200 response.
+    path = "/api/v1/reports/{}".format(DATA_REPORTS["testplan"]["uid"])
+    response = webapp_test_client.get(path)
+    expected_contents = str(DATA_REPORTS["testplan"]["contents"])
+    assert response.status_code == 200
+    assert expected_contents in str(response.data)
+
+    # Use incorrect UID, still expect 200 response.
+    response = webapp_test_client.get("/api/v1/reports/123")
+    assert response.status_code == 200
```
Pyifan, line 146: I thought we are returning a 404 here?

tests/unit/testplan/web_ui/test_web_app.py
```diff
@@ -77,108 +77,92 @@ def _create_tmp_data_files(base_dir):
...
-        path = "/api/v1/reports/123/attachments/attached.file"
-        response = self.client.get(path)
-        expected_contents = str(DATA_REPORTS["testplan"]["contents"])
-        assert response.status_code == 200
-        assert expected_contents in str(response.data)
+
+def test_testplan_attachment(webapp_test_client):
+    """
+    Does sending anything to /api/v1/reports/<uid>/attachments/<uid> respond with
+    501.
```
Pyifan, line 162: docstring seems outdated?

Pyifan (changes requested):

tests/unit/testplan/web_ui/test_web_app.py
kn-ms: This is the --ui flow here. This case there is only one report that just generated during the run, so the report id do not count too much.

Pyifan (approved):

PR #633, opened 2021-05-14
Formatted execution times

markokitonjics:
Execution times were shown in seconds.

## Solution description
Formatted execution times.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


kn-ms (dismissed):

testplan/web_ui/testing/src/Common/utils.js
```diff
@@ -96,10 +96,49 @@ function domToString(dom) {
  * @param {string} uid - string to be encoded
  * @returns {string}
  */
-function encodeURIComponent2 (str) {
+function encodeURIComponent2(str) {
   return encodeURIComponent(encodeURIComponent(str));
 }
 
+/**
```
Pyifan, line 103: I wonder if there's any library we could use instead of implementing this / 60 & % 60 logic ourselves?

Pyifan (changes requested):

testplan/web_ui/testing/src/Common/utils.js
kn-ms: If remember well that is just used in the api communication in interactive mode, I think we can switch it to base64url.

testplan/web_ui/testing/src/Common/utils.js
```diff
@@ -96,10 +96,49 @@ function domToString(dom) {
...
+  var hours = (durationInMilliseconds - minutes) / 60;
+
+  const isDisplayedHours = hours > 0;
+  const isDisplayedMinutes = (minutes > 0) | isDisplayedHours;
+  const isDisplayedSeconds = (seconds > 0) | isDisplayedMinutes;
+
+  var hoursDisplay = isDisplayedHours ? hours + 'h' : '';
+  var minutesDisplay = isDisplayedMinutes ? minutes + 'm' : '';
+  var secondsDisplay = isDisplayedSeconds ? seconds + 's' : '';
+  var millisecondsDisplay = milliseconds + 'ms';
```
yuxuan-ms, line 135: var -> let
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/let

Pyifan (approved):

PR #639, opened 2021-05-19
Fix an issue that ``skip_if`` cannot work for parametrized testcases

raoyitao:
* the parametrized testcase is generated with an ``__name__``, however,
  this name can be a temporary one and be the same as parametrized
  template, so at this time ``skip_if`` decorator can associate the
  predicates with wrong testcase. Another thing is that the generated
  testcase will be renamed to ensure it has a unique ``__name__`` thus
  predicates information will be lost. Fix this issue by associating
  predicates directly with atteibute  ``__skip__`` of the testcase.
* Update testcases.

## Checklist:
- [x] Test
- [x] Documentation (API)
- [ ] MS info leakage check


doc/en/multitest.rst
```diff
@@ -94,6 +94,26 @@ A MultiTest instance can be constructed from the following parameters:
     def post_testcase(self, name, env, result):
         pass
 
+  :py:func:`@testcase <testplan.testing.multitest.suite.skip_if>` decorator can
+  be used to define the precondition to determine if a testcase needs to run.
+  It accepts one or several functions, if any of them returns True, then the
+  decorated testcase will not be executed.
+
```
Pyifan, line 101: typo: @testcase --> @skip_if, suggested change:

:py:func:`@skip_if<testplan.testing.multitest.suite.skip_if>` decorator can be used to annotate a testcase. It take one or more predicates, and if any of them evaluated to True, then  the testcase will be skipped by MultiTest instead of being normally executed. The predicate’s signature must name the argument “testsuite” or a MethodSignatureMismatch will be raised.

Pyifan (changes requested):

Pyifan:
I suggest that we consolidate the implementation for @skip_if & @known_to_fail, by setting the attr on the testcase method itself. It feel hacky to achieve this by counting on the sequence number.

doc/en/multitest.rst
raoyitao: modified

raoyitao:
> I suggest that we consolidate the implementation for @skip_if & @known_to_fail, by setting the attr on the testcase method itself. It feel hacky to achieve this by counting on the sequence number.

changed, each function has a __skip__ attribute which is a list to hold those predicates.

Pyifan (approved):

PR #640, opened 2021-05-19
Fix naming conflict.

yuxuan-ms:
naming conflict of queue

## Solution description
Rename queue to msg_queue

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/common/utils/sockets/fix/server.py
+++ b/testplan/common/utils/sockets/fix/server.py
@@ -578,15 +578,15 @@ class Server(object):
         if self.log_callback:
             self.log_callback("Flushed received message queues")
 
-    def _flush_queue(self, queue):
+    def _flush_queue(self, msg_queue):
         """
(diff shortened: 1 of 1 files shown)


testplan/common/utils/sockets/fix/server.py
```diff
@@ -20,7 +20,7 @@ class ConnectionDetails(object):
     """
 
     def __init__(
-        self, connection, name=None, queue=None, in_seqno=1, out_seqno=1
+        self, connection, name=None, msg_queue=None, in_seqno=1, out_seqno=1
```
kn-ms, line 23: no need to change the public api for this fix, as some might use it with keyword args. It can just remains queue in the constructor, it is not a problem if that hides the package.

Pyifan (approved):

PR #641, opened 2021-05-21
Check if App is already stopped

markokitonjics:
When a `App`/`Driver` was stopped manually inside a test, the testplan cleanup process threw a warning because it assumed the Resource was still running and tried to stop it.

## Solution description
When calling the `stop()` function on an `App` or `Driver` nothing will happen if it's already stopped.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/driver/app.py
+++ b/testplan/testing/multitest/driver/app.py
@@ -307,7 +307,7 @@ class App(Driver):
     def stopping(self):
         """Stops the application binary process."""
         super(App, self).stopping()
-        # 
+        #
         if self.proc is None:
             return
         try:
--- a/testplan/testing/multitest/driver/app.py
+++ b/testplan/testing/multitest/driver/app.py
@@ -307,6 +307,7 @@ class App(Driver):
     def stopping(self):
         """Stops the application binary process."""
         super(App, self).stopping()
+        # 
         if self.proc is None:
             return
         try:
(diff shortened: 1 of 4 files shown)


testplan/testing/multitest/driver/app.py
```diff
@@ -307,6 +307,8 @@ def ensure_app_running_while_extracting_values():
     def stopping(self):
         """Stops the application binary process."""
         super(App, self).stopping()
+        if self.proc is None:
```
Pyifan, line 310: I think the proc's None check could go to kill_process util?

tests/unit/testplan/testing/multitest/driver/myapp/test_app.py
```diff
@@ -326,3 +325,24 @@ def run_app(cwd, runpath):
...
+        binary="echo",
+        args=["hello"],
+        shell=True,
+        runpath=runpath,
+    )
+    app.start()
+    app.wait(app.STATUS.STARTED)
+    assert app.status.tag == app.status.STARTED
+
+    app.stop()
```
Pyifan, line 343: When the stop is called for the first time, self.proc is not None. And when the stop is called for the 2nd time, stop would see STOPPED status and return immediately. Thus I think the code change in App driver is not really exercised here? Is this testcase duplicate with the one for driver?

Pyifan (changes requested):

Pyifan (approved):

PR #645, opened 2021-05-27
Add Plotly chart support

yuxuan-ms:
Add an assertion for displaying [Plotly](https://plotly.com/) in report

## Solution description
Use plotly.js on web ui & save plotly figure object on attachment file. Adjust ag-grid column auto resize setting

## Checklist:
- [x] Test
- [x] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/common/serialization/fields.py
+++ b/testplan/common/serialization/fields.py
@@ -1,6 +1,7 @@
 """
 Custom marshmallow fields.
 """
+import abc
 import pprint
 import warnings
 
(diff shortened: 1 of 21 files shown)


testplan/common/serialization/fields.py
```diff
@@ -33,6 +34,37 @@
...
+        return {
+            "link": self.link,
+            "title": self.title,
+            "new_window": self.new_window,
+            "type": "link",
+        }
+
+
 def _repr_obj(obj):
     # copypasta from unittest code
```
kn-ms, line 69: Could not resist to leave a comment here: This is how we got spaghetti 😄

Pyifan (approved):

Pyifan (dismissed):

Pyifan (approved):

PR #649, opened 2021-06-04
Added custom db_path argument for SQLite driver

markokitonjics:
db_path for SQLite3 driver was not customizable.

## Solution description
Added custom db_path argument for SQLite3 driver.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/examples/DB/Sqlite3/test_plan.py
+++ b/examples/DB/Sqlite3/test_plan.py
@@ -112,7 +112,7 @@ def main(plan):
     :return: Testplan result object.
     :rtype:  ``testplan.base.TestplanResult``
     """
-    my_db_path = os.path.join(tempfile.gettempdir(), "new_db_name")
+    my_db_path = tempfile.gettempdir()
     plan.add(
         MultiTest(
             name="Sqlite3Test",
--- a/examples/DB/Sqlite3/test_plan.py
+++ b/examples/DB/Sqlite3/test_plan.py
@@ -57,8 +57,8 @@ class DBQueries(object):
                 items,
             )
         result.log(
-            'Database file "{}" of driver "{}" created at "{}"'.format(
-                env.db.cfg.db_name, env.db.cfg.name, env.db.db_path
+            'Database file of driver "{}" created at "{}"'.format(
+                env.db.cfg.name, env.db.db_path
             )
         )
 
@@ -112,13 +112,14 @@ def main(plan):
     :return: Testplan result object.
     :rtype:  ``testplan.base.TestplanResult``
     """
-    my_db_path = tempfile.gettempdir()
+    db_file_name = "mydb"
+    my_db_path = os.path.join(tempfile.gettempdir(), db_file_name)
     plan.add(
         MultiTest(
             name="Sqlite3Test",
             suites=[DBQueries()],
             environment=[
-                Sqlite3(name="db", db_name="mydb", db_path=my_db_path)
+                Sqlite3(name="db", db_path=my_db_path)
             ],
         )
     )
--- a/examples/DB/Sqlite3/test_plan.py
+++ b/examples/DB/Sqlite3/test_plan.py
@@ -117,7 +117,9 @@ def main(plan):
         MultiTest(
             name="Sqlite3Test",
             suites=[DBQueries()],
-            environment=[Sqlite3(name="db", db_name="mydb", db_path=my_db_path)],
+            environment=[
+                Sqlite3(name="db", db_name="mydb", db_path=my_db_path)
+            ],
         )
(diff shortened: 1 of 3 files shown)


testplan/testing/multitest/driver/sqlite.py
```diff
@@ -24,6 +24,7 @@ def get_options(cls):
         """
         return {
             "db_name": str,
+            ConfigOption("db_path", default=None): str,
```
Pyifan, line 27: We need docstring for new config options.

examples/DB/Sqlite3/test_plan.py
```diff
@@ -110,11 +112,14 @@ def main(plan):
...
     :rtype:  ``testplan.base.TestplanResult``
     """
+    my_db_path = os.path.join(tempfile.gettempdir(), "new_db_name")
     plan.add(
         MultiTest(
             name="Sqlite3Test",
             suites=[DBQueries()],
-            environment=[Sqlite3(name="db", db_name="mydb")],
+            environment=[
+                Sqlite3(name="db", db_name="mydb", db_path=my_db_path)
```
Pyifan, line 121: It's usage is a bit weird, as ``db_name`` is required, however its value is ignored.

examples/DB/Sqlite3/test_plan.py
```diff
@@ -110,11 +112,14 @@ def main(plan):
     :return: Testplan result object.
     :rtype:  ``testplan.base.TestplanResult``
     """
+    my_db_path = os.path.join(tempfile.gettempdir(), "new_db_name")
```
Pyifan, line 115: Consider using an context mgr here so that the tmp dir is removed after plan run?

Pyifan (changes requested):

examples/DB/Sqlite3/test_plan.py
markokitonjics: We could mark db_name as deprecated and then:

- if db_path is not provided it would default to runpath + driver name
- if db_path is a relative path it would try to connect to runpath + db_path
- if db_path is an absolute path we would try to connect to it

What do you think?

Pyifan: Yuxuan suggested a solution, which I think is neat. We could reuse the db_name config option, it can take a relative or an absolute path -- in the former case, we assume it is relative to runpath, in the latter case, we would just use it directly. This would allow user to place the db outside the runpath and persist that.

A more advanced usage would be for user to specify the runpath and path_cleanup=False for the driver.

kn-ms: I think this is the same as Marko's suggestion, just reuse the db_name parameter. This way users do not need to change code because of deprecation, but the being the db_name a path it makes confusion for any new user.

markokitonjics: I agree with Krisztian. Although it would be nice to reuse the db_name argument, it wouldn't really be a "db name" but more like a "path to the db". Other then that your solution is exactly the same as ours so we do agree on that.

I think we should implement your suggestion under an argument called "db_path". Let us know if you disagree.

Pyifan: Oh, now I re-read I realize they are indeed the same. Uh... between 'breaking the user' and 'have a confusing param name', i'm slightly leaning towards the latter one. For most of users, they are still going to just provide the base name of the db, and it works like a hidden feature that 'db_name' could take a abs path as well. I think the docstring of this param would really matter to explain the usage and behavior. Anyway, i'm not keen on that, your call.

examples/DB/Sqlite3/test_plan.py
markokitonjics: `gettempdir()` is the temporary directory defined by the OS, it shouldn't be removed.

testplan/testing/multitest/driver/sqlite.py
markokitonjics: Yep :)

testplan/testing/multitest/driver/sqlite.py
```diff
@@ -68,7 +78,19 @@ def __init__(self, name, db_name, connect_at_start=True, **options):
     @property
     def db_path(self):
```
Pyifan, line 79: The db_path in config option is the dir portion of the db_path, but the db_path property is the full db_path. I don't think this is good.

Let's use just db_path in config option then, remove db_name.

Pyifan (changes requested):

Pyifan (approved):

PR #653, opened 2021-06-09
tpsreport cli

kn-ms:
tpsreport tool to be able to convert other framework results to testplan format

## Solution description
- refactored some of the result import logic
- cli added to import cppunit/gtest and export testplan json

## Checklist:
- [X] Test
- [ ] Documentation (API)
- [X] MS info leakage check


Pyifan:
This look great, very python 3! I just have some generic questions: 1) Would the script be available for non-venv user? 2) the download_report script, can it benefit from the new importer/exporter utilties as well? 3) do we have documentation for this command line tool?

Pyifan (approved):

PR #663, opened 2021-06-22
Driver name conflicts with internal testplan named entities

raoyitao:
* If there's a driver (e.g. called "logger" or "first") in Multitest
  environment, unfortunately the `Environment` object itself has a
  member with the same name, so when Testplan tries resolving the name
  in context, it will not get the expected driver. `RuntimeEnvironment`
  is presented in user's testcase by argument `env` and we make it
  fetch the drivers with priority by changing `__setattr__` method.
* `Environment` provides a `get` method for compatibility issue.
* Prevent user from modifying context of drivers, or raise.
* Fix an invalid expression in `testplan.environment.Environments`.
* Fix an issue that `Environment.__contains__` does not check the
  initial context.
* `Environment` is not a subclass of `Entity`, so some properties are
  removed from this class and should explicitly access them through
  the `parent`, also `initial_context` is owned by `Environment`.

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -9,8 +9,8 @@ import signal
 import time
 import threading
 import psutil
-from collections import deque, OrderedDict
 import traceback
+from collections import deque, OrderedDict
 
 from schema import Or, And, Use
 
@@ -31,30 +31,11 @@ class Environment(object):
     """
 
     def __init__(self, parent=None):
-        self._resources = OrderedDict()
-        self.parent = parent
-        self.start_exceptions = OrderedDict()
-        self.stop_exceptions = OrderedDict()
-        self._logger = None
-
-    @property
-    def cfg(self):
-        """Configuration obejct of parent object."""
-        return self.parent.cfg if self.parent else None
-
-    @property
-    def runpath(self):
-        """Runpath of parent object."""
-        return self.parent.runpath if self.parent else None
-
-    @property
-    def logger(self):
-        if self._logger is None:
-            if self.parent is not None:
-                self._logger = self.parent.logger
-            else:
-                self._logger = Entity.logger
-        return self._logger
+        self.__dict__["parent"] = parent
+        self.__dict__["_initial_context"] = {}
+        self.__dict__["_resources"] = OrderedDict()
+        self.__dict__["start_exceptions"] = OrderedDict()
+        self.__dict__["stop_exceptions"] = OrderedDict()
 
     def add(self, item, uid=None):
         """
@@ -71,9 +52,14 @@ class Environment(object):
         """
         if uid is None:
             uid = item.uid()
-        item.context = self
+        if uid in dir(self):
+            raise ValueError(
+                f'Identifier "{uid}" is reserved and cannot be used as UID.'
+            )
         if uid in self._resources:
-            raise RuntimeError("Uid {} already in context.".format(uid))
+            raise RuntimeError(f'Uid "{uid}" already in environment.')
+
+        item.context = self
         self._resources[uid] = item
         return uid
 
@@ -86,38 +72,63 @@ class Environment(object):
     def first(self):
         return next(uid for uid in self._resources.keys())
 
-    def __getattr__(self, item):
-        context = self.__getattribute__("_resources")
+    def get(self, key, default=None):
+        # For compatibility reason, acts like a dictionary which has
+        # a `get` method that returns `None` if no attribute found.
+        try:
+            return self.__getitem__(key)
+        except AttributeError:
+            return default
 
-        if item in context:
-            return context[item]
+    def __getattr__(self, name):
+        resources = self.__getattribute__("_resources")
+        initial_context = self.__getattribute__("_initial_context")
 
-        if self.parent and self.parent.cfg.initial_context:
-            if item in self.parent.cfg.initial_context:
-                return self.parent.cfg.initial_context[item]
+        if name in resources:
+            return resources[name]
 
-        return self.__getattribute__(item)
+        if name in initial_context:
+            return initial_context[name]
+
+        raise AttributeError(
+            "'{}' object has no attribute '{}'".format(
+                self.__class__.__name__, name
+            )
+        )
+
+    def __setattr__(self, name, value):
+        if name in self.__dict__:
+            self.__dict__[name] = value
+        elif name in self.__getattribute__("_resources"):
+            raise RuntimeError(
+                'Cannot modify resource "{}" in environment.'.format(name)
+            )
+        elif name in self.__getattribute__("_initial_context"):
+            raise RuntimeError(
+                'Cannot modify attribute "{}" in initial context.'.format(name)
+            )
+        else:
+            super(Environment, self).__setattr__(name, value)
 
     def __getitem__(self, item):
-        return getattr(self, item)
+        return self.__getattr__(item)
 
     def __contains__(self, item):
-        return item in self._resources
+        if item in self.__getattribute__(
+            "_resources"
+        ) or item in self.__getattribute__("_initial_context"):
+            return True
+        else:
+            return False
 
     def __iter__(self):
         return iter(self._resources.values())
 
     def __repr__(self):
-        if self.parent and self.parent.cfg.initial_context:
-            ctx = self.parent.cfg.initial_context
-            initial = {key: val for key, val in ctx.items()}
-            res = {key: val for key, val in self._resources.items()}
-            initial.update(res)
-            return "{}[{}]".format(self.__class__.__name__, initial)
-        else:
-            return "{}[{}]".format(
-                self.__class__.__name__, list(self._resources.items())
-            )
+        initial = {key: val for key, val in self._initial_context.items()}
+        res = {key: val for key, val in self._resources.items()}
+        initial.update(res)
+        return "{}[{}]".format(self.__class__.__name__, initial)
 
     def __len__(self):
         return len(self._resources)
@@ -147,7 +158,8 @@ class Environment(object):
                 msg = "While starting resource [{}]\n{}".format(
                     resource.cfg.name, traceback.format_exc()
                 )
-                self.logger.error(msg)
+                if self.parent and hasattr(self.parent, "logger"):
+                    self.parent.logger.error(msg)
                 self.start_exceptions[resource] = msg
                 # Environment start failure. Won't start the rest.
                 break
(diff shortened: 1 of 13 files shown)


testplan/common/entity/base.py
```diff
@@ -96,10 +101,14 @@ def __getattr__(self, item):
...
-        return self.__getattribute__(item)
+        raise AttributeError(
+            "'{}' object has no attribute '{}'".format(
+                self.__class__.__name__, item
+            )
+        )
 
     def __getitem__(self, item):
-        return getattr(self, item)
+        return self.__getattr__(item)
```
raoyitao, line 111: Mark: in context resolving this work, but still `env.abc` cannot work in testcases.

kn-ms: what if the add would be enhanced to do sanity checks and refuse to add drivers whose name clash with members/inital context keys?

raoyitao: `Environment` class has many members (although only `logger`, `first`, `parent` might be used as driver name). So I think the agent class `RuntimeEnvironment` for user's testcase can help to fetch the driver correctly, while Testplan internal code mostly uses `Environment`. There could still be name conflicts (fortunately `RuntimeEnvironment` only has 2 menbers `_environment` and `runtime_info`)

Pyifan: > what if the add would be enhanced to do sanity checks and refuse to add drivers whose name clash with members/inital context keys?

I would like to have that too!

testplan/testing/py_test.py
```diff
@@ -302,6 +302,16 @@ def _build_pytest_args(self):
         return pytest_args
 
 
+class PyTestRuntimeInfo(testing.RuntimeInfo):
+    """
+    This class provides information about the state of the actual test run
+    that is accessible from the testcase through the environment as:
+    ``env.pytest_runtime_info``. Not used for now.
```
Pyifan, line 309: I prefer not to adding code until they are really needed.

Pyifan (changes requested):

Pyifan:
Generally speaking, i would like to keep RuntimeEnvironment simple -- just a passing through layer. Because we are not getting RuntimeEnvironment everywhere (e.g before/after_start etc, this might need some fix up but is a separate topic). Let's get the issues solved in Environment class.

1) remove logger other unnecessary property
2) add() to check name duplication
3) __getattr__ could search in _resource -> initial_context -> parent(maybe?)

testplan/common/entity/base.py
```diff
@@ -86,6 +92,14 @@ def remove(self, uid):
     def first(self):
         return next(uid for uid in self._resources.keys())
 
+    def get(self, item):
```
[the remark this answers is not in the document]
raoyitao: modified: def get(self, item, default=None):

testplan/common/entity/base.py
```diff
@@ -96,13 +110,29 @@ def __getattr__(self, item):
...
+            "'{}' object has no attribute '{}'".format(
+                self.__class__.__name__, item
+            )
+        )
+
+    def __setattr__(self, name, value):
+        if name in self.__dict__:
+            self.__dict__[name] = value
+        elif name in self.__getattribute__("_resources"):
+            raise RuntimeError('Cannot modify "{}" in context.'.format(name))
```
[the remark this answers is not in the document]
raoyitao: use 'environment' instead of 'context' in error message. similar change also made in `add` method.

testplan/common/entity/base.py
```diff
@@ -96,13 +110,29 @@ def __getattr__(self, item):
...
-        return getattr(self, item)
+        return self.__getattr__(item)
 
     def __contains__(self, item):
-        return item in self._resources
+        if item in self._resources:
+            return True
+        else:
+            if self.parent and self.parent.cfg.initial_context:
+                return item in self.parent.cfg.initial_context
```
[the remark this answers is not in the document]
raoyitao: yes.

testplan/common/utils/context.py
```diff
@@ -65,12 +65,12 @@ def __call__(self, ctx):
         """
         if ctx is None:
             raise ValueError(
-                "Could not retrieve driver {0} value from "
+                'Could not retrieve driver "{}" value from '
```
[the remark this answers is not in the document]
raoyitao: good idea

testplan/environment.py
```diff
@@ -97,12 +97,12 @@ def add(self, env, uid):
     def starting(self):
         """Start all added environments."""
         for env in self._envs:
-            env.start()
+            self._envs[env].start()
```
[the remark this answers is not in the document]
raoyitao: self._envs[uid].start()

testplan/testing/py_test.py
raoyitao: removed from PyTest

testplan/testing/multitest/base.py
```diff
@@ -61,14 +60,14 @@ def iterable_suites(obj):
     return suites
 
 
-class MultiTestRuntimeInfo(object):
+class MultiTestRuntimeInfo(testing_base.RuntimeInfo):
```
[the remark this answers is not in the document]
raoyitao: To keep it simple I have removed the implementation. the only change is rename 'multitest_runtime_info' to 'runtime_info' because this member is already in 'MultiTestInfo' class.

testplan/common/entity/base.py
raoyitao: I think that in `add` we need to check if the `uid` already in resources and parent's initial_context (we have not done this and I'll add some code for it), but it's not smart to check duplicates of its own member, unless we can make sure from now to the future this class is fixed and only have several members, just like we check testcase's name in `suite.py`.

testplan/common/entity/base.py
```diff
@@ -86,6 +92,14 @@ def remove(self, uid):
...
+    def get(self, item):
+        # For compatibility reason, acts like a dictionary which has
+        # a `get` method that returns `None` if no attribute found.
+        try:
+            return self.__getitem__(item)
+        except AttributeError:
+            return None
+
     def __getattr__(self, item):
         context = self.__getattribute__("_resources")
```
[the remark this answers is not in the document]
raoyitao: ah, I have not change these lines, I found nothing wired, maybe it's dizzy there's a lot of methods stating with __getXXX here. Changed to `self.__dict__["_resources"]`.

testplan/common/entity/base.py
```diff
@@ -31,11 +32,11 @@ class Environment(object):
...
-        self._resources = OrderedDict()
-        self.parent = parent
-        self.start_exceptions = OrderedDict()
-        self.stop_exceptions = OrderedDict()
-        self._logger = None
+        self.__dict__["_resources"] = OrderedDict()
+        self.__dict__["parent"] = parent
+        self.__dict__["start_exceptions"] = OrderedDict()
+        self.__dict__["stop_exceptions"] = OrderedDict()
+        self.__dict__["_logger"] = None
```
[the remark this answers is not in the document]
raoyitao: since that there's already so much data members and methods in this class, so I think the absence of these 3 will not 马make substantial change but bring extra trouble

testplan/common/entity/base.py
```diff
@@ -86,6 +92,14 @@ def remove(self, uid):
     def first(self):
```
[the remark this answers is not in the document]
raoyitao: I think this method is used to get the most frequently used resource, but not always local runner (something else can also executes test). I agree that it's not a good assumption but removing it will cause a lot of other changes so I will not do it in this PR.

testplan/common/entity/base.py
```diff
@@ -96,13 +110,29 @@ def __getattr__(self, item):
             if item in self.parent.cfg.initial_context:
```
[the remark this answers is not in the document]
raoyitao: do you mean that we still write code in this way:

plan.add(MultiTest("Initial Context example", [SimpleSuite()], environment=[ContextUser(name="context_user")], initial_context=INITIAL_CONTEXT))

while in Multitest __init__() we put initial_context into the environment ?

testplan/common/entity/base.py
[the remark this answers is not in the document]
raoyitao: removed properties and in our code use "parent.abc" to access it.

testplan/common/entity/base.py
```diff
@@ -31,30 +32,11 @@ class Environment(object):
...
-    @property
-    def logger(self):
-        if self._logger is None:
-            if self.parent is not None:
-                self._logger = self.parent.logger
-            else:
-                self._logger = Entity.logger
-        return self._logger
+        self.__dict__["parent"] = parent
+        self.__dict__["initial_context"] = {}
```
Pyifan, line 36: initial_context --> _initial_context

testplan/testing/multitest/base.py
```diff
@@ -80,23 +79,32 @@ def __init__(self):
...
+        environment: entity.Environment,
+        runtime_info: MultiTestRuntimeInfo,
     ):
         self.__dict__["_environment"] = environment
-        self.__dict__["multitest_runtime_info"] = runtime_info
+        self.__dict__["runtime_info"] = runtime_info
 
     def __getattr__(self, attr):
-        return getattr(self._environment, attr)
+        # Instance of `RuntimeEnvironment` is presented in user's testcase
```
Pyifan, line 101: It is not the purpose of this wrapper to control access to the _environment. I think this could simply be restored to "getattr"

testplan/common/entity/base.py
```diff
@@ -71,9 +53,14 @@ def add(self, item, uid=None):
         """
         if uid is None:
             uid = item.uid()
-        item.context = self
+        if uid in ("parent", "initial_context"):
```
Pyifan, line 56: compare with __ dict __ instead?

testplan/common/entity/base.py
```diff
@@ -86,38 +73,56 @@ def remove(self, uid):
...
+    def __getattr__(self, name):
         context = self.__getattribute__("_resources")
 
-        if item in context:
-            return context[item]
+        if name in context:
+            return context[name]
+
+        if name in self.initial_context:
+            return self.initial_context[name]
```
Pyifan, line 92: can we also get runpath, logger, cfg from parent here?

testplan/common/entity/base.py
```diff
@@ -96,13 +110,29 @@ def __getattr__(self, item):
...
+                self.__class__.__name__, item
+            )
+        )
+
+    def __setattr__(self, name, value):
+        if name in self.__dict__:
+            self.__dict__[name] = value
+        elif name in self.__getattribute__("_resources"):
+            raise RuntimeError('Cannot modify "{}" in context.'.format(name))
+        else:
```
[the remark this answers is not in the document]
raoyitao: add an 'elif'

testplan/common/entity/base.py
raoyitao: modified

testplan/testing/multitest/base.py
raoyitao: simplified

Pyifan (approved):

Pyifan (approved):

PR #664, opened 2021-06-25
Add FormattedValue type in TableLog assertion

yuxuan-ms:
* Add FormattedValue type in TableLog assertion
* Add internal link support
* Fix some small bugs of ag-grid

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TableAssertions/tableAssertionUtils.js
+++ b/testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TableAssertions/tableAssertionUtils.js
@@ -1,8 +1,119 @@
 /** @module tableAssertionUtils */
 
 import React from "react";
-import {domToString} from './../../../Common/utils';
+import { useParams, Link } from "react-router-dom";
+import _ from "lodash";
+import {
+  domToString,
+  generateURLWithParameters
+} from './../../../Common/utils';
 
+// Small component for rendering TableLogLinkCell.
+function LinkRender(props) {
+  const { uid } = useParams();
+  if (props.inner) {
+    const link = generateURLWithParameters(
+      window.location,
+      props.link.startsWith("/")
+        ? `/testplan/${uid}${props.link}`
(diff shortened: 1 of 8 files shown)


testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TableAssertions/tableAssertionUtils.js
```diff
@@ -30,22 +61,65 @@ function tableCellStyle(params) {
...
     }
   }
-  return cell.value;
+  return valueA >= valueB ? 1 : -1;
 }
 
 
+export function processCellForClipboard(param) {
+  if (typeof param.value === "object") {
+    switch (param.value.type) {
```
kn-ms, line 111: we have this switch case too many times in this file which means the behavior of the link or formatted type is describe at least 3 places.  As the data is preprocessed before passed to the table, we could create actual classes of these types which implement an interface that is:

- able to return the renderer for itself
- able to return the value for comparison
- return the value for csv export

this way if a new type coming only a single class need to be created that implementing the type and that only need to be hooked up with the parser in a single place.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/TableAssertions/tableAssertionUtils.js
```diff
@@ -1,8 +1,119 @@
...
+    this.value = cell.title || cell.link;
+  }
+
+  render() {
+    return <LinkRender {...this.cell}/>;
+  }
+
+  compare(linkB) {
+    if (linkB instanceof TableLogLinkCell) {
+      return this.value >= linkB.value ? 1 : -1;
```
kn-ms, line 49: Maybe not a concern but comparison should be w 3 way thing, treating equal similar as greater might mean that ordering back and forth will result different order.

Pyifan (approved):

PR #670, opened 2021-07-07
Add LogLink docs & example.

yuxuan-ms:

## Solution description

## Checklist:
- [ ] Test
- [x] Documentation (API)
- [ ] MS info leakage check


Pyifan:
I suggest that we remove the top level test_plan.py, and make each testsuite an independent test_plan_xxx.py

Pyifan (approved):

PR #672, opened 2021-07-19
Remove customized code for generating temp directory

raoyitao:
* Code base has totally moved to Python3 and the fallback method
  `_TemporaryDirectory` is no longer useful at all, remove it.
* An environment variable is added to suuport customize `runpath`
  for Pytest.
* Specify runpath for test examples.
* Some small code optimization.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


tests/unit/testplan/testing/multitest/driver/mykafka/test_kafka.py
```diff
@@ -39,11 +39,6 @@ def zookeeper_server():
 @pytest.fixture(scope="module")
 def kafka_server(zookeeper_server):
     server = kafka.KafkaStandalone("kafka", cfg_template=kafka_cfg_template)
-
```
Pyifan, line 42: does this test pass?

Pyifan: Also this test has a problem that it doesn't properly set runpath for this server

Pyifan (changes requested):
If we run test_plan_top.py internally, I would like to include oss tests as well. It seems this is not handle in this pr currently.

testplan/base.py
```diff
@@ -449,9 +449,16 @@ class TestplanMock(Testplan):
     def __init__(self, *args, **kwargs):
         # mock testplan could run in threads
         kwargs.setdefault("abort_signals", [])
-        kwargs.setdefault("runpath", default_runpath_mock)
         kwargs.setdefault("parse_cmdline", False)
         kwargs.setdefault("logger_level", logger.DEBUG)
 
+        if os.environ.get("TEST_ROOT_RUNPATH"):
```
Pyifan, line 455: Shall this if-else logic go to default_runpath_mock?

testplan/common/utils/path.py
```diff
@@ -51,21 +51,21 @@ def workspace_root():
     return pwd()
 
 
-def default_runpath(entity):
+def default_runpath(entity=None):
```
Pyifan, line 54: i think this default_runpath is only meant to be used to decide the default runpath for TestRunner. For other entities, the right logic to determine the runpath, shall at least take a look at its parent. I would prefer NOT to change this API for our test code usage.

tests/conftest.py
```diff
@@ -2,49 +2,83 @@
...
+
+
+def _generate_runpath():
+    """
+    Generate a temporary directory unless specified by environment variable.
+    """
+    if os.environ.get("TEST_ROOT_RUNPATH"):
+        yield tempfile.mkdtemp(dir=os.environ["TEST_ROOT_RUNPATH"])
+    else:
+        parent_runpath = default_runpath()
```
Pyifan, line 30: I think it's fine to just use tempfile behavior if TEST_ROOT_RUNPATH is not specified. default_runpath() isn't meant to be used here.

tests/conftest.py
```diff
@@ -2,49 +2,83 @@
...
+    """
+    yield from _generate_runpath()
 
 
 @pytest.fixture(scope="function")
 def mockplan():
     """
-    Return a temporary TestplanMock for testing. Some components needs a
-    testplan for getting runpath and cfg.
+    Return a temporary TestplanMock for testing. Some components need a
```
Pyifan, line 70: Shall mockplan just take a runpath fixture as input, rather than repeating the logic?

Pyifan (changes requested):

testplan/base.py
raoyitao: good idea

tests/conftest.py
raoyitao: sure.

tests/conftest.py
raoyitao: recoverd

testplan/common/utils/path.py
raoyitao: recovered

Pyifan (approved):

PR #675, opened 2021-07-22
Update schema library to later version

raoyitao:
* Some users work with schema whose version >= 0.7.0, however there
  is a known issue which makes it incompatible with current version
  0.6.6 because in several places we have passed callable objects as
  default value of `schema.Optional`, but in the latest version these
  callables will be instantiated during schema validation. So we have
  to solve this issue.
* Fix a security issue reported from CodeQL scanning.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/common/config/base.py
+++ b/testplan/common/config/base.py
@@ -12,7 +12,7 @@ from testplan.common.utils import logger
 
 # A sentinel object meaning not defined, it is useful when you need to
 # handle arbitrary objects (including None).
-ABSENT = Optional._MARKER
+ABSENT = Optional._MARKER  # pylint: disable=protected-access
 
 
 def validate_func(*arg_names):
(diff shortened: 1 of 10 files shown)


testplan/common/config/base.py
```diff
@@ -43,7 +70,7 @@ def ConfigOption(key, default=ABSENT):
     option after we exhaust the entity's containers.
     """
 
-    optional = Optional(key, default=default)
+    optional = _Optional(key, default=default)
```
Pyifan, line 73: Can we not pass default to schema library at all, and handle the default value later in our own code?

Pyifan (changes requested):

testplan/common/config/base.py
raoyitao: The current solution handle the default values in `Config.__init__`, it's our own code. Or do you mean put the `default` in `ConfigOption` but not `_Optional`, however, how can we get all of the `ConfigOption`s (including all that from parents) after validating, that's more complicated. Another reason is that schema libraries might be updated later and it might be able to handle inherited classes of `Optional` just like `Hook` (I think it should be)

raoyitao: Now add a `custom_default` attributes to `Optional` instance and make the original argument `default` always ABSENT.

Pyifan (approved):

PR #679, opened 2021-07-29
Improvement in reloading tasks in interactive mode

raoyitao:
* If argument `extra_deps` is specified, the paths from where to
  import these modules should be added to `sys.path` for reloading.
  Also this ``list`` object can have both paths or modules.
* When scheduling a task, the argument `module` MUST be specified,
  and nested target in the module could be fetched, like "a.b.c".
* Refactor `RunnableManager`, `TestRunner` and `Testplan`.
* Add some missing arguments and fix annotations.
* Update testcases.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -5,13 +5,13 @@ configuration, start/stop/run/abort, create results and have some state.
 
 import os
 import sys
-import signal
 import time
+import signal
 import threading
-import psutil
 import traceback
 from collections import deque, OrderedDict
 
+import psutil
 from schema import Or, And, Use
 
 from testplan.common.config import Config, ConfigOption
@@ -19,6 +19,7 @@ from testplan.common.utils.thread import execute_as_thread
 from testplan.common.utils.timing import wait
 from testplan.common.utils.path import makeemptydirs, makedirs, default_runpath
 from testplan.common.utils.strings import slugify, uuid4
+from testplan.common.utils.validation import is_subclass, has_method
 from testplan.common.utils import logger
 
 
@@ -611,7 +612,6 @@ class RunnableConfig(EntityConfig):
     def get_options(cls):
         """Runnable specific config options."""
         return {
-            # Interactive needs to have blocked propagation.
             # IHandlers explicitly enable interactive mode of runnables.
             ConfigOption("interactive_port", default=None): Or(None, int),
             ConfigOption(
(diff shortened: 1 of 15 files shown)


testplan/common/entity/base.py
```diff
@@ -1123,21 +1143,15 @@ class RunnableManager(Entity):
     :py:class:`Runnable <testplan.common.entity.base.Runnable>` entity
     in a separate thread and handles the abort signals.
 
-    :param parse_cmdline: Parse command lne arguments.
+    :param parse_cmdline: Parse command line arguments.
     :type parse_cmdline: ``bool``
```
kn-ms, line 1147: now the parser argument is no more in the docs, is it intentional?

kn-ms (commented):
Could not we keep the legacy way of working if no module is specified then the target is considered as that all but last component is the module as currently, but mark that path as deprecated for a while, then deco it in a month time or so?

testplan/common/entity/base.py
raoyitao: The original doc string has argument `resources`, `runner` and `parser`, but they don't exist in config options. I add them back, but seems only `resources` and `runner` are reasonably needed here, `parser` is not necessary in a `RunnableManager`, so is `parse_cmdline`. I would like to move them to `Testplan` class, is that OK ?

raoyitao:
> Could not we keep the legacy way of working if no module is specified then the target is considered as that all but last component is the module as currently, but mark that path as deprecated for a while, then deco it in a month time or so?

Oh yes, although there's some ambiguity but some user might wrote code that way, let me update it to support both way.

kn-ms:
> > Could not we keep the legacy way of working if no module is specified then the target is considered as that all but last component is the module as currently, but mark that path as deprecated for a while, then deco it in a month time or so?
> 
> Oh yes, although there's some ambiguity but some user might wrote code that way, let me update it to support both way.

we can make that deprecated so eventually the ambiguity removed, just give some heads-up

testplan/common/entity/base.py
raoyitao: Finally I decide to remove `parser` from TestRunner config, or it will make code complicated. Because we need to reserve the `_default_options` (which is enriched) for parser.

raoyitao:
> > > Could not we keep the legacy way of working if no module is specified then the target is considered as that all but last component is the module as currently, but mark that path as deprecated for a while, then deco it in a month time or so?
> > 
> > 
> > Oh yes, although there's some ambiguity but some user might wrote code that way, let me update it to support both way.
> 
> we can make that deprecated so eventually the ambiguity removed, just give some heads-up

Yeah I add a `warnings.warn` to print the message.

kn-ms (approved):

PR #681, opened 2021-08-02
tpsreport merge

kn-ms:
simple merge functionality for tpsreport

## Solution description
new merge command, with a simple combine logic
new target to display in webui 
new display command as a shortcut to convert ... display
documentation for all tpsreport commands

## Checklist:
- [ ] Test
- [X] Documentation (API)
- [X] MS info leakage check


Pyifan (dismissed):
I think overall this looks very neat and extensible. The merge logic looks a bit simple - there might be quite a few edge cases we want to handle, like name dup of attachements, how to populate start/stop time etc etc. Surely we can add them later when the use case is more clear.

markokitonjics (approved):

PR #698, opened 2021-08-30
Task discovery feature

Pyifan:
new @task_target decorator and plan.schedule_all method -- for auto discover and schedule to tasks.

## Solution description
Note that a task target can be instantiated into multiple task objects according to user specified parameters (similar to testcase parametrization).

## Checklist:
- [X] Test
- [X] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/runners/pools/tasks/base.py
+++ b/testplan/runners/pools/tasks/base.py
@@ -1,15 +1,15 @@
 """Tasks and task results base module."""
 
-import sys
 import inspect
+import threading
 import warnings
-import importlib
 from collections import OrderedDict
 
 import pickle
 import copy
 
 from testplan.common.utils import strings
+from testplan.common.utils.package import import_tmp_module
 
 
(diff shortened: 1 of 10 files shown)


testplan/runners/pools/tasks/base.py
```diff
@@ -231,43 +231,44 @@ def materialize(self, target=None):
...
+            else:
+                module = self._module
+                target = self._target
+
+            if module in sys.modules:
+                mod = sys.modules[module]
+                module_imported = False
+            else:
+                mod = importlib.import_module(module)
+                module_imported = True
```
kn-ms, line 255: This looks to be a pattern used elsewhere does it worth a utility function?

Maybe a context manager that auto unload the module when imported from here?

Pyifan: good suggestion, i have revised this part.

yuxuan-ms (approved):

PR #702, opened 2021-09-15
Fix an issue that DictMatch items cannot be filtered on WebUI

raoyitao:
* Filter options 'Failures only' and 'Hide ignored items' does not
  work for nested items, because only the top-level item has the
  result marked as "failed" or "ignored", should apply this flag
  to descendants. `FixMatch` also has this issue like `DictMatch`.
* `re.Pattern` instance is marked as type "REGEX" after formatted.
* Add testcases.

### DIFF
--- a/testplan/common/utils/convert.py
+++ b/testplan/common/utils/convert.py
@@ -89,7 +89,7 @@ def full_status(status):
     return ""
 
 
-def expand_values(rows, level=0, ignore_key=False, key_path=None):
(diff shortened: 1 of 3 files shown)


Pyifan (changes requested):
Good finding and nice fix, Yitao! Shall we add a simple test for this change please? Thank you!

Pyifan (approved):

PR #703, opened 2021-09-15
Tree view navigation refinements

markokitonjics:
The tree view navigation needed some refinements.

## Solution description
Made the tree view navigation scrollable.
Desatured tags color.
Fixed padding.
Expand / collapse works only with icon click not label.
Reset color of the slashes in the entry icons.
Entry icon cut off workaround fix.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/web_ui/testing/src/Nav/NavEntry.js
+++ b/testplan/web_ui/testing/src/Nav/NavEntry.js
@@ -39,10 +39,7 @@ const NavEntry = (props) => {
       className='d-flex justify-content-between align-items-center'
       style={{
         height: "1.5em",
-        webkitUserSelect: "text",
-        MozUserSelect: "text",
-        MsUserSelect: "text",
-        UserSelect: "text",
+        userSelect: "text"
       }}
     >
       <Badge
--- a/testplan/web_ui/testing/src/Nav/NavEntry.js
+++ b/testplan/web_ui/testing/src/Nav/NavEntry.js
@@ -37,7 +37,13 @@ const NavEntry = (props) => {
   return (
     <div
       className='d-flex justify-content-between align-items-center'
-      style={{ height: "1.5em" }}
+      style={{
+        height: "1.5em",
+        webkitUserSelect: "text",
+        MozUserSelect: "text",
+        MsUserSelect: "text",
+        UserSelect: "text",
+      }}
     >
       <Badge
         className={css(styles.entryIcon, styles[badgeStyle], styles.badge)}
--- a/testplan/web_ui/testing/src/Nav/TagList.js
+++ b/testplan/web_ui/testing/src/Nav/TagList.js
@@ -1,7 +1,8 @@
-import React, {Component} from 'react';
+import React, { Component } from 'react';
 import PropTypes from 'prop-types';
-import {Badge} from 'reactstrap';
-import {StyleSheet, css} from 'aphrodite';
+import { Badge } from 'reactstrap';
+import { StyleSheet, css } from 'aphrodite';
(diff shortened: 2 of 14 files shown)


testplan/web_ui/testing/src/Nav/NavEntry.js
```diff
@@ -37,7 +37,13 @@ const NavEntry = (props) => {
   return (
     <div
       className='d-flex justify-content-between align-items-center'
-      style={{ height: "1.5em" }}
+      style={{
+        height: "1.5em",
+        webkitUserSelect: "text",
```
yuxuan-ms, line 42: https://developer.mozilla.org/en-US/docs/Web/CSS/user-select#browser_compatibility 
UserSelect is enough, all of the modern web browsers support user-select

testplan/web_ui/testing/src/Nav/TagList.js
```diff
@@ -53,6 +53,7 @@ TagList.propTypes = {
 const styles = StyleSheet.create({
   tags: {
     'margin-right': '.4em',
+    'background-color': '#6a95c3'
```
yuxuan-ms, line 56: move the color definition to https://github.com/morganstanley/testplan/blob/main/testplan/web_ui/testing/src/Common/Styles.js or 
https://github.com/morganstanley/testplan/blob/main/testplan/web_ui/testing/src/Common/defaults.js

testplan/web_ui/testing/src/Nav/TreeView.js
```diff
@@ -4,15 +4,37 @@ import TreeView from '@material-ui/lab/TreeView';
...
+const createTreeViewStyles = makeStyles({
+  treeItemLabel: {
+    padding: '5px 0px',
+    '&:hover': {
+      backgroundColor: MEDIUM_GREY
+    },
+    // WORKAROUND:
+    // if display: flex (default) then entry icons get cut off in case a long test case name
+    // if display: block then expand / collapse icons get rendered in a seperate row
+    width: '90%',
```
yuxuan-ms, line 27: you just need to set "overflow: hidden;" in .MuiTreeItem-label

markokitonjics: You are a genius! 😀

yuxuan-ms (dismissed):

testplan/web_ui/testing/src/Nav/TreeView.js
```diff
@@ -67,63 +74,90 @@ TreeViewNav.propTypes = {
...
       selection: selectionuids
     });
   const tags = (
-    (props.displayTags && entry.tags)
-      ? <TagList entryName={entry.name} tags={entry.tags} />
+    (props.displayTags && props.entry.tags)
+      ? <TagList entryName={props.entry.name} tags={props.entry.tags} />
       : null
   );
+  const treeViewClasses = createTreeViewStyles();
```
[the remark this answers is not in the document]
markokitonjics: makeStyles returns a hook. Hooks can only be called inside of the body of a function component. There are other ways to override styles / classes but this one is the simplest and cleanest way imo.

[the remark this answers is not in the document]
kn-ms: my bad, I misread the code, so createTreeViewStyle is not a function but the result of makeStyles which is a hook. So far so good, but then I do not like the name of it. I more like something like getTreeViewStyles so it suggest you getting from somewhere and not recreating every time.

kn-ms (approved):

Pyifan (approved):

PR #705, opened 2021-09-17
Time data should be recorded in report

raoyitao:
* Now only time information of running tests for Multitest is recorded
  but time of setup/teardown (start/stop resources) is missing. For
  other type of tests (ProcessRunnerTest, PyUnit, PyTest) all of the
  time information is not recorded. We should add it.
* Add a new testcase.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/tests/unit/testplan/testing/test_base.py
+++ b/tests/unit/testplan/testing/test_base.py
@@ -1,0 +1,117 @@
+"""TODO."""
+
+import time
+
+from testplan import TestplanMock
+from testplan.common.entity import Resource, Runnable
+from testplan.testing.base import Test, TestResult
+from testplan.runnable import TestRunnerStatus
+from testplan.runners.local import LocalRunner
+from testplan.report import Status, TestGroupReport, ReportCategories
+
+
+class DummyDriver(Resource):
+    def start(self):
+        self.status.change(self.STATUS.STARTING)
+        self.starting()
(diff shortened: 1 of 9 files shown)


tests/unit/testplan/testing/test_base.py
```diff
@@ -0,0 +1,108 @@
...
+        self.resources.add(DummyDriver(), uid="drv1")
+        self.resources.add(DummyDriver(), uid="drv2")
+
+    def run_tests(self):
+        with self._result.report.timer.record("run"):
+            time.sleep(0.5)  # 500ms for execution
+            self._result.report.status_override = Status.PASSED
+
+    def pre_resource_steps(self):
+        self._add_step(lambda: self.result.report.timer.start("setup"))
```
Pyifan, line 47: I think this logic is already in Test class, so DummyTest shall just use the inherited pre_resource_steps and we can test that time is noted.

Pyifan (changes requested):

tests/unit/testplan/testing/test_base.py
raoyitao: updated. remove "setup" and "teardown" field and add other flags for testing.

Pyifan (approved):

PR #718, opened 2021-10-13
result.attach support path of directory

raoyitao:
* `attach` assertion automatically check the path, if it's a directory
  then retrieve files under it (support arguments like `ignore`, `only`
  and `recursive` as filter)
* Write file paths in stdout and generated pdf for that new `Directory`
  assertion entry and all `Attachment` based assertion entries.
* Add new testcases and update examples.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/common/utils/path.py
```diff
@@ -282,38 +283,73 @@ def archive(path, timestamp):
...
+                    dpath = os.path.join(dirpath, dname)
+                    _, _, relpath = dpath.partition(directory)
+                    while relpath.startswith(os.sep):
+                        relpath = relpath[len(os.sep) :]
+                    result.append(relpath)
+
+        for fname in filenames:
+            if not should_ignore(fname):
+                fpath = os.path.join(dirpath, fname)
+                _, _, relpath = fpath.partition(directory)
```
Pyifan, line 347:
```suggestion
                relpath = os.path.relpath(os.path.join(dirpath, fname), directory)
```

testplan/common/utils/path.py
```diff
@@ -282,38 +283,73 @@ def archive(path, timestamp):
...
 
-    :param directory: Path to a directory that will be traversed
+    :param directory: Path to a directory that will be traversed.
     :type directory: ``str``
-    :param topdown: Browse the directory in a top-down approach.
+    :param topdown: Browse the directory in a top-down or bottom-up approach.
     :type topdown: ``bool``
+    :param ignore: list of patterns to ignore  by glob style filtering.
+    :type ignore: ``list``
+    :param only: list of patterns to exclusively consider by glob style
```
Pyifan, line 304:
```suggestion
    :param only: list of patterns to include by glob style
```

testplan/common/utils/path.py
```diff
@@ -282,38 +283,73 @@ def archive(path, timestamp):
...
+        filtering.
+    :type only: ``list``
+    :param recursive: Traverse through all sub directories recursively.
```
Pyifan, line 307:
```suggestion
    :param recursive: Traverse directories recursively; set to False to only list top directory.
```

Pyifan (changes requested):

Pyifan (approved):

PR #723, opened 2021-10-27
Add new button to kill Testplan in interactive mode

raoyitao:
* A button for trigger an event that Testplan should quit at backend
  and client UI will be informed before Testplan dies.
* Make test runner and interactive handler abort execution gracefully.
* Update testcases.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/runnable/interactive/http.py
```diff
@@ -906,3 +915,11 @@ def run(self):
             self._server.serve()
         finally:
             self._server = None
+
+    def stop(self):
```
Pyifan, line 919: Shall we call it teardown instead?

testplan/common/entity/base.py
```diff
@@ -666,6 +666,7 @@ class Runnable(Entity):
     STATUS = RunnableStatus
     RESULT = RunnableResult
     ENVIRONMENT = Environment
+    _STOP_TIMEOUT = 5
```
Pyifan, line 669: This timeout here is quite arbitary. It might be easier to just set _ihandler thread as daemon.

testplan/runnable/interactive/base.py
```diff
@@ -139,17 +140,13 @@ def target(self):
         """
         return self._cfg.target
 
-    def abort_dependencies(self):
-        """
-        We abort our test runner instance.
-        """
-        yield self.target
-
     def aborting(self):
```
Pyifan, line 143: Move self._http_handler.stop() to teardown and call teardown from aborting?

testplan/runnable/interactive/base.py
```diff
@@ -123,7 +123,8 @@ def teardown(self):
 
         self._pool = None
         self._http_handler = None
-        self.target._close_file_logger()
+        if self.target.active:
```
Pyifan, line 126: When teardown is called, active might already be False.

Pyifan (changes requested):

testplan/runnable/interactive/http.py
raoyitao: Actually the http handler is created by interactive handler and set it as parent, so I can change it to `aborting`, in `abort_dependencies` of interactive handler, yield that http handler.

testplan/common/entity/base.py
raoyitao: Yes I will remove it, I found there is an cfg option `abort_wait_timeout` available for this. Also I set `thread.daemon` to True for safety reason. But I still keep `interruptible_join` there to allow interactive handler quit gracefully.

testplan/runnable/interactive/base.py
raoyitao: As mentioned above this function is removed but add `abort_dependencies`

testplan/runnable/interactive/base.py
raoyitao: I removed this condition, because now the test runner wait for the interactive handler thread to quit, so at this time the target (test runner) must be active.

testplan/common/entity/base.py
```diff
@@ -358,8 +358,7 @@ def get_options(cls):
             ConfigOption("initial_context", default={}): dict,
             ConfigOption("path_cleanup", default=False): bool,
             ConfigOption("status_wait_timeout", default=600): int,
-            ConfigOption("abort_wait_timeout", default=30): int,
-            # active_loop_sleep impacts cpu usage in interactive mode
+            ConfigOption("abort_wait_timeout", default=10): int,
```
Pyifan, line 361: revert

Pyifan (dismissed):

Pyifan (approved):

PR #729, opened 2021-11-16
Interactive reloading should support `plan.schedule`

raoyitao:
Interactive reloading should support `plan.schedule`

* Interactive reload has many issues, to be on the safe side, it is
  suggested to directly import packages/modules in the main script.
  This limitation should be removed and any import syntax should be
  freely used, except for `from mod import *` which is not encouraged.
* Multitest instances added by `plan.add` can be properly reloaded
  interactively, in the way of `plan.schedule` and `plan.schedule_all`
  all the instantiated objects should also be able to be reloaded.
* Add functional testcases for each corner case.
* Fix unit test of reloader.
* Update document.

## Checklist:
- [x] Test
- [x] Documentation (API)
- [ ] MS info leakage check


Pyifan (changes requested):

testplan/runnable/interactive/reloader.py
```diff
@@ -356,10 +404,23 @@ def import_hook(self, name, caller=None, *args, **kwargs):
...
-        self._curr_caller = caller
-        return modulefinder.ModuleFinder.import_hook(
+        previous_caller = self._curr_caller
+
+        if caller is None:
+            caller_frame = inspect.stack()[2]
+            self._curr_caller = caller_frame.frame.f_locals.get("m")
+            assert (
+                self._curr_caller is not None
+            ), "Source code of `modulefinder` library has changed !!"
```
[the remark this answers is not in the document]
raoyitao: As I known that the library does not change all the time while python version is changing, I'm afraid that one day the library is totally updated, then we can quickly know from this assertion.

doc/en/interactive.rst
```diff
@@ -33,37 +33,51 @@ interrupt processes started as part of the test environment.
...
-reload API and only applies to modules that have either been modified or have
-had its dependencies reloaded.
-
-Please note the ``__main__`` module (i.e. test_plan.py) cannot be reloaded.
-Because of this, it is recommended that test suites are defined in seperate modules.
-Another known limitation is that we are not able to reload module that is under
-namespace package, and the workaround is to add dummy __init__.py file.
+Code-reloading is triggered by reload API request, and there is a reload button
+on interactive mode UI. By default, testplan will look under the directory of
+the main script for modules to reload - those who have either been modified or
```
Pyifan, line 45: This is not precise, we will look under the directory of the main script and the directories where the tasks are scheduled from.

doc/en/interactive.rst
```diff
@@ -33,37 +33,51 @@ interrupt processes started as part of the test environment.
...
+* Testplan is not able to reload module that is under *Namespace Package*, the
+  workaround is to add dummy __init__.py file.
+* Unlike testcases that can be added/removed/modified, adding/removing/renaming
+  testsuite class will not take effect after reloading.
+* Modules or objects imported using ``from some_module import *`` syntax are
+  not guaranteed to be properly reloaded. Such syntax is not encouraged in
+  Python anyway as it may bring naming conflicts.
+* In interactive mode, all modules containing test targets will be loaded at
+  the same time, so pay attention that avoid scheduling tasks having duplicate
+  module name but from different paths.
```
Pyifan, line 79: If the modules from which tasks are scheduled have the same name, they won't reload properly.

doc/newsfragments/reload_schedule_task_target.change.rst
```diff
@@ -0,0 +1 @@
+* Remove interactive reload limitation: the scheduled task target should be supported.
```
Pyifan, line 1: Improvement for interactive mode reload feature (link to how reload works): can now pick up changes to the test added by plan.schedule().

Pyifan (approved):

Pyifan (approved):

PR #734, opened 2021-11-24
Add grouped time info to parametrized testcases

nober97:
The time info was missing for parametrized testcases in the closed group view. Since the info was missing from the report as well, the solution had to be implemented on the backend side.

## Solution description
The problem was that the timer attribute of the group report was only initialized but they were not started and stopped. So, this implementation provides that for each parametrized group report, its timer is set to the bounding time box that is fits in, meaning that the elapsed time is going to be defined as the difference between earliest start time and the latest end time of the testcases included in that parametrization group.

## Checklist:
- [x] Test
- [x] MS info leakage check


testplan/testing/multitest/base.py
```diff
@@ -833,8 +850,25 @@ def _run_parallel_testcases(self, testsuite, execution_groups):
 
         # Add all non-empty parametrization reports into the list of returned
         # testcase reports, to be added to the suite report.
+        # Calculate runtime of the parametrized group as well
         for param_report in parametrization_reports.values():
             if param_report.entries:
+                start_times = []
```
kn-ms, line 856: Here you repeat yourself, mostly, why are the checks more strict here than in the linear case?

testplan/testing/multitest/base.py
```diff
@@ -773,6 +773,23 @@ def _run_serial_testcases(self, testsuite, testcases):
                     )
                     break
 
+        if parametrization_reports:
+            for param_report in parametrization_reports.values():
+                if param_report.entries:
+                    start_times = []
+                    end_times = []
+                    for testcase in param_report.entries:
+                        timer = getattr(testcase, "timer")
```
kn-ms, line 782: why not just testcase.timer

testplan/testing/multitest/base.py
```diff
@@ -773,6 +773,23 @@ def _run_serial_testcases(self, testsuite, testcases):
...
+                        start_time = getattr(timer["run"], "start")
```
kn-ms, line 783: here you can use timer["run"].start and timer["run"]. end as they are just Interval types which is essentially a named tuple

Pyifan (dismissed):
Nice and exquisite!

testplan/testing/multitest/base.py
```diff
@@ -1154,3 +1161,24 @@ def _split_by_exec_group(testcases):
...
+    :param param_report: parametrized group report
+    :return: the parametrized group report with its runtime information
+    """
+    group_start_time = param_report.entries[0].timer["run"].start
+    group_end_time = param_report.entries[0].timer["run"].end
+    for testcase in param_report.entries:
+        timer = testcase.timer
+        start_time = timer["run"].start
+        end_time = timer["run"].end
+        if start_time < group_start_time:
```
kn-ms, line 1178: how about something like this?

```suggestion
        group_start_time =  start_time if group_start_time is None else min(group_start_time, start_time)
```

kn-ms (approved):

PR #737, opened 2021-12-01
launch web browser if browse argument has been passed on command line

tjisanaMS:

## Solution description

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/runnable/base.py
```diff
@@ -945,7 +945,8 @@ def _post_exporters(self):
                 report_urls.append(exporter.report_url)
             if getattr(exporter, "_web_server_thread", None):
                 # Give priority to open report from local server
-                webbrowser.open(exporter.report_url)
+                if self.cfg.browse:
+                    webbrowser.open(exporter.report_url)
                 report_opened = True
```
kn-ms, line 950: I know with the current implementation it does not make a difference but `report_opened = True` should only be set when browser opened.

tjisanaMS: Done

Pyifan (approved):

PR #738, opened 2021-12-02
Make possible to style assertion description on web UI

raoyitao:
* User wants to color the assertions by their own on web UI, thus
  a common argument for all assertion entries is provided that user
  can specify CSS style in a dictionary.
* Add a new testcase for this feature.
* Update document with a new example.
* Fix testcase `test_assertion_order` which has never run.
* Simplify the code of deserializing various reports.

## Checklist:
- [x] Test
- [x] Documentation (API)
- [ ] MS info leakage check


tests/unit/testplan/testing/multitest/test_result.py
```diff
@@ -58,14 +60,55 @@ def test_assertion_orders():
...
     # pylint: disable=invalid-sequence-index
-    assertions = (
+    assertions = [
         entry
         for entry in mtest.report.flatten()
         if isinstance(entry, dict) and entry["meta_type"] == "assertion"
+    ]
+
+    for idx, desc in enumerate(expected):
+        assert desc == assertions[idx]["description"]
```
Pyifan, line 70: how are these 2 lines different from original?

testplan/testing/multitest/entries/schemas/base.py
```diff
@@ -45,6 +46,7 @@ class GroupSchema(Schema):
     meta_type = fields.String()
     description = custom_fields.Unicode(allow_none=True)
     entries = GenericEntryList(allow_none=True)
+    # assertion_style = fields.Dict(keys=fields.String(), values=fields.String())
```
Pyifan, line 49: Remove commented code

testplan/testing/multitest/base.py
```diff
@@ -288,20 +288,26 @@ def get_test_context(self):
         :rtype: ``list`` of ``tuple``
         """
         ctx = []
-        sorted_suites = self.cfg.test_sorter.sorted_testsuites(self.cfg.suites)
+        sorted_suites = (
+            self.cfg.test_sorter.sorted_testsuites(self.cfg.suites)
+            if hasattr(self.cfg, "test_sorter")
```
Pyifan, line 293: Can multitest have a default test_sorter so that we don't need to check this?

testplan/testing/multitest/base.py
```diff
@@ -288,20 +288,26 @@ def get_test_context(self):
...
                 if getattr(suite, "strict_order", False)
+                or not hasattr(self.cfg, "test_sorter")
                 else self.cfg.test_sorter.sorted_testcases(suite, testcases)
             )
 
             testcases_to_run = [
                 case
                 for case in sorted_testcases
-                if self.cfg.test_filter.filter(
+                if not hasattr(self.cfg, "test_filter")
```
Pyifan, line 309: same as above

testplan/testing/base.py
```diff
@@ -195,7 +195,11 @@ def report(self):
     @property
     def stdout_style(self):
         """Stdout style input."""
-        return self.cfg.stdout_style
+        return (
+            self.cfg.stdout_style
+            if hasattr(self.cfg, "stdout_style")
```
Pyifan, line 200: can stdout_style have a default value, so that we don't need to do hasattr check?

testplan/report/testing/schemas.py
```diff
@@ -297,14 +288,12 @@ class Meta:
 
     @post_load
     def make_test_report(self, data, **kwargs):
-        status_override = data.pop("status_override", None)
```
Pyifan, line 300: why are we removing all these pop lines?

Pyifan (changes requested):

Pyifan:
Let also add example, documentation and newsfragment for this PR.

tests/unit/testplan/testing/multitest/test_result.py
raoyitao: the testcase raises an exception, so in the original for loop, the `actual` is enumerated but it is empty, so no assert statement is really executed. now the bug is fixed and we just enumerate the `expected`.

testplan/report/testing/schemas.py
raoyitao: I guess that it is originally needed, but as code evolves (the constructor of those Report classes changes), the code of deserialization does not follow. So I refine the code and remove the redundant.

testplan/testing/multitest/base.py
raoyitao: We cannot have a default `test_sorter` (as well as `test_filter`, `stdout_style`). This is an old issue. Because in the past we have another arguments to decide which cfg option is preferred if it is defined in both parent and child instance, but that argument is removed. So, now `TestConfig` class does not have default `test_sorter`, `test_filter` and `stdout_style` because these cfg options should be retrieved from `TestRunner`. This is reasonable or the cfg option we defined in @test_plan cannot take effect. But a small problem is that in the unit test, a single `MultiTest` instance cannot run and every time we want to write a test we have to pass a lot of arguments to it (that is why `test_assertion_order` always raise an exception). So with these `hasattr` check things become simple.

testplan/testing/multitest/entries/schemas/base.py
raoyitao: will remove it.

testplan/testing/multitest/base.py
Pyifan: Ok, then i think it would make better sense to add the multitest to a mockplan in test code -- we are doing this in many places.

raoyitao:
> 

Document added

doc/newsfragments/custom_style_of_assertion_header.feature.rst
```diff
@@ -0,0 +1 @@
+* Support custom style (e.g. background color) of assertion header on web UI.
```
Pyifan, line 1: Announcement need to link to example / doc, so that user know how to use this feature.

Pyifan (changes requested):

Pyifan:
> 

We have a bunch of examples for different assertion. I think it will be good to have one example dedicated for assertion styling, demonstrate changed text font/size/color/background, it will be easier for user to copy-paste.

doc/en/assertions.rst
```diff
@@ -1989,3 +1990,33 @@ These utilities are simple, composable and callable objects and produce more rea
 
       >>> str(custom_comparator)
       (<value> > 2 and <value> < 5)
+
+
+Custom Style of Assertion Header on Web UI
+==========================================
+The generated test report can be viewed in a browser (:ref:`link <Output_Browser>`), including details of assertions.
```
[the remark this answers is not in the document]
raoyitao: fixed

doc/en/assertions.rst
```diff
@@ -1989,3 +1990,33 @@ These utilities are simple, composable and callable objects and produce more rea
 
       >>> str(custom_comparator)
       (<value> > 2 and <value> < 5)
+
+
+Custom Style of Assertion Header on Web UI
```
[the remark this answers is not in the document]
raoyitao: fixed

doc/en/assertions.rst
```diff
@@ -1989,3 +1990,33 @@ These utilities are simple, composable and callable objects and produce more rea
...
+              description='Equality example',
+              custom_style={'background-color': '#FBFBAA'},
+          )
+          result.log(
+              ['foo', 'bar', 'baz'],
+              description='Log of raw data',
+              custom_style={'background-color': '#CFBBCF'},
+          )
+
+As a result the web UI should look like this:
```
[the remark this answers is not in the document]
raoyitao: fixed

doc/newsfragments/custom_style_of_assertion_header.feature.rst
raoyitao: added some new marks.

* Support custom style (e.g. background color) of assertion header on web UI, refer to :ref:`document <Styling_Assertions>`

raoyitao:
> 

examples added

Pyifan (approved):

Pyifan (approved):

Pyifan (approved):

PR #739, opened 2021-12-09
Allow `schedule_all` to pass arguments to `task_target`

raoyitao:
* When many test targets of the same kind use the same arguments to
  initialize `Task` objects, it is useful to set `args` and `kwargs`
  in `schedule_all` method.
* Allow the decorator `task_target` used without parentheses following.

## Checklist:
- [ ] Test
- [x] Documentation (API)
- [ ] MS info leakage check


doc/en/pools.rst
```diff
@@ -331,45 +331,72 @@ in modules that matches the name_pattern under current working directory.
 
 .. code-block:: python
 
-    @task_target()
-    def make_multitest1():
-        # a test target shall only return 1 runnable object
```
Pyifan, line 336: This line literally means "a test target SHALL only return 1 runnable object", as it only return 1 Multitest object.

doc/en/pools.rst
```diff
@@ -331,45 +331,72 @@ in modules that matches the name_pattern under current working directory.
...
-        test = MultiTest(name="Proj2-Suite1", suites=[Suite1()])
+    @task_target
+    def make_multitest():
+        # a test target only returns 1 runnable object
+        test = MultiTest(name="MTest", suites=[Suite()])
         return test
 
 Once found, task object will be created from the target, and scheduled to pool.
-It is also possible to create multiple task objects out of one target:
```
Pyifan, line 341: Why remove this line of doc?

doc/en/pools.rst
```diff
@@ -331,45 +331,72 @@ in modules that matches the name_pattern under current working directory.
...
+Moreover, arguments for tasks can be specified in `schedule_all`, it is useful
+when many targets of the same kind share the same arguments.
+
+.. code-block:: python
+
+    plan.schedule_all(
+        path=".",
+        name_pattern=r".*tasks\.py$",
+        args=(2,),
+        kwargs={"stop_on_error": False},
```
Pyifan, line 350: This differs from plan.schedule() interface, let's just add **options to schedule_all().

doc/en/pools.rst
```diff
@@ -331,45 +331,72 @@ in modules that matches the name_pattern under current working directory.
```
Pyifan, line 331: Look above, schedule_all() is introduced in the beginning. If you want to introduce the **options param of schedule_all, it goes there. Do more mix schedule_all with make_multitest()

doc/en/pools.rst
```diff
@@ -331,45 +331,72 @@ in modules that matches the name_pattern under current working directory.
...
         return test
 
 Once found, task object will be created from the target, and scheduled to pool.
-It is also possible to create multiple task objects out of one target:
+Moreover, arguments for tasks can be specified in `schedule_all`, it is useful
+when many targets of the same kind share the same arguments.
+
+.. code-block:: python
+
+    plan.schedule_all(
```
Pyifan, line 346: Here, mixing make_multitest() - schedule_all() - make_multitest(), makes the documentation confusing.

doc/en/pools.rst
```diff
@@ -331,45 +331,72 @@ in modules that matches the name_pattern under current working directory.
...
-                suites=[sub_proj1.suites.Suite1],
+                suites=[ComplicatedSuite],
             ),
         ),
         # additional args of Task class
         rerun=1,
         weight=1,
     )
     def make_multitest(name, part_tuple=None, suites=None):
-        # a test target shall only return 1 runnable object
```
Pyifan, line 366: This also literally mean the make_multitest function returns 1 multitest object.

Pyifan (changes requested):

doc/en/pools.rst
raoyitao: this annotation is really confusing, apparently in function `make_multitest` there is only one object returned. I think the annotation here (or better above the function definition) should tell user, because this function is decorated with argument `parameters`, then finally 3 multitest objects can be retieved.

doc/en/pools.rst
raoyitao: put it back.

doc/en/pools.rst
raoyitao: add it back

doc/en/pools.rst
raoyitao: **options are used to initializing a Task object, while arguments of `schedule_all`  are for general purpose.

doc/en/pools.rst
raoyitao: I'm not very understanding it .

doc/en/pools.rst
raoyitao: then if task object (like `make_multitest`) needs arguments specified by user (not hard coded by `parameters`), how could we pass it ?

PR #740, opened 2021-12-09
Make the url in UI info box clickable

tjisanaMS:

## Solution description

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/web_ui/testing/src/Toolbar/Toolbar.js
```diff
@@ -503,10 +504,24 @@ const getToggledButtonStyle = (status) => {
...
   if (!report || !report.information) {
     return "No information to display.";
   }
   const infoList = report.information.map((item, i) => {
+    const linkifyIgnore = ['user', 'command_line_string', 'python_version', 'hostname', 'start', 'end']
+    if (!linkifyIgnore.includes(item[0])) {
+      return (
+        <tr key={i}>
+          <td className={css(styles.infoTableKey)}>{item[0]}</td>
+          <td className={css(styles.infoTableValue)}><div dangerouslySetInnerHTML={{__html:linkifyUrls(
```
kn-ms, line 518: There is too many duplication here with the ignored case.

I see two option 

1. create a function that take an item, and do the check and ither return the div with linkifyUrls or item[0]
2. use the ternary operator inline

testplan/web_ui/testing/src/Toolbar/Toolbar.js
```diff
@@ -503,14 +505,37 @@ const getToggledButtonStyle = (status) => {
...
+
   if (!report || !report.information) {
     return "No information to display.";
   }
   const infoList = report.information.map((item, i) => {
+    const linkifyIgnore = ['user', 'command_line_string',
+    'python_version', 'hostname', 'start', 'end'];
+
+
+    let columnData = !linkifyIgnore.includes(item[0]) ? < td className = {
```
kn-ms, line 518: probably an exact match would be better instead of includes

kn-ms: The ternary operator is hard to read here, and you repeating the same <td>.

a couple of options you can put the content goes into the <td> into a variable (here you can use the ternary operator )which is either the <div> or just the string content. Alternatively the working out of the linkify thing can be moved to it's own method together with the list of ignored fields, and just use that method in the template.

testplan/web_ui/testing/src/Toolbar/Toolbar.js
tjisanaMS: Done

testplan/web_ui/testing/src/Toolbar/Toolbar.js
tjisanaMS: Done

kn-ms (approved):

Pyifan (approved):

PR #742, opened 2021-12-14
`task_target` can be used without parentheses

raoyitao:
* Allow the decorator `task_target` used without parentheses following.
* When `schedule_all` is used to discover task targets, once a target
  is found then should be added to executor, rather adding all targets
  after directory traversed, so the log of 'Discovered task target' and
  'Task created' can match for easy tracing, also the exception can be
  raised in advance if there's Test instances with duplicate name.

## Checklist:
- [ ] Test
- [x] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/doc/en/pools.rst
+++ b/doc/en/pools.rst
@@ -327,49 +327,50 @@ to do task discovery.
     )
 
 In the code above, testplan will go look for @task_target decorated functions
-in modules that matches the name_pattern under current working directory.
+in modules that matches the ``name_pattern`` under current working directory.
 
 .. code-block:: python
 
-    @task_target()
-    def make_multitest1():
-        # a test target shall only return 1 runnable object
-        test = MultiTest(name="Proj2-Suite1", suites=[Suite1()])
+    @task_target
+    def make_multitest():
+        # A test target shall only return 1 runnable object
+        test = MultiTest(name="MTest", suites=[Suite()])
         return test
 
 Once found, task object will be created from the target, and scheduled to pool.
-It is also possible to create multiple task objects out of one target:
+It is possible to create multiple task objects out of one target with
+`parameters` specified:
 
 .. code-block:: python
 
     @task_target(
         parameters=(
             # positional args to be passed to target, as a tuple or list
-            ("Proj1-Suite2", None, [sub_proj1.suites.Suite2]),
+            ("MTest1", None, [SimpleSuite1, SimpleSuite2]),
             # keyword args to be passed to target, as a dict
             dict(
-                name="Proj1-Suite1",
+                name="MTest2-1",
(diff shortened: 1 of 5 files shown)


testplan/runnable/base.py
```diff
@@ -483,41 +487,32 @@ def schedule_all(self, path=".", name_pattern=r".*\.py$", resource=None):
...
-                                args = None
-                                kwargs = param
-                            elif isinstance(param, (tuple, list)):
-                                args = param
-                                kwargs = None
-                            else:
-                                raise TypeError(
-                                    "task_target's parameters can only contain"
-                                    f" dict/tuple/list, but received: {param}"
-                                )
```
Pyifan, line 504: I like this old exception msg better, which is more helpful for user.

testplan/runnable/base.py
```diff
@@ -462,19 +462,23 @@ def schedule_all(self, path=".", name_pattern=r".*\.py$", resource=None):
         :type resource: :py:class:`~testplan.runners.pools.base.Pool`
         """
 
+        def create_task(task_kwargs, resource):
```
Pyifan, line 465: Shall we name this function schedule_task instead of create_task?

doc/en/pools.rst
```diff
@@ -327,49 +327,59 @@ to do task discovery.
...
         return test
 
+.. code-block:: python
+
+    plan.schedule_all(
+        path=".",
+        name_pattern=r".*tasks\.py$",
+        resource="MyPool",
+    )
+
```
Pyifan, line 347: This block repeats line 323, pls remove.

doc/en/pools.rst
```diff
@@ -327,49 +327,59 @@ to do task discovery.
...
             dict(
-                name="Proj1-Suite1",
+                name="MTest2-2",
                 part_tuple=(1, 2),
-                suites=[sub_proj1.suites.Suite1],
+                suites=[ComplicatedSuite],
             ),
         ),
-        # additional args of Task class
+        # additional argumentss of Task class
```
Pyifan, line 370: typo

doc/en/pools.rst
```diff
@@ -327,49 +327,59 @@ to do task discovery.
...
         rerun=1,
         weight=1,
     )
     def make_multitest(name, part_tuple=None, suites=None):
-        # a test target shall only return 1 runnable object
+        # By parameterizing this task target totally 3 runnable objects
+        # can be retrieved during discovering task targets.
```
Pyifan, line 376: Please restore the original doc string.

Pyifan (changes requested):
Please also add new fragment.

Pyifan (approved):

PR #744, opened 2021-12-22
Enable sorting and filtering for PyTest/PyUnit

raoyitao:
* Update sorting and filter module so that they can deal with basic
  string type as well as testsuite/testcase instances.
* Test sorter can be applied to PyTest and PyUnit.
* Test filter can be applied to PyTest and PyUnit (but PyUnit cannot
  support testcase-level filtering).
* Remove a fixture that is no longer used.
* Add and update testcases.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/testing/filtering.py
```diff
@@ -291,13 +291,25 @@ def parse_pattern(self, pattern):
...
+            return True
+        elif isinstance(test, str):
+            return fnmatch.fnmatch(test, self.test_pattern)
         return fnmatch.fnmatch(test.name, self.test_pattern)
 
     def filter_suite(self, suite):
-        # For test suite uid is the same as name, just like that of Multitest
+        if suite is None:
+            return True
+        elif isinstance(suite, str):
```
Pyifan, line 303: I think a common problem I see is this if-else branch based on parameter type.

Pyifan:
For 3rd party testing frameworks, I suggest that we leave filtering to the framework itself. All testplan needs to do is to construct the pattern string and pass it to frameworks' arg (if supported). I'm not a big fan of we first extract the context and then apply our own filter logic.

Pyifan (changes requested):

PR #745, opened 2021-12-22
Allow ProcessRunnerTest based tests to take extra arguments

nober97:
Some tests can take extra user parameters. However, they can't be added to the executable. The goal is to let ProcessRunnerTest based tests to pass user arguments to the executable in list and in execute mode as well.

## Solution description
ProcessRunnerTest's config was extended with two parameters (pre_args and post_args) that meant to be prepended and appended to the list of args to the test runnable in list_command and test_command methods. The original methods were made to private members of the class.

## Checklist:
- [x] Test
- [x] MS info leakage check

### DIFF
--- a/testplan/testing/base.py
+++ b/testplan/testing/base.py
@@ -474,6 +474,12 @@ class ProcessRunnerTest(Test):
                     This can be disabled by providing a list of
                     numbers to ignore.
     :type ignore_exit_codes: ``list`` of ``int``
+    :param pre_args: List of arguments to be prepended before the
+        arguments of the test runnable.
+    :type pre_args: ``list``
+    :param post_args: List of arguments to be appended before the
+        arguments of the test runnable.
+    :type post_args: ``list``
 
     Also inherits all
     :py:class:`~testplan.testing.base.Test` options.
--- a/testplan/testing/base.py
+++ b/testplan/testing/base.py
@@ -476,10 +476,10 @@ class ProcessRunnerTest(Test):
     :type ignore_exit_codes: ``list`` of ``int``
     :param pre_args: List of arguments to be prepended before the
         arguments of the test runnable.
-    :type pre_args: ``list``
+    :type pre_args: ``list`` of ``str``
     :param post_args: List of arguments to be appended before the
         arguments of the test runnable.
-    :type post_args: ``list``
+    :type post_args: ``list`` of ``str``
 
     Also inherits all
     :py:class:`~testplan.testing.base.Test` options.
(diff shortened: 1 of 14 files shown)


testplan/testing/base.py
```diff
@@ -488,11 +490,19 @@ class ProcessRunnerTest(Test):
...
             proc_env.update(options["proc_env"])
         options["proc_env"] = proc_env
+        if pre_args:
+            options["pre_args"] = pre_args
+        else:
+            options["pre_args"] = []
+        if post_args:
+            options["post_args"] = post_args
+        else:
+            options["post_args"] = []
```
Pyifan, line 505: If we remove pre_args & post_args from the parameter list, then they gets to **options, and will be updated to self.cfg.pre_args/post_args automatically when Entity class's __init__ is  called. Then we don't need these few lines.

testplan/testing/base.py
```diff
@@ -529,7 +539,13 @@ def test_command(self):
...
         :rtype: ``list`` of ``str``
         """
-        return [self.cfg.binary]
+        if self.cfg.pre_args:
+            cmd = [*self.cfg.pre_args, self.cfg.binary]
+        else:
+            cmd = [self.cfg.binary]
+        if self.cfg.post_args:
+            cmd.extend(self.cfg.post_args)
+        return cmd
```
Pyifan, line 548: Similar logic is repeated in derived class, and I think we could avoid this. One possible way is to move the original logic to test_command to _test_command (and inheritances will override accordingly), and in base class test_command, add pre/post_args (and inheritances do not need to override).

Pyifan (changes requested):
We also need news fragment, and a test (or an example would also do)

testplan/testing/base.py
```diff
@@ -437,6 +437,8 @@ def get_options(cls):
                 None, float, int, Use(parse_duration)
             ),
             ConfigOption("ignore_exit_codes", default=[]): [int],
+            ConfigOption("pre_args", default=[]): list,
+            ConfigOption("post_args", default=[]): list,
```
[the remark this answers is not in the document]
nober97: Solved

testplan/testing/base.py
```diff
@@ -488,11 +490,19 @@ class ProcessRunnerTest(Test):
     _VERIFICATION_TESTCASE_NAME = "ExitCodeCheck"
     _MAX_RETAINED_LOG_SIZE = 4096
 
-    def __init__(self, **options):
+    def __init__(self, pre_args=None, post_args=None, **options):
```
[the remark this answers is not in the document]
nober97: Removed.

testplan/testing/base.py
nober97: Removed.

Pyifan (approved):

PR #747, opened 2021-12-22
Make empty result object not to be false in bool context

nober97:
The Result object can be true only if it contains at least one entry in it. However, a user wants the Result object to be true when it is a valid Result object, even without any entry.
## Solution description
The implementation of the built-in bool() method of the Result class provides this functionality.

## Checklist:
- [x] Test
- [x] MS info leakage check


doc/newsfragments/empty_result.feature.rst
```diff
@@ -0,0 +1 @@
+* Make empty result object not to be false in bool context
```
Pyifan, line 1: Let's add a period in the end.

Pyifan (changes requested):

doc/newsfragments/empty_result.feature.rst
nober97: Solved

Pyifan (approved):

PR #760, opened 2022-02-06
Add yarn.lock

M6AI:
* Generated yarn.lock file.
* Minor refactoring to install-testplan-ui.
* Removed yarn.lock from gitignore.


install-testplan-ui
```diff
@@ -49,8 +47,10 @@ def parse_cli_args():
     return parser.parse_args()
 
 
-def yarn_installed():
-    """Check if yarn has been installed."""
+def is_yarn_installed() -> bool:
```
M6AI, line 50: Function names are better phrased as verbs or questions.

testplan/testing/py_test.py
```diff
@@ -546,7 +546,7 @@ def pytest_exception_interact(self, node, call, report):
             )
             details = (
                 "File: {}\nLine: {}\n{}: {}".format(
-                    trace.path.strpath,
+                    str(trace.path),
```
M6AI, line 549: The latest version of pytest (7.0.0) now installed as dependency during the Github actions. The path attribute of the traceback entry so far has been treated as a LocalPath object which indeed has the strpath attribute; with the latest pytest this becomes a Pathlike object (for instance WindowsPath), that simply needs to be converted to str.

Suggest avoiding the usage of undocumented attributes, like strpath was in this case. Simply calling str() would have given the same value.

PR #761, opened 2022-02-08
Add npm package lock

M6AI:


install-testplan-ui
```diff
@@ -92,9 +96,9 @@ def install_and_build_ui(path=TESTPLAN_UI_DIR, dev=False, verbose=False):
             production_cmd = ""
 
         logging.info("Installing Testplan UI dependencies...")
-        logging.info("Installing to path: {}".format(os.path.abspath(path)))
+        logging.info(f"Installing to path: {os.path.abspath(path)}")
         subprocess.check_call(
-            "yarn install {} --network-timeout 600000".format(production_cmd),
+            f"npm install{production_cmd}",
```
M6AI, line 101: NPM does not support the --network-timeout option.

testplan/testing/py_test.py
```diff
@@ -546,7 +548,7 @@ def pytest_exception_interact(self, node, call, report):
             )
             details = (
                 "File: {}\nLine: {}\n{}: {}".format(
-                    trace.path.strpath,
+                    str(trace.path),
```
M6AI, line 551: Breaking change from pytest version 7.0.0 that LocalPath objects are now replaced with Pathlike. In any case, the strpath attribute was not part of the "public" pytest interface at any point so this is a desired change in my book even if fixing pytest version at less than 7.

testplan/web_ui/testing/package.json
```diff
@@ -43,7 +43,7 @@
     "react-scripts": "^3.4.1",
     "react-spinners": "^0.6.0",
     "react-syntax-highlighter": "^11.0.2",
-    "react-test-renderer": "16.6.0",
+    "react-test-renderer": "16.8.5",
```
M6AI, line 46: This update is needed to allow the shallow method of enzyme to satisfy the Rule of Hooks with useState hook.

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/CopyButton.js
```diff
@@ -1,7 +1,77 @@
...
+      children,
+      ...props
+    } = this.props;
+    const elem = React.Children.only(children);
+
+    return React.cloneElement(
+      elem, {...props, onClick: this.onClick, onTouchTap: this.onClick}
+    );
+  }
+}
```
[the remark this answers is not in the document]
Pyifan: It seems we will need to distribute the original project's license file as well, if we copy-paste their code.

Pyifan (changes requested):

testplan/web_ui/testing/src/AssertionPane/AssertionTypes/CopyButton.js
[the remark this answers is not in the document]
M6AI: Thanks Yifan, very good point. The package is distributed under the MIT licence https://github.com/ukrbublik/react-copy-html-to-clipboard/blob/master/LICENSE. You are more experienced in this: what is the set of changes I need to make so that we are on the safe side?

PR #767, opened 2022-02-21
Use flask-restx to replace flask-restplus

raoyitao:
* flask-restplus is an underlying package that is used in Testplan
  Webapp, however, it is not in development since 2019 and many of
  its dependencies have been updated and become incompatible (e.g.
  werkzeug), there is a similar library flask-restx which provides
  almost the same API so we can use it instead.
* Update the version of flask and werkzeug correspondingly.
* Update code for API change.
* Fix a CodeQL alert of "py/path-injection"


requirements-basic.txt
```diff
@@ -15,9 +15,12 @@ terminaltables
 pyparsing
 cycler
 requests>=2.4.3
-flask<2.0.0
-flask_restplus
+flask>2.0.0
+werkzeug>2.0.0
+flask_restx
 cheroot
+ipaddress
```
yuxuan-ms, line 22: python3 built-in package?

testplan/web_ui/web_app.py
```diff
@@ -94,20 +92,21 @@ def get(self, report_uid, assertions_uid):
 class TestplanAttachment(Resource):
     def get(self, report_uid, attachment_path):
         """Get an attachment for a specific Testplan report given their uids."""
-        attachment_path = os.path.abspath(
-            os.path.join(
-                app.config["DATA_PATH"], defaults.ATTACHMENTS, attachment_path
-            )
+        base_path = os.path.join(app.config["DATA_PATH"], defaults.ATTACHMENTS)
+        normlised_attachment_path = os.path.normpath(
```
yuxuan-ms, line 96: use pathlib?

testplan/web_ui/web_app.py
```diff
@@ -94,20 +92,21 @@ def get(self, report_uid, assertions_uid):
...
         )
-
-        if os.path.exists(attachment_path):
-            return send_from_directory(
-                directory=os.path.dirname(attachment_path),
-                filename=os.path.basename(attachment_path),
-            )
-        else:
+        if not (
+            normlised_attachment_path.startswith(base_path)  # security check
```
Pyifan, line 100: Why are we doing normpath and security check here?

testplan/web_ui/web_app.py
raoyitao: sure

testplan/web_ui/web_app.py
raoyitao: because `attachment_path` can contain substring like `../` which will cause security issues, and it is detected by CodeQL scanning.

Pyifan (approved):

PR #771, opened 2022-03-01
Remote refactor

kn-ms:
Some refactoring of the remote pool/worker related code

## Solution description
- The WorkerBase now serves as a base for Workers as Worker was doing to much
- UnboundRemoteResourceConfig that represent a remote resource configuration without a port assigned for possible
  remote resources that got their assignement async after construction time.
- environment now do not try to stop aborted resources

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/common/entity/base.py
```diff
@@ -1498,7 +1505,7 @@ def run(self):
             raise self._runnable.result
         return self._runnable.result
 
-    def _handle_abort(self, signum):
+    def _handle_abort(self, signum, frame):
```
Pyifan, line 1508: what is frame for?

Pyifan (changes requested):

Pyifan:
hi Krisztian, I'm not sure about the now even more complicated class hierarchy, is there any room for simplification? Otherwise this looks good and I'm going to look at the internal PR next.

testplan/common/entity/base.py
kn-ms: That is expected from a signal handler: https://docs.python.org/3/library/signal.html#signal.signal

kn-ms:
> hi Krisztian, I'm not sure about the now even more complicated class hierarchy, is there any room for simplification? Otherwise this looks good and I'm going to look at the internal PR next.

With the old remote worker config you were not able to capture configuration for remote worker without a set hostname:port. This make it difficult to use the remote worker in a composition where the remote worker can be recreated on different hosts from a previously captured config as hosts joining and leaving the plant/run

Pyifan (approved):

PR #773, opened 2022-03-01
Mark decorator

M6AI:
In current implementation, all assertions point to the function calling the assertion. This may not be the desired approach in the presence of utility and helper functions further away from the actual testcase.

## Solution description
The "mark" decorator allows setting the line number and filepath as the ones in the frame of the decorated function. If multiple "marks" are present, the closest to the assertion is used.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/testing/multitest/result.py
```diff
@@ -99,7 +101,23 @@ def __exit__(self, exc_type, exc_value, tb):
...
-def assertion(func):
+def mark(func: Callable) -> Callable:
+    """
+    Sets the decorated function's filepath and line-range in assertion state.
+    """
+
+    @wraps(func)
+    def wrapper(*args, **kwargs):
+        filepath = inspect.getfile(func)
+        lines, start = inspect.getsourcelines(func)
```
kn-ms, line 112: these values can be work out in the decorator once on decoration time, no need to do inspect an all call of the function

testplan/testing/multitest/result.py
```diff
@@ -99,7 +101,23 @@ def __exit__(self, exc_type, exc_value, tb):
...
+        assertion_state.filepath = filepath
+        assertion_state.line_range = range(start, start + len(lines))
```
kn-ms, line 114: file_path and line_range need to be managed as if they would be in a stack, so if there were something in thise attributes they need to be reset to that after the function call. As without that:

if you have an helper call chain like this

```
testcase -> marked_helper1 -> non_marked_helper -> marked_helper2 -> assertion1 
                                                \
                                                 -> assertion2

```
assertion 1 will correctly reported in marked_helper2 while assertion 1 will be reported in non_marked_helper as the frame walk will not see the marked helper in file_path, line_range

testplan/testing/multitest/result.py
```diff
@@ -143,11 +168,13 @@ def wrapper(result, *args, **kwargs):
         finally:
             if top_assertion:
                 assertion_state.in_progress = False
+                assertion_state.filepath = None
+                assertion_state.line_range = None
```
kn-ms, line 172: cleaning this in an assertion will make it not working for any more assertion. If you have a call chain where two assertion is called from a non marked helper which is called from a marked one, the first assertion will be reported properly into the marked helper, but the second one will be reported to the non marked one.

testplan/testing/multitest/result.py
```diff
@@ -99,7 +101,23 @@ def __exit__(self, exc_type, exc_value, tb):
 assertion_state = threading.local()
 
 
-def assertion(func):
+def mark(func: Callable) -> Callable:
```
kn-ms, line 104: I do not like the name `mark` for this decorator. 
Some ideas:
- `assertion_baseline`
- `assertion_line_target`
- `assertion_helper`
- `grouped_assertion`

doc/en/download/Assertions.rst
```diff
@@ -98,3 +98,16 @@ Required files:
 test_plan.py
 ++++++++++++
 .. literalinclude:: ../../../examples/Assertions/Plotly/test_plan.py
+
+
+Marking
```
kn-ms, line 103: We might need a bit of documentation that explain the use case we try to solve, that may point to this good example.

examples/Assertions/Marking/test_plan.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+@test_plan(name="Plan")
+def main(plan):
+    plan.add(MultiTest(name="MultiTest", suites=[Suite()]))
+    return plan
+
+
+if __name__ == "__main__":
+    main()
```
kn-ms, line 67: `sys.exit(not main())`

examples/Assertions/Marking/test_plan.py
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/env python
```
kn-ms, line 1: This is a filing example, suggest to turn it into a passing one with passing assertions, or log assertions. That way the example at least test if it is running without a problem. However this is a nice deterministic testsuite, the line numbers can be validated in a more unit/functional level of test.

testplan/testing/multitest/result.py
M6AI: Would prefer to go with a verb-like originating from mark to make sure it makes sense immediately from user-perspective. What about mark_line_target in this case? Or even, mark_trace_target?

examples/Assertions/Marking/test_plan.py
M6AI: Is your suggestion to add unit test to see whether we exactly point to where we want and turn the example into something that passes everywhere? I do like the visuality of the all failing because it immediately tells me where I wanted to look.

That said I am OK with that, just asking for clarification.

doc/en/download/Assertions.rst
M6AI: Sure thing.

M6AI: Added below. Please opine.

testplan/testing/multitest/result.py
Pyifan: Let me try to make this more difficult, lol. How about:

- @report_target
- @trace_line_target
- @mark_line_scope

M6AI: I like `report_target`, combines most of the things we wrote out in different names. @kn-ms what do you think?

kn-ms (commented):
The decorator name need to be changed in the docs and news fragment as well.

kn-ms (dismissed):

kn-ms (approved):

PR #774, opened 2022-03-04
Bugfix/rpyc timeout config

M6AI:
RPyC comes with a default timeout of 30 seconds. In case the remote service takes more time to start, the RPyC timeout happens.

## Solution description
Setting the default RPyC timeout to None allows the Testplan timeout to drive the status wait.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/common/remote/remote_service.py
+++ b/testplan/common/remote/remote_service.py
@@ -1,7 +1,12 @@
+"""
+Module implementing RemoteService class. Based on RPyC package.
+"""
 import os
 import re
 import signal
 import subprocess
+
+import rpyc
 
 from testplan.common.config import ConfigOption
 from testplan.common.entity import Resource, ResourceConfig
@@ -11,13 +16,8 @@ from testplan.common.remote.remote_resource import (
 )
 from testplan.common.utils.match import match_regexps_in_file
 from testplan.common.utils.path import StdFiles
-from testplan.common.utils.process import (
-    subprocess_popen,
-    kill_process,
-)
+from testplan.common.utils.process import subprocess_popen, kill_process
 from testplan.common.utils.timing import get_sleeper
-
-import rpyc
 
 RPYC_BIN = os.path.join(
     os.path.dirname(rpyc.__file__),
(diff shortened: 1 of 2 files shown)


M6AI:
Noticed that remote service tests are skipped due to no remote host configured. Tested it internally. Maybe mocking the connection is a good idea because currently the test has no real purpose in the repo as it is.

testplan/common/remote/remote_service.py
```diff
@@ -114,6 +125,19 @@ def starting(self):
...
+            "\tOut file: %s\n"
+            "\tErr file: %s",
+            self,
+            cmd,
+            self.runpath,
+            self.proc.pid,
+            self.std.out_path,
+            self.std.err_path,
+        )
         self.logger.debug(
```
kn-ms, line 141: I guess this need to go now in favor of the above

testplan/common/remote/remote_service.py
```diff
@@ -122,8 +146,8 @@ def starting(self):
             f"\tErr file: {self.std.err_path}\n"
         )
 
-    def _wait_started(self, timeout=None):
-
+    def _wait_started(self, timeout: float = None) -> None:
+        """TODO"""
```
kn-ms, line 150: any more todo here?

M6AI: ? There was no docstring originally, but for now, I have not gone through this completely and simply added a TODO.

testplan/common/remote/remote_service.py
M6AI: Disagree on this one. The last time we voted for lazy logging. Consistency, I think is more important, than nicely looking function call. A solution in the middle is to modify the message so that it wants less in a single go. In any case, I would stick to lazy logging unless we go full on f-strings/format method.

kn-ms: I just tried pointing out that you did not removed the fstring version of the log, if I see it correctly on github.

M6AI: Ah... sorry, let me remove that. :S

kn-ms (approved):

PR #799, opened 2022-04-20
Feature/junit exporter tpr

M6AI:
Conversion to JUnit XML output through TPR is not possible which makes it inconvenient when test evidence is needed in that format.

## Solution description
The existing XML Exporter can be integrated through a new "to" action in the CLI.

## Checklist:
- [x] Test
- [x] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/exporters/testing/http/__init__.py
+++ b/testplan/exporters/testing/http/__init__.py
@@ -42,7 +42,7 @@ class HTTPExporterConfig(ExporterConfig):
 
 class HTTPExporter(Exporter):
     """
-    JSON exporter.
+    HTTP exporter.
 
     :param http_url: Http url for posting data.
     :type http_url: ``str``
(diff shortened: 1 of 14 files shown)


M6AI:
One thing I am a bit unsure about: conversion works fine, but reading JUnit and then writing it back to JUnit results in considerably different naming patterns. Content and structure is kept, but the exporter adds more information and constructs testsuite and testcase names differently. May not be a problem at all, but raising this.

testplan/exporters/testing/http/__init__.py
```diff
@@ -42,26 +42,32 @@ def get_options(cls):
 
 class HTTPExporter(Exporter):
     """
-    Json Exporter.
+    JSON exporter.
```
kn-ms, line 45: This is an HTTPExporter

kn-ms (approved):

PR #800, opened 2022-04-20
Before `Worker` changed to `STOPPED` status it should be dead

raoyitao:
* A pool worker has a handler for a standalone thread or process.
  Before the pool exit and wait workers to be `STOPPED`, should check
  that worker is not alive through its handler.
* Avoid changing `Resource`'s status in its startig/stopping/aborting
  methods unless necessary you know clearly the process of how status
  of `Resource` changes. BTW, These methods are interfaces for derived
  class and in which better not call start/stop/abort methods because
  they manage the process of status transition.
* Fix a bug that pool monitor starts before workers actually start so
  it may fetch workers's out-of-date status (STOPPED after last run)
  and unexpectedly abort. Thread lock should be applied for workers.
* Update testcases.

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -1311,7 +1311,7 @@ class Resource(Entity):
             or self.status == self.STATUS.STARTED
         ):
             self.logger.debug(
-                "start() has been called on %r, skip starting", self
+                "start() has been called on %s, skip starting", self
             )
             return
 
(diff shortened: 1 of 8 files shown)


kn-ms (commented):
Could the tests extended to test this handles is no more live when get to stop?
Please add a news fragment

raoyitao:
> Could the tests extended to test this handles is no more live when get to stop? Please add a news fragment

I updated a few testcases for this change. An internal change on Treadmill Worker will work with this. There is chaos about managing various pool workers' status, so I want to make it clear. This change is not noticeable for users, seems we don't need a news fragment ?

Pyifan:
``

PR #802, opened 2022-04-25
Bugfix/parser timeout default

M6AI:
Recent change that allows 0 as a valid timeout means that 0 timeout from command line result in immediate timeout.

## Solution description
Reverting the change for the TestRunner.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


Pyifan (changes requested):
The news fragment needs to conform with new format

M6AI:
> The news fragment needs to conform with new format

Addressed.

PR #803, opened 2022-04-26
Use relative path for attachments in JSON report

raoyitao:
* The "_attachments" directory is always near the main Testplan JSON
  report, so the reletive path should be used, thus the saved reports
  can be moved freely to any location, TPR script is able to display
  the report in web UI.

### DIFF
--- a/testplan/exporters/testing/base.py
+++ b/testplan/exporters/testing/base.py
@@ -2,7 +2,7 @@
 """
 Implements base exporter objects.
 """
-import os
+import pathlib
 from typing import Dict, List, Optional
 from shutil import copyfile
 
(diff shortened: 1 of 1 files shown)


testplan/exporters/testing/base.py
```diff
@@ -191,6 +192,6 @@ def save_attachments(report: TestReport, directory: str) -> Dict[str, str]:
             dst_path = os.path.join(directory, dst)
             makedirs(os.path.dirname(dst_path))
             copyfile(src=src, dst=dst_path)
-            moved_attachments[dst] = dst_path
+            moved_attachments[dst] = os.path.join(".", ATTACHMENTS, dst)
```
Pyifan, line 195: if we need a relative path here, maybe change dst_path to relative path?

Pyifan (changes requested):
This will need a news fragment as well

testplan/exporters/testing/base.py
raoyitao: I think not necessary. `os.path.join(".", ATTACHMENTS, dst)` is a string written into json report. If we actually copy a file, using the absolute path is much reliable that it does not matter where to execute script.

Pyifan: But move_attachements[dst] shall be the actual place where we copy attachment to.

Pyifan (changes requested):

PR #808, opened 2022-05-09
Feature/append pre post always

M6AI:
Pre/post steps of the report are appended in one batch as a group only at the end of the MultiTest during post resource steps. In case of an environment startup failure, even the before_start step is lost.

## Solution description
Appending the group straight away during pre resource steps allows us to leverage mutability of entries and all non-skipped steps are present in the report.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/base.py
+++ b/testplan/testing/multitest/base.py
@@ -487,6 +487,7 @@ class MultiTest(testing_base.Test):
                     label="before_start", func=self.cfg.before_start
                 )
             )
+        self._add_step(self.append_pre_post_step_report)
 
     def pre_main_steps(self):
         """Runnable steps to be executed after environment starts."""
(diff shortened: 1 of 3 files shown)


testplan/testing/multitest/base.py
```diff
@@ -487,6 +487,7 @@ def pre_resource_steps(self):
                     label="before_start", func=self.cfg.before_start
                 )
             )
+        self._add_step(self.append_pre_post_step_report)
```
kn-ms, line 490: if there is no before_start configured
nothing create the `self._pre_post_step_report`  so the `self.append_pre_post_step_report` will not append anything, if I reading this correctly.

kn-ms: If this call is moved here, the docstring of `append_pre_post_step_report` is no more valid, needs update

kn-ms (approved):

PR #809, opened 2022-05-09
In dict match all content under excluded key should be marked ignored

raoyitao:
* If a key should be ignored for comparison, the value should still be
  compared with details displayed on web UI, that is, sub keys can be
   mapped one by one for easy viewing.
* On web UI not only the excluded keys are marked with grey color, but
  all that sub keys under it should all be marked "ignored" in grey.
* Update testcase.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/doc/newsfragments/1937_changed.dict_match_for_ignored_fields.rst
+++ b/doc/newsfragments/1937_changed.dict_match_for_ignored_fields.rst
@@ -1,0 +1,1 @@
+When :py:meth:`dict.match <testplan.testing.multitest.result.DictNamespace.match>` ``exclude_keys`` has nested dict, align the values under the same key between actual and expected on Web UI.
--- a/testplan/common/utils/comparison.py
+++ b/testplan/common/utils/comparison.py
@@ -491,20 +491,29 @@ def _cmp_dicts(lhs, rhs, ignore, only, report_mode, value_cmp_func):
             should_ignore = False
         return should_ignore
 
+    def mark_ignored(val):
+        """
+        Whatever the result is, just mark all flags of comparison to be "i"
+        (which means ignored), because the upper level key is in ignore list.
+        """
+        if val[0] == 1:
+            return 1, [mark_ignored(item) for item in val[1]]
+        elif val[0] == 2:
+            return 2, [
+                (item[0], Match.IGNORED, mark_ignored(item[2]))
+                for item in val[1]
+            ]
+        elif val[0] == 3:
+            return 3, Match.IGNORED, mark_ignored(val[2])
+        else:
+            return val
+
     results = []
     match = Match.IGNORED
+
     for iter_key, lhs_val, rhs_val in _idictzip_all(lhs, rhs):
-        if should_ignore_key(iter_key):
-            if report_mode == ReportOptions.ALL:
-                results.append(
-                    _build_res(
-                        key=iter_key,
-                        match=Match.IGNORED,
-                        lhs=fmt(lhs_val),
-                        rhs=fmt(rhs_val),
-                    )
-                )
-        else:
+        ignore_key = should_ignore_key(iter_key)
+        if report_mode == ReportOptions.ALL or not ignore_key:
             result = _rec_compare(
                 lhs_val,
                 rhs_val,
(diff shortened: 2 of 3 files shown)


tests/unit/testplan/testing/multitest/test_result.py
```diff
@@ -404,6 +404,37 @@ def test_flattened_comparison_result(self, dict_ns):
             and tea_1[4][0] == "REGEX"
         )
 
+    def test_exclude_key_on_nested_object(self, dict_ns):
```
kn-ms, line 407: how about to add tests that are having actual differences and so ignored failed comparisons.

raoyitao: yeah, I extended this testcase, adding nested objects which has sub keys that have different values.

testplan/common/utils/comparison.py
```diff
@@ -491,20 +491,29 @@ def should_ignore_key(key):
...
-                    _build_res(
-                        key=iter_key,
-                        match=Match.IGNORED,
-                        lhs=fmt(lhs_val),
-                        rhs=fmt(rhs_val),
-                    )
-                )
-        else:
+        ignore_key = should_ignore_key(iter_key)
+        if report_mode == ReportOptions.ALL or not ignore_key:
```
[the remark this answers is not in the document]
raoyitao: it correct, in order to make code concise, finally I made the logic hard to understand (report mode and ignore status are mixed). So I decide to make it as readable as possible, even without lambda, just writing `_rec_compare` twice does not matter.

kn-ms (commented):
Looking good, please add a newsfragment

raoyitao:
> 

yeah, a fragment is added.

doc/newsfragments/dict_match_for_ignored_fields.rst
```diff
@@ -0,0 +1 @@
+* When comparing 2 dictionaries even some fields are marked as "ignored", the corresponding values should also be compared, thus the result of `DictMatch` can be rendered on Web UI clearly that comparable items are presented side by side.
```
Pyifan, line 1: We don't need the asterisk in news fragment. And the description here is a bit too implementation oriented. Which assertion are we fixing here (dict.match?), what was the original issue (nested value in ignored?), try to describe the change from user perspective please.

Pyifan (changes requested):

doc/newsfragments/dict_match_for_ignored_fields.rst
raoyitao: ammended

doc/newsfragments/1937_changed.dict_match_for_ignored_fields.rst
```diff
@@ -0,0 +1 @@
+:py:meth:`dict.match <testplan.testing.multitest.result.DictNamespace.match>` should also compare the fields even their keys are marked as "ignored", thus for this assertion the nested values in a dictionary can be displayed in a compact view on Web UI (comparable fields are placed in the same row).
```
Pyifan, line 1: dict.match should compare "ignored" fields -- it doesn't make sense when reading from user perspective right?

btw, avoid "should" in news fragment -- what does it mean? it makes the announcement sounds like a issue description.

Pyifan (changes requested):

doc/newsfragments/1937_changed.dict_match_for_ignored_fields.rst
```diff
@@ -0,0 +1 @@
+In :py:meth:`dict.match <testplan.testing.multitest.result.DictNamespace.match>` assertion, the nested values under ignored keys are still compared one by one, so that 2 dictionaries can be displayed in a compact view on Web UI (comparable fields are placed in the same row).
```
Pyifan, line 1: When :py:meth:`dict.match <testplan.testing.multitest.result.DictNamespace.match>` ``exclude_keys`` has nested dict, align the values under the same key between actual and expected on Web UI.

Pyifan (approved):

PR #811, opened 2022-05-13
Change the place where Resource's `post_start` and `post_stop` methods are called

raoyitao:
* A resource's `start` method has the following steps:
  - pre_start
  - change status to starting
  - starting
  - wait stats to started
  - post_start

  the `stop` process is similar. `post` methods should be placed in
  `wait` so that user only need to call `wait` after `start` or `stop`
  and things become easier, this is important for `Driver`, e.g. its
  `restart` method is often called, which is made up of ony `start`,
  `_wait_started`, `stop` and `_wait_stoppped`.

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -177,7 +177,6 @@ class Environment:
         # Wait resources status to be STARTED.
         for resource in resources_to_wait_for:
             resource.wait(resource.STATUS.STARTED)
-            resource.post_start()
             resource.logger.debug("%s started", resource)
 
     def start_in_pool(self, pool):
@@ -211,7 +210,6 @@ class Environment:
         for resource in resources_to_wait_for:
             if resource not in self.start_exceptions:
                 resource.wait(resource.STATUS.STARTED)
-                resource.post_start()
                 resource.logger.debug("%s started", resource)
 
     def stop(self, is_reversed=False):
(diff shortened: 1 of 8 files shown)


testplan/common/entity/base.py
```diff
@@ -1294,13 +1289,12 @@ def stop(self):
             return
 
         self.logger.debug("Stopping %s", self)
-        self.status.change(self.STATUS.STOPPING)
         self.pre_stop()
+        self.status.change(self.STATUS.STOPPING)
```
Pyifan, line 1293: same, let's not make this change here.

Pyifan (changes requested):
async_start=False Resource restart() will call wait() twice right? once in start() once in wait().

Let's add a testcase that async_start=True/False resource restart() will call all pre_start/post_start/pre_stop/post_stop exactly once.

yuxuan-ms (approved):

raoyitao:
> async_start=False Resource restart() will call wait() twice right? once in start() once in wait().
> 
> Let's add a testcase that async_start=True/False resource restart() will call all pre_start/post_start/pre_stop/post_stop exactly once.

Fixed all the comment.

Pyifan (approved):

Pyifan (changes requested):
still missing testcase

Pyifan (approved):

PR #814, opened 2022-05-16
Append pre and post steps report unconditionally.

M6AI:
* Moved append_pre_post_step_report from post steps to pre steps.
* Updated unit test with proper order of expected data.
* Added news fragment.

Pre/post step report is appended as part of the post resource steps which makes it absent in case the MultiTest breaks earlier.

## Solution description
Pre/post step report is appended as an empty TestGroupReport in the pre resource steps if any pre/post step is configured and built upon continuously.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/doc/newsfragments/1918_changed.append_pre_post_report_unconditionally.rst
+++ b/doc/newsfragments/1918_changed.append_pre_post_report_unconditionally.rst
@@ -1,0 +1,1 @@
+Pre/post steps report is now appended before resource startup and built upon continuously.
--- a/testplan/testing/multitest/base.py
+++ b/testplan/testing/multitest/base.py
@@ -262,6 +262,7 @@ class MultiTest(testing_base.Test):
                 uid="Pre/Post Step Checks",
                 category=ReportCategories.TESTSUITE,
             )
+            self._pre_post_step_report.status = Status.PASSED
         return self._pre_post_step_report
 
     @property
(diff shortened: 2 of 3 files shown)


Pyifan (changes requested):
If we refer to what we do for TestCaseReport:
        if new_status == RuntimeStatus.FINISHED:
            self._status = Status.PASSED  # passed if case report has no entry

We are only setting default status to passed if testcase has been run. This makes better sense especially from interactive mode perspective. So a couple of suggestions:

1) we might want to observe how pre_post_step_report looks like on interactive mode
2) we need a testcase to verify 'if pre_post_step_report is empty, the parent is passed rather than unknown'
3) does it make better sense to call it "Before/After Step Checks"? (i know this is not part of the jira, but since we are here...lol)

M6AI:
> 1. we might want to observe how pre_post_step_report looks like on interactive mode
Let me run an example for this.
> 2. we need a testcase to verify 'if pre_post_step_report is empty, the parent is passed rather than unknown'
Let me reach out to you internally.
> 3. does it make better sense to call it "Before/After Step Checks"? (i know this is not part of the jira, but since we are here...lol)
Sure thing, we can have some voting around whether pre/post or before/after is more popular (latter seems to be the case with users as that is the name for attributes they pass to constuctor.

Pyifan (approved):

PR #817, opened 2022-05-24
Pass custom environment to ZooKeeperStandalone driver at stopping.

havasd:
This is needed if the ZK or Java is not available from the default paths and
user supplied custom env vars to override this.

When we run tests in an isolated environment or limited availability to Java or ZooKeeper and we provide custom environment variables which contains the necessary paths for Java or ZK, ZooKeeper fails to stop because it can't find Java properly. This is because the custom environment variables are not passed to the stop cmd.

## Solution description
Pass the custom environment variables to the stop cmd of the ZooKeeper driver.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


doc/newsfragments/2003_changed_zk_env.rst
```diff
@@ -0,0 +1 @@
+* Pass driver's custom environment to :py:class:`ZookeeperStandalone <testplan.testing.multitest.driver.zookeeper.ZookeeperStandalone>` stopping function.
```
Pyifan, line 1: Thanks for the PR! Can you drop the asterisk from the news fragment and rename the 
file to 2003_changed.zk_env.rst please. 

Otherwise all good! Thanks again!

Pyifan (changes requested):

doc/newsfragments/2003_changed_zk_env.rst
havasd: done

kn-ms (approved):

Pyifan (approved):

PR #818, opened 2022-05-26
Interactive mode to stop resources upon abort.

M6AI:
The abortion of the interactive mode does not stop the resources in itself. This keeps certain external processes alive after the interactive mode is exited.

## Solution description
The stopping of test resources for all resources can be added to the teardown method of the handler and takes care of the issue in a simple way.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/runnable/interactive/base.py
```diff
@@ -143,6 +143,8 @@ def teardown(self):
         if self._pool is None or self._http_handler is None:
             raise RuntimeError("setup() not run")
 
+        for test_uid in self.all_tests():
```
kn-ms, line 146: can you add a testcase please?

kn-ms (approved):

PR #821, opened 2022-05-30
Fix an issue that arg `testcase_report_target` cannot work as expected

raoyitao:
* the argument `testcase_report_target` is used for Multitest object
  to make all testcases marked as report target. However, it cannot
  work well for parametrized testcases, the definition of generated
  function will unexpectedly be marked. But it will not affect the
  `report_target` decorator which can still work as expected.
* Also it should be work with "pre" and "post" decorator when the
  assertions is written in pre-testcase or post-testcase methods.
* Add testcases.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/result.py
+++ b/testplan/testing/multitest/result.py
@@ -101,13 +101,30 @@ class ExceptionCapture:
 assertion_state = threading.local()
 
 
-def report_target(func: Callable) -> Callable:
+def report_target(func: Callable, ref_func: Callable = None) -> Callable:
     """
     Sets the decorated function's filepath and line-range in assertion state.
+    If the target function is a parametrized function, should refer to its
+    parametrized template to find information of the original function.
+
+    :param func: The target function about which the information of
+        source path and line range will be retrieved.
+    :param ref_func: The parametrized template if `func` is a generated
+        function, otherwise ``None``.
     """
-    filepath = inspect.getfile(func)
-    lines, start = inspect.getsourcelines(func)
-    line_range = range(start, start + len(lines))
+    module = inspect.getmodule(ref_func or func)
+    filepath = module.__file__ if module else None
+    if filepath is None:
+        try:
+            filepath = inspect.getsourcefile(ref_func or func)
+        except TypeError:
+            pass
+
+    try:
+        lines, start = inspect.getsourcelines(ref_func or func)
+        line_range = range(start, start + len(lines))
+    except OSError:
+        pass
 
     @wraps(func)
     def wrapper(*args, **kwargs):
@@ -148,6 +165,8 @@ def assertion(func: Callable) -> Callable:
                                     in assertion_state.line_range
                                 ):
                                     break
+                            else:
+                                frame = call_stack[1]
                         entry.file_path = os.path.abspath(frame.filename)
                         entry.line_no = frame.lineno
                     finally:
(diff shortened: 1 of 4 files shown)


tests/unit/testplan/testing/multitest/test_result.py
```diff
@@ -91,6 +100,33 @@ def test_group_marking_multitest(mockplan, flag):
...
+        "A0": get_line_no(ParametrizedGroupMarking.case, 2),
+        "B0": get_line_no(ParametrizedGroupMarking.case, 3)
+        if flag
+        else get_line_no(helper, 1),
+        "C0": get_line_no(intermediary, 2),
+        "A1": get_line_no(ParametrizedGroupMarking.case, 2),
+        "B1": get_line_no(ParametrizedGroupMarking.case, 3)
+        if flag
+        else get_line_no(helper, 1),
+        "C1": get_line_no(intermediary, 2),
```
M6AI, line 125: No concern, but maybe at this point this became a bit hard to read due to multiple ternary-like, at least for me. Why we have two times the `if flag else` piece?

testplan/testing/multitest/result.py
```diff
@@ -148,6 +150,13 @@ def wrapper(result, *args, **kwargs):
                                     in assertion_state.line_range
                                 ):
                                     break
+                            else:
+                                raise RuntimeError(
+                                    "Cannot find a matching frame in the call"
+                                    f" stack of function `{func.__name__}`."
+                                    f" Refer to file: {call_stack[1].filename}"
+                                    f" Line: {call_stack[1].lineno}"
+                                )
```
M6AI, line 159: This exception is actually getting thrown in the Windows job. Please see if is expected and we should assert it being raised otherwise just fix the test.

M6AI:
News fragment is missing, please add it.

tests/unit/testplan/testing/multitest/test_result.py
raoyitao: ugly code, I have updated it.

testplan/testing/multitest/result.py
raoyitao: > News fragment is missing, please add it.

yes, add a news fragment.

raoyitao: this exception here is for testing, because I think this for loop sometimes cannot work correctly. At least I found when assertion is written in pre testcase or post testcase, the result is not correct. I have s simple fix for it.

Pyifan (dismissed):

testplan/testing/multitest/result.py
```diff
@@ -101,13 +101,25 @@ def __exit__(self, exc_type, exc_value, tb):
 assertion_state = threading.local()
 
 
-def report_target(func: Callable) -> Callable:
+def report_target(func: Callable, ref_func: Callable = None) -> Callable:
     """
     Sets the decorated function's filepath and line-range in assertion state.
+    If the target function is a parametrized function, should refer to its
+    parametrized template to find information of the original function.
```
M6AI, line 108: Can you please add this to a `:param ref_func:` directive? Given this is a decorator, I doubt the `func` parameter or the return value needs to be explained at docstring level, but I think the `ref_func` is an important and non-trivial addition to the decorator. Also helps in IDE.

M6AI (changes requested):

Pyifan (dismissed):

testplan/testing/multitest/result.py
raoyitao: doc string added.

M6AI (approved):

M6AI (approved):
Thanks for removing the `type` directives as per offline discussion.

PR #822, opened 2022-05-31
Make pool abort more robust.

Pyifan:
Pool abort leaves thread pool behind.

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [X] MS info leakage check


testplan/common/utils/process.py
```diff
@@ -229,7 +231,7 @@ def execute_cmd(
     start_time = time.time()
 
     handler = subprocess.Popen(cmd, stdout=stdout, stderr=stderr, env=env)
-    stdout, stderr = handler.communicate()
+    stdout, stderr = handler.communicate(timeout=timeout)
```
kn-ms, line 234: Should not the subprocess being cleaned killed, and logs collected in case of timeout? This might be able to leave behind running processes. The outer scope would not have the handle to clean up.

yuxuan-ms (approved):

testplan/common/utils/process.py
Pyifan: i agree, i have removed this change ;)

PR #836, opened 2022-07-11
Bugfix/ignored display unordered

M6AI:
In current implementation, ignored nested structures in dictionary matching result in a faulty display due to lazy formatting of values below top level.

## Solution description
The deep recursive comparison gets all pairing right and we can introduce an optional behavior by setting the value comparison function to None and setting the match value conditional on that rather than executing the actual comparison.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/common/utils/comparison.py
+++ b/testplan/common/utils/comparison.py
@@ -556,14 +556,14 @@ def _rec_compare(
     ):
         return _build_res(
             key=key,
-            match=Match.PASS if lhs_cat == rhs_cat else Match.FAIL,
+            match=Match.IGNORED if value_cmp_func is None else Match.PASS if lhs_cat == rhs_cat else Match.FAIL,
             lhs=fmt(lhs),
             rhs=fmt(rhs),
         )
 
     ## CALLABLES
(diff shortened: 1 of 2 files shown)


M6AI:
> I like this solution, very smart! Can we add a new frag please?

Yes, working on it. Glad you like it, but let me share the credits with Yitao, who used the recursive in the original solution already, and with Krisztian who suggested the approach as a possibility.

testplan/common/utils/comparison.py
```diff
@@ -471,16 +474,32 @@ def _partition(results):
     return lhs_vals, rhs_vals
 
 
-def _cmp_dicts(lhs, rhs, ignore, only, report_mode, value_cmp_func):
+def _cmp_dicts(
+    lhs: Dict,
+    rhs: Dict,
+    ignore: Container,
```
M6AI, line 480: Wanted to subscript with ```[Hashable]```, but the linter catches it as unsubsrciptable. It appears to be known issue.

Pyifan (approved):

Pyifan (approved):

PR #855, opened 2022-08-25
Bugfix/one sided repeated group display

M6AI:
Reported in https://github.com/morganstanley/testplan/issues/848. The root cause is as follows: if in a dictionary comparison there is either a unique key referencing a container or the key is in both dictionaries, but one side is referencing a container value while the other not (container as handled in Testplan), then visual separators for separating groups break the construction of the result table in dictionary flattening.

## Solution description
For now, the visual separator row cannot be removed simply from the generator as it seems to be going "spaghetti" with some other logic. The row is of a special form and can be captured if only appears on one side of the comparison. This is exploited for a higher up filtering, just before the table is put together.

## Checklist:
- [x] Test
- [x] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/common/utils/convert.py
+++ b/testplan/common/utils/convert.py
@@ -98,7 +98,6 @@ def expand_values(rows, level=0, ignore_key=False, key_path=None, match=""):
                     match=match,
                 )
         elif isinstance(val, list):
-            #if key is not Absent:  # if key is Absent and it is a list, then we bypass
             yield (tuple(key_path), level, key, match, "")
             yield from expand_values(
                 val, level=level, key_path=key_path, match=match
--- a/testplan/common/utils/convert.py
+++ b/testplan/common/utils/convert.py
@@ -1,11 +1,24 @@
 """Conversion utilities."""
 import itertools
+from typing import Union, Tuple, Iterable, Callable, List, Sequence
 
 from .reporting import Absent
 
 
-def make_tuple(value, convert_none=False):
-    """Shortcut utility for converting a value to a tuple."""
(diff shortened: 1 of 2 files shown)


testplan/common/utils/convert.py
```diff
@@ -233,12 +240,26 @@ def flatten_dict_comparison(comparison):
...
+                and left[0][2] != right[0][2]
+            ):
                 lpart = left.pop(0)
-            elif len(left[0][0]) < len(right[0][0]):
+            # NOTE: same as above but from right-hand side perspective
+            elif (
+                len(left[0][0]) < len(right[0][0])
+                or right[0][2] is Absent
+                and left[0][2] != right[0][2]
+            ):
```
M6AI, line 258: These two updates to the conditionals are the actual changes. The row that is causing the mismatch upon the scenario discussed is readily identified (key is Absent on exactly one side). The behavior should be as in case of longer keypath so it is just incremental condition to the keypath length check, then handling should be identical.

tests/unit/testplan/common/utils/test_convert.py
```diff
@@ -0,0 +1,40 @@
...
+        [0, 54, "Passed", ("int", "2"), ("int", "2")],
+        [0, 851, "Failed", ("int", "4"), (None, "ABSENT")],
+        [0, 20001, "Failed", (None, "ABSENT"), ("str", "SAMPLE")],
+    ],
+]
+
+
+@pytest.mark.parametrize(
+    "comparison, result_table", zip(comparisons, result_tables)
+)
```
M6AI, line 37: Right now this is a single parameter pair, but this may be good to extend with more examples.

Pyifan (approved):

PR #857, opened 2022-08-31
Display Driver metadata

M6AI:
Driver metadata like name, context, status, various paths, host and port where it applies, and alike are not easily accessed and displayed in Testplan report. We aim to provide a simple solution for logging this information on-demand using customizable metadata extractors.

## Solution description
In the internal layer of the package we already log OS, hardware, pwd, and similar information using log or dict log. Enhancing the existing Driver implementation with a metadata extractor method using a customizable extraction function config option, we can add a simple utility that can derive metadata from the environment and save it to the report.

## Checklist:
- [x] Test
- [x] Documentation (API)
- [ ] MS info leakage check


testplan/common/utils/helper.py
```diff
@@ -0,0 +1,189 @@
...
+"""
+
+__all__ = [
+    "log_pwd",
+    "log_hardware",
+    "log_cmd",
+    "log_environment",
+    "attach_log",
+    "attach_driver_logs_if_failed",
+    "extract_metadata",
```
Pyifan, line 17: i would suggest a slightly more explicit name e.g extract_driver_info

testplan/testing/multitest/driver/base.py
```diff
@@ -159,6 +142,7 @@ def __init__(
         super(Driver, self).__init__(**options)
         self.extracts = {}
         self._file_log_handler = None
+        self._metadata_extractor = metadata_extractor
```
Pyifan, line 145: we can refer to this as self.cfg.metadata_extractor, and no need for an additional data field.

testplan/testing/multitest/driver/base.py
```diff
@@ -411,7 +408,15 @@ def get_lines_at_tail(log_file, max_count):
...
-    def __repr__(self):
-        """String representation."""
+    def extract_metadata(self) -> Dict:
+        """
+        Extracts driver metadata as described in the extractor function.
 
+        :return: driver metadata
+        """
+        # pylint: disable=not-callable
+        return self.metadata_extractor(self)
```
Pyifan, line 418: what if metadata_extractor is None?

testplan/common/utils/helper.py
```diff
@@ -0,0 +1,189 @@
...
+    Saves metadata of each driver to the report.
+
+    :param env: environment
+    :param result: testcase result
+    """
+    for resource in env:
+        result.dict.log(
+            resource.extract_metadata(),
+            description=f'Metadata of driver "{resource.name}"',
+        )
```
Pyifan, line 101: We need all extract_metadata of different drivers return somewhat uniform data. And i would prefer a table.log to multiple dict.log. How about define a DriverInfo class, and require the callable to return an instance of such? Must have fields: name, klass, listening ports, outgoing ports, admin port; nice to have fields: binary, config, logpath etc.

examples/Best Practice/common_utilities/test_plan.py
```diff
@@ -0,0 +1,81 @@
+#!/usr/bin/env python
```
Pyifan, line 1: Let's add a separate example for extract_driver_info, that has a few interconnected drivers (e.g client -> app -> server), and then from the logged driver info, we can figure out how they are connected.

Pyifan (changes requested):

testplan/testing/multitest/driver/base.py
M6AI: Good catch, addressing.

examples/Best Practice/common_utilities/test_plan.py
M6AI: Yeah, sorry, this is actually incomplete, I did not even add anything to it, just moved the example.

testplan/common/utils/helper.py
M6AI: We can do that. The concept here was that each driver is logged separate so the table is more "atomic". I have no problem with the single table.log.

testplan/testing/multitest/driver/base.py
M6AI: Sure, let me remove it and the property as well.

doc/newsfragments/1983_new.driver_metadata_and_common_utilities.rst
```diff
@@ -0,0 +1,2 @@
+Added a callable metadata extractor attribute to :py:class:`Driver <testplan.testing.multitest.driver.base.Driver>` objects that can be called individually or for the entire environment on-demand through :py:meth:`extract_metadata <testplan.common.utils.helper.extract_metadata>`
+Added new helper functions for reporting of generic testplan execution information under :py:module:`helper <testplan.common.utils.helper>`.
```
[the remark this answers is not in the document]
M6AI: Please check the internal PR, the namespace is still there so no break is expected. The attachment of driver logs look for binlog, which is internal, thus it remained there too.

examples/Best Practice/Common Utilities/test_plan_metadata.py
```diff
@@ -0,0 +1,124 @@
...
+                    name="client",
+                    host=context("server", "{{host}}"),
+                    port=context("server", "{{port}}"),
+                    metadata_extractor=metadata_extractor_client,
+                ),
+            ],
+            before_start=before_start,
+            after_start=after_start
+        )
+    )
```
M6AI, line 120: @Pyifan I did not add an App into the environment as our simplest examples for that scenario seem a bit too much already to me if I don a user hat. Please take a look, if this suffices, if not, I will add something that interacts with these two resources.

Pyifan (changes requested):

examples/Best Practice/Common Utilities/test_plan_metadata.py
```diff
@@ -0,0 +1,106 @@
...
+    :return: driver name, driver class, host, and connecting port metadata
+    """
+    return DriverMetadata(
+        name=driver.name,
+        driver_metadata={
+            "class": driver.__class__.__name__,
+            "host": driver.host or driver.cfg.host,
+            "port": driver.port or driver.cfg.port,
+            "server_port": driver.server_port,
+        },
```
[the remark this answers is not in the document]
M6AI: Let me give it a thought, I imagined Connection more to be a labelled graph edge kind of structure which can be processed almost readily by out-of-the-box visualization tools. That said no immediate objection to the above, we can always just refine.

testplan/testing/multitest/driver/base.py
```diff
@@ -411,7 +442,21 @@ def get_lines_at_tail(log_file, max_count):
...
-        """String representation."""
+    def extract_driver_metadata(self) -> DriverMetadata:
+        """
+        Extracts driver metadata as described in the extractor function.
 
+        :return: driver metadata
+        """
+        if self.cfg.metadata_extractor is None:
+            raise ValueError(
+                "To extract driver metadata, a valid `metadata_extractor`"
```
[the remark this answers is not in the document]
M6AI: Name is basically used just for access. We can provide empty metadata by default, making metadata also a field coming with a default factory for now.

testplan/common/utils/context.py
```diff
@@ -40,8 +40,12 @@ def __init__(self, driver: str, value: str):
         Create a new context value
         """
         self.driver = driver
+        self.value = value
         self.template = parse_template(value)
 
+    def __str__(self):
+        return f"{self.driver}, {self.value}"
```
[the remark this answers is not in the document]
M6AI: Sure, making the change.

testplan/testing/multitest/driver/tcp/client.py
```diff
@@ -88,6 +88,10 @@ def port(self):
         """Client port number assigned."""
         return self._port
 
+    @property
+    def server_port(self):
+        return self._client.port if self._client is not None else None
```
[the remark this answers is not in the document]
M6AI: Do you mean to add a new `_server_port` attribute in constructor, probably set to `None`, then in `starting` assign it the value that is currently the `server_port` local variable in `starting`? I mean this:
```
    def starting(self):
        """Start the TCP client and optionally connect to host/post."""
        super(TCPClient, self).starting()
        server_host = expand(self.cfg.host, self.context)
        server_port = expand(self.cfg.port, self.context, int) # HERE!
        ...
```

kn-ms (approved):

PR #859, opened 2022-09-15
Reset uid when saving report in interactive mode.

yuxuan-ms:
There are some potential issues if we use same testcase name in different testsuites. 

## Solution description
Reset uid when saving report in interactive mode.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/doc/newsfragments/2117_changed.interactive_uid.rst
+++ b/doc/newsfragments/2117_changed.interactive_uid.rst
@@ -1,0 +1,1 @@
+Fixed an issue where report saved from interactive mode having problem with testcases having the same name.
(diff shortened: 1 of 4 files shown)


doc/newsfragments/2117_changed.interactive_uid.rst
```diff
@@ -0,0 +1 @@
+Reset uid when saving report in interactive mode.
```
kn-ms, line 1: This is a user facing releasenote, this does not tell to much to the user. Can it be like: "Fixed an issue where report saved from interactive mode having problem with testcases having the same name" or similar, that describe what the user will see?

Pyifan (approved):

Pyifan (approved):

PR #884, opened 2022-12-22
Feature/http server

rnemes:
Added the possibility to access the data of the request received in HTTP Server object.

## Solution description
Introduced a ReceivedRequest class and added seome helper functions

## Checklist:
- [x] Test
- [x] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/examples/Transports/HTTP/http_basic.py
+++ b/examples/Transports/HTTP/http_basic.py
@@ -1,5 +1,5 @@
 """Tests HTTP requests between a server and a client."""
-import json
+import json, time
 
 from testplan.common.utils.context import context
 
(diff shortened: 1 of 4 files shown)


M6AI:
Black formatter needs to be applied.

examples/Transports/HTTP/http_basic.py
```diff
@@ -1,5 +1,5 @@
 """Tests HTTP requests between a server and a client."""
-import json
+import json, time
```
M6AI, line 2: For consistency, please push `import time` onto a new line. Wondering if this import is used. Probably was there for local testing?

M6AI (changes requested):
Added a couple of comments on todos, lets talk offline as well. The example is essentially a functional test, wonder if some unit test would make sense too.

testplan/testing/multitest/driver/http/server.py
```diff
@@ -14,27 +15,124 @@
...
+        ):
+            return json.loads(self.raw_data)
+        else:
+            return None
+
+
+class HTTPResponse:
+    """
+    HTTPResponse containing the status code, headers and content.
+    """
```
M6AI, line 99: Looks like a namedtuple to me. Lets discuss.

rnemes: This code was already here, I only moved it upper because HTTPRequestHandler uses it.

Pyifan (commented):
This looks pretty good! Well done Robi! I will leave it to you and Marci to figure out any remaining details that might need to be polished.

M6AI (approved):

PR #893, opened 2023-01-18
Automatically jump to first failed entry in WebUI

rnemes:
Testplan should open the report at the first failed testcase.

## Solution description
When the user doesn't specify any UID in the URL, the frontend will look for the first failure, compose a new URL and automatically redirect. Need to pass the selected UID to left navigation panel to automatically expand it.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/web_ui/testing/src/Report/BatchReport.js
+++ b/testplan/web_ui/testing/src/Report/BatchReport.js
@@ -13,6 +13,7 @@ import {
   GetCenterPane,
   GetSelectedEntries,
   MergeSplittedReport,
+  findFirstFailure,
   filterReport,
   getSelectedUIDsFromPath,
 } from "./reportUtils";
@@ -85,7 +86,7 @@ class BatchReport extends React.Component {
       filteredReport.report.status === "failed"
       || filteredReport.report.status === "error"
     )
-      ? this.findFirstFailure(filteredReport.report)
+      ? findFirstFailure(filteredReport.report)
       : [filteredReport.report.uid];
 
     const redirectPath = this.props.match.params.selection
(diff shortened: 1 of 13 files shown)


testplan/web_ui/testing/src/Report/BatchReport.js
```diff
@@ -117,54 +127,87 @@ class BatchReport extends React.Component {
...
+        if (reportEntry.entries[entry].status === "failed"
+          || reportEntry.entries[entry].status === "error") {
+          return selection.concat(
+            this.findFirstFailure(reportEntry.entries[entry])
+          );
+        }
+      }
+    }
+  }
+
```
M6AI, line 210: I would move this to reportUtils.js. Technically, this works independent of the BatchReport as a routine.

testplan/web_ui/testing/src/Report/BatchReport.js
```diff
@@ -77,23 +80,30 @@ class BatchReport extends React.Component {
       processedReport,
       this.state.filteredReport.filter
     );
+    
+    const firstFailedUID = (
+      filteredReport.report.status === "failed"
+      || filteredReport.report.status === "error"
+    )
```
M6AI, line 87: Maybe worth a comment that it is not looking for "non-passed", but explicitly for failed or error. For instance unknown status is exempt here.

testplan/web_ui/testing/src/Report/BatchReport.js
rnemes: Agree, moving to reportUtils

kn-ms (approved):

PR #896, opened 2023-01-31
Message content for Fail assertion

M6AI:
Fail assertion only carries content in the form of description which can become unaesthetic in case of longer explanation for the intended failure.

## Solution description
Added message attribute and associated rendering to fail assertion in the likeness of log.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/entries/assertions.py
+++ b/testplan/testing/multitest/entries/assertions.py
@@ -72,7 +72,7 @@ class Assertion(BaseEntry):
 
     meta_type = "assertion"
 
-    def __init__(self, description=None, category=None, flag=None):
+    def __init__(self, description=None, category=None, flag=None): 
         super(Assertion, self).__init__(
             description=description, category=category, flag=flag
         )
@@ -122,6 +122,24 @@ class IsFalse(IsTrue):
 
 
 class Fail(Assertion):
+    def __init__(self, description=None, category=None, flag=None, message=None):
+        if isinstance(message, str):
+            self.message = message
+        elif isinstance(message, bytes):
+            self.message = message.decode()
+        else:
+            self.message = pprint.pformat(message)
+
+        if not description:
+            description = next((l for l in self.message.split("\n") if l), "")
+        if len(description) > 80:
+            description = description[0:80] + "..."
+
+        super(Fail, self).__init__(
+            description=description, category=category, flag=flag
+        )
+        
+                
     def evaluate(self):
         return False
 
--- a/testplan/testing/multitest/entries/assertions.py
+++ b/testplan/testing/multitest/entries/assertions.py
@@ -72,7 +72,7 @@ class Assertion(BaseEntry):
 
     meta_type = "assertion"
 
-    def __init__(self, description=None, category=None, flag=None): 
+    def __init__(self, description=None, category=None, flag=None):
         super(Assertion, self).__init__(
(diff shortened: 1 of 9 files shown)


M6AI:
Credit to John, Adzagogo. PR still misses newsfragment and a few docstrings, will apply soon.

testplan/testing/multitest/entries/stdout/assertions.py
```diff
@@ -463,10 +463,22 @@ def get_assertion_details(self, entry):
...
 @registry.bind(assertions.Fail)
 class FailRenderer(AssertionRenderer):
-    def get_header_text(self, entry):
-        """Always return text in red as it is an explicit failure."""
-        return Color.red(entry.description)
+    def get_header(self, entry):
+        if entry.description:
+            return Color.red(entry.description)
+        elif isinstance(entry.message, str):
+            return Color.red(str(entry.message))
```
M6AI, line 470: Not sure if we need this.

testplan/testing/multitest/entries/stdout/assertions.py
```diff
@@ -463,10 +463,22 @@ def get_assertion_details(self, entry):
...
+            return self.get_default_header(entry)
 
+    def get_assertion_details(self, entry):
+        if isinstance(entry.message, str):
+            if entry.description:
+                return Color.red(str(entry.message))
+            else:
+                return None
+        else:
+            return pprint.pformat(entry.message)
```
M6AI, line 481: Missing import for pprint.

testplan/testing/multitest/entries/assertions.py
```diff
@@ -122,6 +122,24 @@ def evaluate(self):
...
+        if isinstance(message, str):
+            self.message = message
+        elif isinstance(message, bytes):
+            self.message = message.decode()
+        else:
+            self.message = pprint.pformat(message)
+
+        if not description:
+            description = next((l for l in self.message.split("\n") if l), "")
+        if len(description) > 80:
```
M6AI, line 137: Needs indentation.

testplan/testing/multitest/entries/stdout/assertions.py
Pyifan: agree, i think we always have a description

testplan/testing/multitest/entries/assertions.py
```diff
@@ -122,6 +123,25 @@ def evaluate(self):
 
 
 class Fail(Assertion):
+    def __init__(
+        self, description=None, category=None, flag=None, message=None
```
Pyifan, line 127: To achieve the same behavior with result.log, shall we have 'message' as the first and required parameter?

testplan/testing/multitest/result.py
```diff
@@ -1608,7 +1608,13 @@ def log_code(self, code, language="python", description=None):
         return entry
 
     @assertion
-    def fail(self, description, category=None, flag=None):
+    def fail(
```
Pyifan, line 1611: We usually use result.log like this:
result.log("This is a log assertion")
result.log(cmd_output, description="execute {cmd}")

I would expect result.fail to exhibits same usage and behavior. Thus shall we change the order of the parameters, and make it consistent with result.log?

PR #897, opened 2023-02-02
Fix reloader selection login

kn-ms:
If multiple module depends on the same changed module in interactive mode, the reload will just reload one of them.

## Solution description
- now take into account if a dependency was already reloaded on a different path of the graph and consider it as a reloaded
- updated testcase to have this scenario
- updated example to show this scenario

## Checklist:
- [X] Test
- [ ] Documentation (API)
- [X] MS info leakage check


Pyifan (approved):

M6AI (approved):
Confirming Yifan's approval until she is back on team.

PR #898, opened 2023-02-03
fix change_directory to restore PWD properly

kn-ms:
Change Directory was not setting back PWD to it's original value, but set it to the previous actual working dir.

## Solution description
Now the previous value is restored
No specific test but this fixing some task related flaky testcases

## Checklist:
- [X] Test
- [ ] Documentation (API)
- [X] MS info leakage check


Pyifan (approved):

M6AI (approved):
Confirming Yifan's approval until she is back on team.

PR #901, opened 2023-02-07
Interactive tree view

M6AI:
Interactive mode is not supported in UI tree view.

## Solution description
For now, the code is duplicated and appropriately updated from a combination of the static tree view and the interactive entries.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/web_ui/testing/src/Nav/InteractiveNav.js
```diff
@@ -35,27 +35,57 @@ const InteractiveNav = (props) => {
...
-        filter={null}
-        displayEmpty={true}
-        displayTags={false}
-        displayTime={false}
-        selectedUid={GetSelectedUid(props.selected)}
-        handleClick={props.handleClick}
-        envCtrlCallback={props.envCtrlCallback}
-        url={props.url}
-      />
+      {renderNavigation(props)}
```
M6AI, line 38: Change here is just preparation for unifying the two `Nav`.

testplan/web_ui/testing/src/Nav/InteractiveTreeView.js
```diff
@@ -0,0 +1,326 @@
+import React from 'react';
```
M6AI, line 1: This is essentially duplication of the TreeView, but entries are interactive entries inside each node.

testplan/web_ui/testing/src/Nav/NavList.js
```diff
@@ -1,67 +1,115 @@
...
-import {STATUS} from "../Common/defaults";
+import {
+    STATUS,
+    RUNTIME_STATUS,
+    NAV_ENTRY_ACTIONS,
+} from "../Common/defaults";
 
 /**
  * Render a vertical list of all the currently selected entries children.
+  * Render a vertical list of all the currently selected entries children for
```
zhenyu-ms, line 17: Looks like sort of duplicate with the line above.

M6AI: Indeed. Fixing, thanks @zhenyu-ms for catching it.

kn-ms (approved):

PR #903, opened 2023-02-16
Feature/allow "complex" test tasks being scheduled by Testplan

zhenyu-ms:
Make Testplan capable of scheduling Multitests using lambdas & nested functions.

## Solution description
Use `dill` instead of built-in `pickle` for serialization.

We tend to have more reference in serialized data, since source code is available remotely.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


tests/functional/testplan/runners/pools/func_pool_base_tasks.py
```diff
@@ -1,24 +1,21 @@
...
-from testplan.common.utils.path import fix_home_prefix
-from testplan.testing.multitest import MultiTest, testsuite, testcase
-from testplan.testing.multitest.base import MultiTestConfig
 from testplan.common.utils.strings import slugify
+from testplan.testing.multitest import MultiTest, testcase, testsuite
+from testplan.testing.multitest.base import MultiTestConfig
 
 
 @testsuite
 class MySuite:
```
Pyifan, line 13: Can we de-dup MySuite between this file and test_pool_base? (and get_mtest too)

Pyifan (changes requested):

tests/functional/testplan/runners/pools/func_pool_base_tasks.py
zhenyu-ms: As discussed offline, we will keep both since one represents testsuite in current module and the other for testsuite in imported module.

Pyifan (approved):

PR #906, opened 2023-02-27
Capture driver logs on error

Attakay78:
Capture driver error logs when driver fails on start or stop.

### DIFF
--- a/doc/newsfragments/1831_new.driver_log_capture.rst
+++ b/doc/newsfragments/1831_new.driver_log_capture.rst
@@ -1,0 +1,1 @@
+Drivers now can attach some extra logs to the report in case of startup/shutdown failure, with implementing: :py:meth:`fetch_error_log <testplan.common.entity.base.Resource.fetch_error_log>`.
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -1413,9 +1413,10 @@ class Resource(Entity):
 
     def fetch_error_log(self) -> List[str]:
         """
-        Fetch error message from the log files of resource.
+        Override this method in Resource subclasses to automatically add any
+        useful logs into the report, in case of startup/shutdown exception.
 
-        :return: text from log file
+        :return: text from log files
         """
         return []
 
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -188,6 +188,14 @@ class Environment:
             for resource in self._resources
         )
 
+    def _fetch_log_messages(self, resource, message):
+        if hasattr(resource, 'fetch_error_log'):
+            fetch_msg = '\n'.join(resource.fetch_error_log())
+            msg = f"{message}\n{fetch_msg}\n"
+            return msg
+
+        return message
+
     def start(self):
         """
         Starts all resources sequentially and log errors.
@@ -204,6 +212,7 @@ class Environment:
                 msg = "While starting resource [{}]\n{}".format(
                     resource.cfg.name, traceback.format_exc()
                 )
+                msg = self._fetch_log_messages(resource=resource, message=msg)
                 resource.logger.error(msg)
                 self.start_exceptions[resource] = msg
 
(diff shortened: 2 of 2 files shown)


testplan/common/entity/base.py
```diff
@@ -188,6 +188,14 @@ def all_status(self, target):
             for resource in self._resources
         )
 
+    def _fetch_log_messages(self, resource, message):
+        if hasattr(resource, 'fetch_error_log'):
```
kn-ms: if you expect content of the environment to be Resources and that class has implementation of fetch_error_log you do not need the check. If you enable non Resources, then I suggest to go with Protocol as a much modern alternative.

testplan/common/entity/base.py
```diff
@@ -188,6 +188,14 @@ def all_status(self, target):
             for resource in self._resources
         )
 
+    def _fetch_log_messages(self, resource, message):
```
kn-ms, line 191: I do not think this functions should have a role to take a prefix message and append to it, this just should format the part that is the log message and format it, do the concatenation where you prepre the message.

let this has a single role

testplan/common/entity/base.py
```diff
@@ -204,6 +212,7 @@ def start(self):
                 msg = "While starting resource [{}]\n{}".format(
                     resource.cfg.name, traceback.format_exc()
                 )
+                msg = self._fetch_log_messages(resource=resource, message=msg)
                 resource.logger.error(msg)
                 self.start_exceptions[resource] = msg
```
kn-ms, line 217: These lines are kind of repeted, the can be factored out to something like

```python
def record_resource_exception(message, resource, exception, msg_store):
    msg = "\n".join( message.format(resource.cfg.name), 
                     traceback.format_exc(), 
                     self._fetch_log_messages(resource=resource))
    resource.logger.error(msg)
    msg_store[resource] = msg
```

testplan/common/entity/base.py
Attakay78: I am maintaining the class hierarchy implementation.

testplan/common/entity/base.py
```diff
@@ -188,6 +188,17 @@ def all_status(self, target):
             for resource in self._resources
         )
 
+    def _record_resource_exception(self, message, resource, msg_store):
+        fetch_msg = "\n".join(resource.fetch_error_log())
+
+        msg = message.format(
+            resource.cfg.name,
+            traceback.format_exc(),
+            fetch_msg,
```
kn-ms, line 197: Make these keyword args, so it is visible from the format string what goes where. This is just a Nice To Have

testplan/common/entity/base.py
```diff
@@ -1413,9 +1413,9 @@ def post_stop(self):
 
     def fetch_error_log(self) -> List[str]:
         """
-        Fetch error message from the log files of resource.
+        Override this method to extract several lines of text from log files of resource.
```
kn-ms, line 1416: How about something like this which explain when this method would be used.

"Override this method in Driver subclasses to automatically add any useful logs into the report, in case of startup/shutdown exception."

doc/newsfragments/1831_new.driver_log_capture.rst
```diff
@@ -0,0 +1,2 @@
+Added :py:meth:`_record_resource_exception <testplan.common.entity.base.Environment._record_resource_exception>` to fetch error log messages from failed resources during Environment start or stop.
```
kn-ms, line 1: This is too technical. No need to mention the private helper, that is an implementation details.

doc/newsfragments/1831_new.driver_log_capture.rst
```diff
@@ -0,0 +1,2 @@
...
+Added :py:meth:`fetch_error_log <testplan.common.entity.base.Resource.fetch_error_log>` to fetch several lines of text from log files of resources.
```
kn-ms, line 2: Drivers now can attach some extra logs to the report in case of startup/shutdown failure, with implementing: :py:meth:`fetch_error_log <testplan.common.entity.base.Resource.fetch_error_log>`

doc/newsfragments/1831_new.driver_log_capture.rst
Attakay78: Alright, noted.

testplan/common/entity/base.py
Attakay78: Should be in Resource subclasses rather. This docs sounds good.

kn-ms (approved):

PR #907, opened 2023-03-01
Enabled port resolution with port aliases

nakjemmyMS:
TCPServer/Client does not support resolution of port alias. Further, this type of resolution may be needed across more drivers.

## Solution description
TCPServer and TCPClient have been enhanced to accept a port alias 'str' which will be resolved into a valid port `int`

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/driver/tcp/client.py
+++ b/testplan/testing/multitest/driver/tcp/client.py
@@ -3,11 +3,11 @@
 import socket
 from typing import Union, Tuple, Optional
 
-from schema import Use, Or
-
+from schema import Or
+from testplan.common.utils import networking
 from testplan.common.utils.timing import TimeoutException, TimeoutExceptionInfo
 from testplan.common.config import ConfigOption
-from testplan.common.utils.context import is_context, expand, ContextValue
+from testplan.common.utils.context import expand, ContextValue
 from testplan.common.utils.sockets import Client
 
 from ..base import Driver, DriverConfig
@@ -25,8 +25,8 @@ class TCPClientConfig(DriverConfig):
         Schema for options validation and assignment of default values.
         """
         return {
-            "host": Or(str, lambda x: is_context(x)),
-            "port": Or(Use(int), lambda x: is_context(x)),
+            "host": Or(str, ContextValue),
+            "port": Or(int, str, ContextValue),
             ConfigOption("interface", default=None): Or(None, tuple),
             ConfigOption("connect_at_start", default=True): bool,
         }
@@ -67,7 +67,7 @@ class TCPClient(Driver):
         self,
         name: str,
         host: Union[str, ContextValue],
-        port: Union[int, ContextValue],
+        port: Union[int, str, ContextValue],
         interface: Optional[Union[str, Tuple[str, int]]] = None,
         connect_at_start: bool = True,
         **options
@@ -167,7 +167,9 @@ class TCPClient(Driver):
         """Start the TCP client and optionally connect to host/post."""
         super(TCPClient, self).starting()
         self._server_host = expand(self.cfg.host, self.context)
-        self._server_port = expand(self.cfg.port, self.context, int)
+        self._server_port = networking.port_to_int(
+            expand(self.cfg.port, self.context)
+        )
         self._client = Client(host=self._server_host, port=self._server_port)
         if self.cfg.connect_at_start:
             self.connect()
--- a/testplan/testing/multitest/driver/tcp/server.py
+++ b/testplan/testing/multitest/driver/tcp/server.py
@@ -3,15 +3,15 @@ TCPServer driver classes.
 """
 
 import socket
-from typing import Union, Optional
+from typing import Optional, Union
 
-from schema import Use
-
+from schema import Or
 from testplan.common.config import ConfigOption
-from testplan.common.utils.context import ContextValue
+from testplan.common.utils import networking
+from testplan.common.utils.context import ContextValue, expand
 from testplan.common.utils.documentation_helper import emphasized
-from testplan.common.utils.timing import TimeoutException, TimeoutExceptionInfo
 from testplan.common.utils.sockets import Server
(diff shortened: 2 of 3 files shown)


testplan/testing/multitest/driver/tcp/client.py
```diff
@@ -167,7 +167,11 @@ def starting(self):
         """Start the TCP client and optionally connect to host/post."""
         super(TCPClient, self).starting()
         self._server_host = expand(self.cfg.host, self.context)
-        self._server_port = expand(self.cfg.port, self.context, int)
+        self._server_port = (
+            expand(self.cfg.port, self.context, int)
```
kn-ms, line 171: the expanded port might be a string as well, need to transform that

testplan/testing/multitest/driver/tcp/server.py
```diff
@@ -152,7 +153,14 @@ def receive(self, size=None, conn_idx=None, timeout=30):
     def starting(self):
         """Starts the TCP server."""
         super(TCPServer, self).starting()
-        self._server = Server(host=self.cfg.host, port=self.cfg.port)
+        self._server = Server(
+            host=self.cfg.host,
+            port=(
+                expand(self.cfg.port, self.context, int)
```
kn-ms, line 159: the expanded port might be a string as well, need to transform that

tests/unit/testplan/testing/multitest/driver/test_tcp_server.py
```diff
@@ -38,6 +39,46 @@ def tcp_client(tcp_server, runpath_module):
...
+        yield server
+
+
+@pytest.fixture(scope="module")
+def tcp_client_port_aliased(tcp_server_port_aliased, runpath_module):
+    """Start and yield a TCP client driver using port aliasing."""
+    client = TCPClient(
+        name="client_port_aliased",
+        host=tcp_server_port_aliased.host,
+        port=tcp_server_port_aliased.port,
```
kn-ms, line 74: here the port is for sure numeric so the new logic is not excercised

tests/unit/testplan/testing/multitest/driver/test_tcp_server.py
```diff
@@ -38,6 +39,46 @@ def tcp_client(tcp_server, runpath_module):
...
 
+@pytest.fixture(scope="module")
+def tcp_server_port_aliased(runpath_module):
+    """Start and yield a TCP server driver using port aliasing."""
+    env = Environment()
+
+    def get_mocked_port_name():
+        for port in range(2000, 65535):
+            try:
+                return socket.getservbyport(port)
```
kn-ms, line 50: nothing guaranties that on the testing hosts there are port names configured. better to patch port_to_int to return what you expect

kn-ms (dismissed):

testplan/testing/multitest/driver/tcp/client.py
nakjemmyMS: Thanks for the feedback @kn-ms 
The third param will be used as the constructor. Please check the implementation of the expand

testplan/testing/multitest/driver/tcp/server.py
nakjemmyMS: The third param will be used as the constructor similar to the tcp_client commend

tests/unit/testplan/testing/multitest/driver/test_tcp_server.py
nakjemmyMS: Because the `tcp_client_port_aliased` uses the `tcp_server_port_aliased `fixture. there is a dependency (chain) so it is the actual test `test_send_receive_with_port_aliasing` which asserts it.

tests/unit/testplan/testing/multitest/driver/test_tcp_server.py
nakjemmyMS: Yeah 😊. It looks a bit hacky. I was thinking of how to create a port alias at runtime but that would demand root access so I found this new way which worked for multiple environments given a range of [**un-privileged ports.**](https://www.w3.org/Daemon/User/Installation/PrivilegedPorts.html) I would be glad if you could help me get a better way around it.

tests/unit/testplan/testing/multitest/driver/test_tcp_server.py
kn-ms: tcp_server_port_aliased.port always return an int as the port is already resolved during startup time, so the client will do not need to run the resolution.

tests/unit/testplan/testing/multitest/driver/test_tcp_server.py
kn-ms: If I were you, I would write a helper to resolve port (being it int, context, value or string) use that in both client and server, and write a pure unittest for that function, with patching the port_to_int (https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch) to return the desired port for the portname the test test is using. I would not do any server startup, and magic to figure out if named ports exists, not talking about the possibility that the named port you pick might be already taken and your test is just failing. This is a nice way of adding flaky tests.

M6AI:
So just to summarize the changes here and there:

- no need for dco
- a new utility for the "ternary" is added (perhaps to the same module as port_to_int is living)
- a unittest for that guy is added by patching the return value of port_to_int

testplan/testing/multitest/driver/tcp/client.py
nakjemmyMS: Okay got it now. Thanks for the clarification

tests/unit/testplan/testing/multitest/driver/test_tcp_server.py
nakjemmyMS: Thanks for the feedback. I will do the necessary refinements.

M6AI (dismissed):

M6AI (approved):

PR #908, opened 2023-03-02
Configurable driver log collector

rnemes:
Need a way to customize driver log collecting.

## Solution description
Implemented a callable class that can be configured with certain filename patterns to collect.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/.github/workflows/test_pr.yml
+++ b/.github/workflows/test_pr.yml
@@ -76,6 +76,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         python-version: [3.7, 3.8, '3.10', 3.11]
+      fail-fast: false
 
     steps:
       - run: echo "BUILD ${{ matrix.os }} ${{ matrix.python-version }} ${{needs.is_source_changed.outputs.is_source_changed}}"
--- a/doc/en/download/Best Practice.rst
+++ b/doc/en/download/Best Practice.rst
@@ -20,6 +20,7 @@ Example:
     .. code-block:: python
 
       from testplan.common.utils import helper
+
       log_collector = helper.DriverLogCollector(
           name="custom_log_collector",
           file_pattern=["*.log"],
--- a/doc/en/download/Best Practice.rst
+++ b/doc/en/download/Best Practice.rst
@@ -6,11 +6,29 @@ Best Practice
 .. _example_common_utilities:
 
 Helper Utilities
-----------------
+================
 
 Testplan provides helper functions and a predefined testsuite that make it
 easy for the user to add common testplan execution infomation - such as
 env var, pwd, log file, driver metadata - to the test report.
+
+DriverLogCollector
+------------------
+You can specify your custom log collector using :py:class:`~testplan.common.utils.helper.DriverLogCollector`.
+It will attach the specified log files to the report for each driver.
+Example:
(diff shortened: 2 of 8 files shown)


.github/workflows/test_pr.yml
```diff
@@ -76,6 +76,7 @@ jobs:
       matrix:
         os: [ubuntu-latest, windows-latest]
         python-version: [3.7, 3.8, '3.10', 3.11]
+      fail-fast: false
```
M6AI, line 79: Is this an agreed change within scope?

doc/en/download/Best Practice.rst
```diff
@@ -6,12 +6,30 @@ Best Practice
...
 env var, pwd, log file, driver metadata - to the test report.
 
+DriverLogCollector
+------------------
+You can specify your custom log collector using :py:class:`~testplan.common.utils.helper.DriverLogCollector`.
+It will attach the specified log files to the report for each driver.
+Example:
+    .. code-block:: python
+
+      from testplan.common.utils import helper
```
M6AI, line 22: Add one or two empty line to be inline with how we format code blocks of similar nature.

testplan/common/utils/helper.py
```diff
@@ -35,6 +38,58 @@
...
+    :param recursive: Recursively traverse sub-directories and attach
+        all files, default is to only attach files in top directory.
+    :param failure_only: Only collect files on failure.
+    """
+
+    def __init__(
+        self,
+        name: str = "DriverLogCollector",
+        description: str = "logs",
+        ignore: list = None,
```
M6AI, line 61: Can we specify this more, like `List[str]` using `List` from `typing` module. Same goes for `file_pattern` parameter.

testplan/common/utils/interface.py
```diff
@@ -1,144 +1,7 @@
...
-                ", **{0}".format(self.keywords) if self.keywords else "",
-            ]
-        )
-
-        return "{0}({1})".format(self.name, args)
-
-    def __repr__(self):
-        return "<{}> - {}".format(self.__class__.__name__, self.__str__())
+from typing import Union, Optional
+from inspect import signature
```
M6AI, line 4: Change order of these imports please (will explain offline).

testplan/common/utils/interface.py
```diff
@@ -157,22 +20,20 @@ class MethodSignatureMismatch(Exception):
...
+) -> Optional[bool]:
     """
     Checks if the given function's signature matches the given list of args
 
     :param func: function whose signature to check
-    :type func: ``callable``
     :param args_list: list of arg names to match as signature
-    :type args_list: ``list`` of ``str``
 
     :return: ``None`` or ``True``
```
M6AI, line 32: No need to write the return values, those are coming from `Optional[bool]` already. Also I think the thing here is actually just `bool` and the function's docstring needs a `raises` directive. Because it cannot come back void, it either raises or returns a bool.

M6AI (changes requested):

.github/workflows/test_pr.yml
rnemes: Yes, Krisz gave it a green light

M6AI (approved):

PR #909, opened 2023-03-14
Change assertion time format to 24 hour format.

rnemes:
Change assertion time format on WebUI to 24 hour format.

## Solution description
When "time information" is enabled on the web report it shows a 12 hour format without AM/PM, so it is misleading.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


zhenyu-ms (dismissed):

Pyifan (dismissed):

M6AI (changes requested):
Change looks sufficient, but as discussed offline, we can add a test as the snapshot captures the format if the display is turned on.

M6AI (approved):

kn-ms (approved):

PR #917, opened 2023-03-23
Revamped logging levels in exporters in preparation to remove exporter_info

kn-ms:
exporter_info will be retired soon need to use new user_info loglevel

## Solution description
changed all occurences,
added some typehints

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [X] MS info leakage check

### DIFF
--- a/testplan/exporters/testing/http/__init__.py
+++ b/testplan/exporters/testing/http/__init__.py
@@ -4,7 +4,7 @@ must be able to handle POST request and receive data in JSON format.
 """
 
 import json
-from typing import Any, Tuple, Union
+from typing import Any, Tuple, Union, Optional
 
 import requests
 from schema import Or, And, Use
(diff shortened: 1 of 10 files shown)


testplan/exporters/testing/http/__init__.py
```diff
@@ -92,16 +93,15 @@ def _upload_report(
 
         return response, errmsg
 
-    def export(self, source) -> Union[None, str]:
-        """TODO"""
+    def export(self, source: TestReport) -> Union[None, str]:
```
M6AI, line 96: Do we have preference atm between Union[None, ...] and Optional[...]? Just dropping it in, reading up the PEPs there is no clear recommendation and both will move to, eventually, to the newer type annotation format with operators.

M6AI (approved):

testplan/exporters/testing/http/__init__.py
kn-ms: TBH, I just copied around whatever was there already. But I like the Optional more, lets go with that.

M6AI (approved):

PR #918, opened 2023-03-23
Fix/adjust log level for Kafka, Zookeeper & various test wrappers

zhenyu-ms:
As title goes.

## Solution description
Information helpful to user should be either at user_info or info level.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/testing/multitest/driver/kafka.py
```diff
@@ -82,14 +82,24 @@ def __init__(
 
     @emphasized
     @property
-    def host(self) -> Optional[str]:
+    def host(self) -> str:
         """Host to bind to."""
+        if self.status != self.STATUS.STARTED:
```
kn-ms, line 87: host and port might be used in config file instantiation from the template, I do not think we need to require these only available when started

testplan/testing/multitest/driver/zookeeper.py
```diff
@@ -87,23 +87,33 @@ def __init__(
 
     @emphasized
     @property
-    def host(self) -> Optional[str]:
+    def host(self) -> str:
         """Host to bind to."""
+        if self.status != self.STATUS.STARTED:
```
kn-ms, line 92: Same as with Kafka

testplan/testing/multitest/driver/kafka.py
zhenyu-ms: sure thing, maybe I've lost my mind when writing these lines 😵

testplan/testing/cpp/cppunit.py
```diff
@@ -208,7 +210,7 @@ def test_command_filter(self, testsuite_pattern, testcase_pattern):
             )
 
         if testcase_pattern not in ("*", self._VERIFICATION_TESTCASE_NAME):
-            self.logger.debug(
+            self.logger.info(
```
M6AI, line 213: Wondering if this should be user_info due to the fact it is not about what is being executed one by one, but more of a summary statement that "look, we are going to run a whole test suite now". I will let others opine too.

testplan/testing/cpp/hobbestest.py
```diff
@@ -182,7 +181,7 @@ def test_command_filter(self, testsuite_pattern="*", testcase_pattern="*"):
 
         # At the beginning no testcase exists in test suite
         if testcase_pattern not in ("*", self._VERIFICATION_TESTCASE_NAME):
-            self.logger.debug(
+            self.logger.info(
```
M6AI, line 184: Same story.

testplan/testing/junit.py
```diff
@@ -197,7 +199,7 @@ def test_command_filter(self, testsuite_pattern, testcase_pattern):
             )
 
         if testcase_pattern not in ("*", self._VERIFICATION_TESTCASE_NAME):
-            self.logger.debug(
+            self.logger.info(
```
M6AI, line 202: One more.

testplan/testing/multitest/driver/kafka.py
```diff
@@ -82,8 +82,12 @@ def __init__(
 
     @emphasized
     @property
-    def host(self) -> Optional[str]:
+    def host(self) -> str:
         """Host to bind to."""
+        if self._host is None:
+            raise RuntimeError(
+                "Host not resolved yet, shouldn't be accessed now."
+            )
```
M6AI, line 90: Hm, cannot we just return the None? I think this will immediately break at least metadata extraction unless the exception handling is there. I would suggest making log dependent if that is why we need it.

zhenyu-ms: but I think metadata extraction will only be triggered after driver has at least finished with pre_start step? otherwise there won't be any metadata to be extracted I think

M6AI: It works with ContextValue as well. But it is true we are not using it other than Proxima after_start.

M6AI (approved):

PR #928, opened 2023-04-03
Enhance LogMatcher to support log rotation

Attakay78:
Enhance LogMatcher to support log file rotation.

### DIFF
--- a/testplan/common/utils/match.py
+++ b/testplan/common/utils/match.py
@@ -4,6 +4,7 @@ Module of utility types and functions that perform matching.
 import os
 import re
 import time
+from abc import ABCMeta, abstractmethod
 from typing import Dict, List, Match, Optional, Pattern, Tuple, Union
 
 from . import logger, timing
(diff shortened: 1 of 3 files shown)


testplan/common/utils/match.py
```diff
@@ -57,6 +58,211 @@ def match_regexps_in_file(
...
+    def validate_marks(
+        mark1: LogFilePosition, mark2: LogFilePosition, log_path: str = None
+    ) -> bool:
+        """
+        Checks the position validity of mark1 and mark2.
+        Returns True if mark1 has higher position than mark2,
+        meaning mark2 is been searched before mark1. Else returns
+        False.
+        """
+        if (mark1.inode == mark2.inode) and (mark1.position >= mark2.position):
```
kn-ms, line 260: In this strategy inode's should not differ, so that even can be an assertion.

testplan/common/utils/match.py
```diff
@@ -131,6 +342,8 @@ def match(
         start_time = time.time()
         end_time = start_time + timeout
         read_mode = "rb"
+        continue_search = True
+        current_position = self.position.position
```
kn-ms, line 346: You should not look into LogPosition in the LogMatcher code, it should not even know the details of that other than it represent a position by the stratgy

testplan/common/utils/match.py
```diff
@@ -57,6 +58,211 @@ def match_regexps_in_file(
...
+        """
+        Implementation for opening and returning log handles with the
+        given file position.
+        """
+        raise NotImplementedError
+
+    @staticmethod
+    @abstractmethod
+    def get_position(
+        position: int, file_handle: TextIO = None
```
kn-ms, line 106: position should come from the file handle position not outside

testplan/common/utils/match.py
```diff
@@ -246,18 +481,50 @@ def match_between(self, regex, mark1, mark2):
...
-                    break
+        try:
+            start_position = self.marks[mark1]
+            end_position = self.marks[mark2]
 
-        return match
+            log_handle = self.strategy.open_logfile_in_position(
+                position=start_position, log_path=self.log_path, mode=read_mode
+            )
+            current_inode, current_position = (
```
kn-ms, line 491: Do not rely what is in LogPositions, just comapare psotions

testplan/common/utils/match.py
```diff
@@ -296,22 +563,66 @@ def get_between(self, mark1=None, mark2=None):
...
+            log_handle = self.strategy.open_logfile_in_position(
+                position=start_position, log_path=self.log_path, mode="r"
+            )
+            current_inode, current_position = (
+                start_position.inode,
+                start_position.position,
+            )
+            while True:
+                with open(log_handle.name, "r") as log:
+                    while (end_position.inode != current_inode) or (
```
kn-ms, line 603: Just use comparable positions not the internals of it

kn-ms (changes requested):

testplan/common/utils/match.py
```diff
@@ -296,22 +536,49 @@ def get_between(self, mark1=None, mark2=None):
...
+
+            while self.strategy.end_position_reached(
+                end_position=end_position, file_handle=log_handle
+            ):
+                line = log_handle.readline()
+
+                if line:
+                    lines_between.append(line)
+                else:
+                    log_handle = self.strategy.get_next_file(
```
kn-ms, line 575: old log handle need to be closed

kn-ms (changes requested):

testplan/common/utils/match.py
```diff
@@ -57,6 +58,200 @@ def match_regexps_in_file(
...
+        """
+        Checks the position validity of mark1 and mark2.
+        Returns True if mark1 has higher position than mark2,
+        meaning mark2 is been searched before mark1. Else returns
+        False
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def end_position_reached(
```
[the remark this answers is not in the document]
Attakay78: So i'm actually searching from a start position to the end so I want to know if the end position has been reached.

testplan/common/utils/match.py
Attakay78: For the NoLogRotationStrategy I will be returning the same handle so I am not closing it there. But in the other strategy types that require rotation, the old file handle is closed in the **get_next_file** function before returning the new log handle.

testplan/common/utils/match.py
```diff
@@ -57,6 +58,203 @@ def match_regexps_in_file(
...
+        """Returns the start position for the current log file."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def validate_marks_order(
+        self, mark1: LogFilePosition, mark2: LogFilePosition
+    ) -> bool:
+        """
+        Checks the position validity of mark1 and mark2.
+        Returns True if mark1 has higher position than mark2,
```
kn-ms, line 142: not intuitive if it want mark1 is after mark2

testplan/common/utils/match.py
```diff
@@ -57,6 +58,203 @@ def match_regexps_in_file(
...
+        False
+        """
+        raise NotImplementedError
+
+    @abstractmethod
+    def position_passed(
+        self, end_position: LogFilePosition, file_handle: TextIO
+    ) -> bool:
+        """
+        Checks if the log search has reached the end of the file.
```
kn-ms, line 153: call the parameter just position, and update the docstring as this supposed to check if it is passing the position not end of file

testplan/common/utils/match.py
```diff
@@ -57,6 +58,203 @@ def match_regexps_in_file(
...
+        """
+        return mark1.position >= mark2.position
+
+    def position_passed(
+        self, end_position: LogFilePosition, file_handle: TextIO
+    ) -> bool:
+        """
+        Checks if the log search has reached the end of the file.
+        """
+        return end_position.position > file_handle.tell()
```
kn-ms, line 255: should not this be true if the file_handle is over the provided position?

testplan/common/utils/match.py
```diff
@@ -83,15 +290,13 @@ def seek(self, mark: Optional[str] = None):
         :param mark: Name of the mark.
         """
         if mark is None:
-            self.position = 0
+            self.position = self.strategy.seek()
         else:
-            self.position = self.marks[mark]
+            self.position = self.strategy.seek(mark=self.marks[mark])
```
kn-ms, line 295: here the position you get from the mark should be good, no need to involve the strategy.

testplan/common/utils/match.py
```diff
@@ -83,15 +290,13 @@ def seek(self, mark: Optional[str] = None):
         :param mark: Name of the mark.
         """
         if mark is None:
-            self.position = 0
+            self.position = self.strategy.seek()
```
kn-ms, line 293: use get_start_of_file_position

testplan/common/utils/match.py
```diff
@@ -57,6 +58,203 @@ def match_regexps_in_file(
...
+
+    @abstractmethod
+    def get_next_file(
+        self, file_handle: TextIO, mode: Optional[str] = None
+    ) -> Tuple[TextIO, bool]:
+        """Returns the next file handle and a truthy value whether a new file was returned."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def seek_eof(self) -> LogFilePosition:
```
kn-ms, line 114: get_end_of_file_position is the same, remove seek and left the other

testplan/common/utils/match.py
```diff
@@ -57,6 +58,203 @@ def match_regexps_in_file(
...
+        """Returns the end file position of the current log file."""
+        raise NotImplementedError
+
+    @abstractmethod
+    def seek(self, mark: Optional[LogFilePosition] = None) -> LogFilePosition:
```
kn-ms, line 119: No need for seek, as if mark is None one can just use get_start_of_file_position, if there is a mark this would always return that

testplan/common/utils/match.py
```diff
@@ -246,18 +468,41 @@ def match_between(self, regex, mark1, mark2):
...
+                end_position=end_position, file_handle=log_handle
+            ):
+                line = log_handle.readline()
+
+                if line:
+                    match = regex.match(line)
+                    if match:
+                        break
+                else:
+                    log_handle, _ = self.strategy.get_next_file(
```
kn-ms, line 496: previous logfile is not closed

testplan/common/utils/match.py
```diff
@@ -139,23 +333,43 @@ def match(
...
                         break
-                elif timeout > 0:
-                    time.sleep(LOG_MATCHER_INTERVAL)
                 else:
-                    break
 
-            self.position = log.tell()
+                    previous_log_handle = log_handle
+
+                    log_handle, is_new_file = self.strategy.get_next_file(
```
Pyifan, line 354: Can we absort the get_next_file thing into strategy?

I think the strategy layer shall serve as a virtual and continuous log file, it provides readline() method that LogMatcher can call without knowing we are stepping into another rotated log segment.

Maybe Strategy is not the right name - i'm really talking about an abstraction of the rotating log.

kn-ms: We can call it LogStream or LogSource than, what the interface would be

- `readline() -> Union[str|None]`
- `position() -> Position`  Position is an opaque that can be just used for comparision, and for seek
- `seek(p: Position)` 
- `sof() -> Position`
- `eof() -> Position`
- `compare(position1: Position, position2: Position) -> int` classic compare meaning
- `close()` maybe for optimalization from matcher to close the stream when it does not need it for a while.

I like that we have a single component touching the file handles. 

The question is what is the lifetime of the object, 
 - Is it a member of the Matcher, then is it hold a file handle to the logfile at actual position? When does it close it, ? This would remove the need to store the position in the Matcher
 - Is it recreated and seeked for every match operation as the file is opened today.

Pyifan: Combined with the internal comment and some offline discussion, if a file name pattern in regex + inode create time for putting things in order will suffice for most of the use cases with log rotation -- I don't think we need to introduce new public class here for user to extend, but would like to make the out-of-box usage really simple and straight-forward and trim down the lines of code we need to maintain.

I will suggest a usage like following:

log_matcher = LogMatcher(log_path="path/to/log/file", rotation_pattern="engine1.*log")

If rotation_pattern is specified, we can create a LogStream inside LogMatcher. I think the LogStream instance can be long-living in LogMatcher, but then the underlying log file, maybe open/close every time match() is called (this is what we do today with non-rotating logfile).

kn-ms: I do not like the bit where we leave out the potential for extensibility configuring LogMatcher with something LogStream could make this much more useful. Boxing this into pattern and directory would make this again a breaking change when a new need is coming.

Pyifan: Do you have any potential use case in mind, that would go beyond the file name pattern in regex?

Do you prefer something like this:

log_matcher = LogMatcher(log_path="path/to/log/file", log_rotation_strategy=RegexStrategy(pattern="engine1.*log")

where RegexStrategy is provided out- of-box, and will only be used to identify the group of rotated logs.

Pyifan: I think maybe a log_path="path/to/log/engine*log" might be a good enough heuristic for now. We can introduce a strategy parameter in future if different requirement come up, then we don't break anything.

testplan/common/utils/match.py
```diff
@@ -296,22 +538,54 @@ def get_between(self, mark1=None, mark2=None):
...
-        if mark1 is not None and mark2 is not None:
-            if self.marks[mark1] >= self.marks[mark2]:
-                raise ValueError(
-                    'Mark "{}" must be present before mark "{}"'.format(
-                        mark1, mark2
+        try:
+            log_handle = None
+
+            if mark1 is not None and mark2 is not None:
+                if self.strategy.invalid_position_order(
```
Attakay78, line 545: @kn-ms I am thinking the invalid_position_order should be maintained instead of the compare since we might have multiple files and will not be doing the comparison only on the position but on the files inode as well.

testplan/common/utils/match.py
```diff
@@ -296,22 +538,54 @@ def get_between(self, mark1=None, mark2=None):
...
+                self.marks[mark2]
+                if mark2 is not None
+                else self.strategy.get_end_of_file_position()
+            )
+
+            log_handle = self.strategy.open_logfile_in_position(
+                position=start_position, mode="r"
+            )
+
+            while self.strategy.position_not_passed(
```
Attakay78, line 570: @kn-ms Is there a need to remove this function? Because I have maintained it just that we will not be passing any file_handle.

PR #931, opened 2023-04-04
Refactor!/turn status check into predicates

zhenyu-ms:
We don't have checking started or stopped predicates in the current driver interface, which is not good for simultaneous driver starting.

## Solution description
Turn `started_check` & `stopped_check` into predicates, which actually conform better with their names.

This is a breaking change for end-user with custom drivers.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/common/utils/match.py
+++ b/testplan/common/utils/match.py
@@ -121,8 +121,9 @@ class LogMatcher(logger.Loggable):
         :param regex: Regex string or compiled regular expression
             (``re.compile``)
         :param timeout: Timeout in seconds to wait for matching process,
-            0 means should not wait and return whatever matched on initial
-            scan, defaults to 5 seconds
+            0 means matching till EOF and not waiting for new lines, any
+            value greater than 0 means doing matching up to such seconds,
+            defaults to 5 seconds
         :param raise_on_timeout: To raise TimeoutException or not
         :return: The regex match or None if no match is found
         """
@@ -141,16 +142,16 @@ class LogMatcher(logger.Loggable):
         with open(self.log_path, read_mode) as log:
             log.seek(self.position)
 
-            while match is None:
+            while True:
+                if timeout > 0 and time.time() > end_time:
+                    break
                 line = log.readline()
                 if line:
                     match = regex.match(line)
                     if match:
                         break
-                elif timeout:
+                elif timeout > 0:
                     time.sleep(LOG_MATCHER_INTERVAL)
-                    if time.time() > end_time:
-                        break
                 else:
                     break
 
--- a/tests/unit/testplan/common/utils/test_match.py
+++ b/tests/unit/testplan/common/utils/test_match.py
@@ -8,7 +8,7 @@ from testplan.common.utils import timing
 from testplan.common.utils.match import LogMatcher, match_regexps_in_file
 
 
-@pytest.yield_fixture(scope="module")
+@pytest.fixture(scope="module")
 def basic_logfile():
     """Write a very small logfile for basic functional testing."""
     log_lines = ["first\n", "second\n", "third\n", "fourth\n", "fifth\n"]
@@ -22,7 +22,7 @@ def basic_logfile():
     os.remove(filepath)
 
 
-@pytest.yield_fixture(scope="module")
+@pytest.fixture(scope="module")
 def large_logfile():
     """Write a larger logfile for more realistic performance testing."""
     with tempfile.NamedTemporaryFile("w", delete=False) as logfile:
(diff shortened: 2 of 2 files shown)


testplan/common/entity/base.py
```diff
@@ -1380,7 +1383,7 @@ def fetch_error_log(self) -> List[str]:
         """
         return []
 
-    def _wait_started(self, timeout=None):
+    def _wait_started(self, timeout: Optional[float] = None):
```
Pyifan, line 1386: change docstring accordingly

testplan/testing/multitest/driver/app.py
```diff
@@ -323,28 +323,14 @@ def starting(self) -> None:
...
-                    f"{self} has unexpectedly stopped with: {proc_result}"
-                )
-            return extract_values_result
-
-        wait(
-            ensure_app_running_while_extracting_values,
-            timeout,
-            raise_on_timeout=True,
-        )
+    def started_check(self) -> ActionResult:
```
Pyifan, line 326: Shall we still add docstring?

Pyifan (commented):
Looks good and we need a news fragment as this might be a breaking change for user.

Pyifan (approved):

doc/en/drivers_custom.rst
```diff
@@ -20,7 +20,7 @@ a binary as a sub-process.
 
 Here is a custom driver inherits the built-in
```
M6AI, line 21:
```suggestion
Here is a custom driver that inherits the built-in
```
While we are at it, maybe correct this typo from original code.

testplan/common/utils/match.py
```diff
@@ -2,16 +2,18 @@
...
-from . import logger
+from . import logger, timing
 
 LOG_MATCHER_INTERVAL = 0.25
 
 
-def match_regexps_in_file(logpath, log_extracts):
+def match_regexps_in_file(
+    logpath: os.PathLike, log_extracts: List[Pattern]
+) -> Tuple[bool, dict, List[Pattern]]:
```
M6AI, line 16: Can we be more specific what sort of `dict` is coming back? If types are complex just go with some umbrella, but we like to do `Dict[..., ...]`.

testplan/common/utils/match.py
```diff
@@ -147,27 +151,27 @@ def match(self, regex, timeout=5):
                     match = regex.match(line)
                     if match:
                         break
-                else:
+                elif timeout:
```
M6AI, line 154: I seem to be missing where `timeout` is being used other than truthy/falsy, yet in the `raise` below it seems to be made explicit we use it. Can you please pinpoint me?

testplan/common/utils/match.py
```diff
@@ -176,49 +180,50 @@ def not_match(self, regex, timeout=5):
...
         try:
-            match = self.match(regex, timeout)
+            self.match(regex, timeout, raise_on_timeout=True)
         except timing.TimeoutException:
             pass
-
-        if match:
+        else:
             raise Exception(
                 "Unexpected match[{}] found in {}s".format(
```
M6AI, line 194: This can be an f-string.

testplan/common/utils/match.py
zhenyu-ms: Ah it's indirectly used in `end_time`, which is used on line 156.

testplan/testing/multitest/driver/fix/client.py
```diff
@@ -170,22 +170,20 @@ def sendersub(self) -> str:
...
-                    "FixClient %s not able to connect - %s", self, exc
-                )
-            else:
-                break
+    def started_check(self) -> ActionResult:
+        try:
+            self.reconnect()
+        except Exception as exc:
+            self.logger.debug(
+                "FixClient %s not able to connect - %s", self, exc
```
zhenyu-ms, line 178:
```suggestion
                "%s not able to connect - %s", self, exc
```

testplan/common/utils/match.py
M6AI: Nice, I see, just had to expand the lines from front. Thanks, it looks good this way.

testplan/common/utils/match.py
zhenyu-ms: Sure thing, it actually contains extracted values with name of key and type of value specified by regular expression.

zhenyu-ms: oh actually it's just string, will fix

M6AI (dismissed):
Looks good to me after offline walkthrough.

testplan/common/utils/match.py
```diff
@@ -147,27 +147,29 @@ def match(self, regex, timeout=5):
                     match = regex.match(line)
                     if match:
                         break
-                else:
+                elif timeout:
                     time.sleep(LOG_MATCHER_INTERVAL)
                     if time.time() > end_time:
                         break
+                else:
+                    break
```
Pyifan, line 155: refactor as we discussed

Pyifan (approved):

PR #933, opened 2023-04-07
tls support for FixServer and FixClient

kn-ms:

## Solution description

## Checklist:
- [ ] Test
- [X] Documentation (API)
- [X] MS info leakage check


doc/newsfragments/2465_new.fix_tls.rst
```diff
@@ -0,0 +1,2 @@
+Testplan now has support for usr1 and usr2 signal handlers. Triggering the usr1 signal drops Testplan to pdb mode
```
Pyifan, line 1: need to update news frag.

Pyifan (changes requested):
Looks great, just need to update news frag i guess.

testplan/common/utils/sockets/tls.py
```diff
@@ -0,0 +1,64 @@
...
+        self,
+        key: Union[PathLike, str],
+        cert: Union[PathLike, str],
+        cacert: OPTIONAL_PATH,
+    ):
+        self.key = Path(key)
+        self.cert = Path(cert)
+        self.cacert = Path(cacert)
+
+    def get_context(self, purpose: ssl.Purpose):
```
M6AI, line 59: Return annotation missing.

testplan/common/utils/sockets/tls.py
```diff
@@ -0,0 +1,64 @@
...
+
+
+class TLSConfig(ABC):
+    """
+    Defines the Protocol a TLSConfig need to have
+    """
+
+    @abstractmethod
+    def get_context(self, purpose: ssl.Purpose) -> ssl.SSLContext:
+        """
```
M6AI, line 15: Missing description in docstring.

examples/Transports/FIX/over_one_session.py
```diff
@@ -28,6 +29,7 @@
 from testplan.common.utils.context import context
 from testplan.testing.multitest import MultiTest, testsuite, testcase
 from testplan.testing.multitest.driver.fix import FixServer, FixClient
+from testplan.common.utils.sockets.tls import SimpleTLSConfig
 
 CODEC = Codec(spec=FixSpec(SPEC_FILE))
```
M6AI, line 34: Independent of current scope, but why do we use this approach here? The Codec from pyfixmsg work without a specification even though it will have issues with groups and encoded tags.

testplan/common/utils/sockets/tls.py
```diff
@@ -0,0 +1,64 @@
...
+
+        :param purpose: Either host or client certificate
+        :return: should return the configured SSLContext
+        """
+        ...
+
+
+OPTIONAL_PATH = Optional[Union[PathLike, str]]
```
M6AI, line 23: I think this can go after imports as PEP suggests.

examples/Transports/FIX/over_one_session.py
Pyifan: I think this is to demonstrate to user about how to provide the fix spec.

Pyifan (approved):

PR #938, opened 2023-05-03
Change env passed to setup/teardown/after_stop etc from Environment t…

Attakay78:
Change env passed to setup/teardown/after_stop etc from Environment to RuntimeEnvironment.


kn-ms (commented):
Generally looking ok, I'm not sure though if we should represent these functions as testcases in the runtime info. Should this be different types? @Pyifan

Pyifan:
> Generally looking ok, I'm not sure though if we should represent these functions as testcases in the runtime info. Should this be different types? @Pyifan

Yea, I think it's fine.

kn-ms (approved):

PR #944, opened 2023-05-10
Bug/TextAttachment error handling

rnemes:
In special cases (eg. attachment is not available on the remote server) the TextAttachment gets an Object as an error and crashes the UI.

## Solution description
Store the correct error message based on the response type instead of the whole Object in errorHandler.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check


M6AI (approved):

kn-ms (changes requested):
Please add a newsfragment

kn-ms (approved):

PR #945, opened 2023-05-12
TreeView now scroll selection into view.

kn-ms:
- if selection is not opened it open the tree to there
- revamped layout of the major components on screen, now use flexbox instead of absolute positions.

## Solution description

## Checklist:
- [X] Test
- [ ] Documentation (API)
- [X] MS info leakage check


doc/newsfragments/2507_new.selection_sync.rst
```diff
@@ -0,0 +1 @@
+Selection in the breadcrumb bar make the selection scroll into view in the treeview as well. When opening an url with selection, the selected item gets visible on the treeview.
```
M6AI, line 1:
```suggestion
Selection in the breadcrumb bar makes the selection scroll into view in the tree navigation as well. When opening an URL with selection, the selected item is reflected on the tree view navigation too.
```

M6AI (changes requested):
QQ: why the styling changes? I see CSS being touched in places, would like to know if they have to do with the functionality for some reason or if it just looks better.

M6AI (approved):

PR #946, opened 2023-05-12
remove example certificates from the repo

kn-ms:

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [X] MS info leakage check


M6AI (commented):
Do we need a newsfragment for this change?

M6AI (approved):

kn-ms:
> Do we need a newsfragment for this change?

I do not think we need it, this affect only an example, and the readme describe what need to be done. The example need some manual work from the user anyway, It was never, self contained, just run it and you are ok type of example.

PR #947, opened 2023-05-12
Added error boundary to WebUI

rnemes:
Wrapping main React components in Error Boundary to prevent unexpected UI crash.

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check


doc/newsfragments/2513_new.added_errorBoundary.rst
```diff
@@ -0,0 +1 @@
+Wrapping main webUI components in Error Boundary to prevent unexpected UI crash.
```
M6AI, line 1: Can you please give an example of such a crash (say most commonly observed by you) so that users can relate?

M6AI: Also use either "error boundary" if referencing the concept or "ErrorBoundary" if it is the React component referenced.

testplan/web_ui/testing/src/Common/ErrorBoundary.js
```diff
@@ -0,0 +1,37 @@
...
+    this.state = { error: null };
+  };
+
+  static getDerivedStateFromError(error) {
+    return { error: error };
+  }
+
+  componentDidCatch(error, info) {
+    this.logErrorToServices(error, info.componentStack);
+  }
```
M6AI, line 15: Given I am not an expert here: basically the getDerivedStateFromError and componentDidCatch methods ensure the capture of the error updates the error field of the state, correct? I was looking at this for getting a sense https://legacy.reactjs.org/docs/error-boundaries.html but did not dig into it much.

testplan/web_ui/testing/src/index.js
```diff
@@ -17,10 +18,21 @@ import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
...
+          </ErrorBoundary>
+        )}
+      />
       <Route
         path="/interactive/:uid?/:selection*"
-        component={InteractiveReport}
+        render={({match}) => (
+          <ErrorBoundary>
+            <InteractiveReport match={match}/>
+          </ErrorBoundary>
```
M6AI, line 34: QQ: isn't it enough to wrap the BatchReport and InteractiveReport? If error already leads out of the component, my naive guess is that any lower level error would propagate up to this very point and these two ErrorBoundary instances would take care of everything.

M6AI (commented):
Left a few questions when you have time.

doc/newsfragments/2513_new.added_errorBoundary.rst
rnemes: I am referring to the name of the concept and tried to emphasize it, hence the capitals. Going to rephrase then anyway.

testplan/web_ui/testing/src/Common/ErrorBoundary.js
rnemes: > If you define static getDerivedStateFromError, React will call it when a child component (including distant children) throws an error during rendering. This lets you display an error message instead of clearing the UI.
> 
> Typically, it is used together with [componentDidCatch](https://react.dev/reference/react/Component#componentDidCatch) which lets you send the error report to some analytics service. A component with these methods is called an error boundary.
Source: https://react.dev/reference/react/Component#static-getderivedstatefromerror

testplan/web_ui/testing/src/index.js
rnemes: Already discussed offline, but putting here as well.
We want to catch render errors in the lowest component level possible (where it makes sense) to keep the functionality of the rest of the page. If we only wrap the BatchReport for example, a render failure in the Assertion level will prevent the user to access every other AssertionGroups.

testplan/web_ui/testing/src/Report/InteractiveReport.js
```diff
@@ -613,57 +614,63 @@ class InteractiveReport extends BaseReport {
...
+              <AbortButton
+                key="abort-button"
+                aborting={this.state.aborting}
+                abortCbk={this.abortTestplan}
+              />,
+              <SaveButton key="save-button" />,
+            ]}
+          />
+        </ErrorBoundary>
+        <ErrorBoundary>
```
kn-ms, line 651: if we do not have Nav I think it is enough to catch at top most, maybe the same for the toolbar.

PR #949, opened 2023-05-16
Feature/interactive filter suite case

M6AI:
Interactive mode does not support running only the filtered report. This change is to enable selective running with respect to the filter expression.

## Solution description
Previously, the test endpoints consumed UIDs to identify which entry to run with all of its children. Current update changes how the JSON body of the PUT request is used to construct all testsuites and testcases that need to be run, then intersect this container with the test context of the entry.

## Checklist:
- [x] Test
- [x] Documentation (API)
- [ ] MS info leakage check


testplan/report/testing/base.py
```diff
@@ -320,12 +274,27 @@ def runtime_status(self):
...
-        """Set the runtime_status of all child entries."""
+    def runtime_status(self, new_status: str) -> None:
+        """
+        Sets the runtime status of all child entries.
+        """
         for entry in self:
             entry.runtime_status = new_status
         self._runtime_status = new_status
 
+    def set_runtime_status_filtered(
```
M6AI, line 285: This serves as an alternative setter, may replace entirely the existing one, but the behavior under such a change is not tested exhaustively. The purpose of the method is to work only with entries passed in addition to the new status.

testplan/runnable/interactive/http.py
```diff
@@ -46,6 +54,89 @@ def wrapper(*args, **kwargs):
...
+    cases = []
+    for entry in suite_entry["entries"]:
+        if entry["category"] == ReportCategories.TESTCASE:
+            cases.append(entry["name"])
+        elif entry["category"] == ReportCategories.PARAMETRIZATION:
+            cases.extend(_extract_cases_from_parametrization(entry))
+    return cases
+
+
+def _extract_suites_cases(shallow_entry: Dict) -> Dict[str, List[str]]:
```
M6AI, line 87: This function is used to extract a testsuite to testcase list dictionary that is in line with how the suites and cases are prepared for running in the iterative runner of the MultiTest.

testplan/runnable/interactive/http.py
```diff
@@ -46,6 +54,89 @@ def wrapper(*args, **kwargs):
...
+    entries = {}
+    for child in entry["entries"]:
+        if child["category"] == ReportCategories.TESTCASE:
+            entries[child["name"]] = {}
+        else:
+            entries[child["name"]] = _extract_entries(child)
+    return entries
+
+
+def validate_put_json_body(json_body: Dict) -> None:
```
M6AI, line 129: Just moving out this repeated validation into a utility function.

testplan/runnable/interactive/http.py
```diff
@@ -46,6 +54,89 @@ def wrapper(*args, **kwargs):
...
+    elif category == ReportCategories.PARAMETRIZATION:
+        suites_cases[
+            shallow_entry["parent_uids"][2]
+        ] = _extract_cases_from_parametrization(shallow_entry)
+    elif category == ReportCategories.TESTCASE:
+        suites_cases[shallow_entry["parent_uids"][2]] = [shallow_entry["name"]]
+    return suites_cases
+
+
+def _extract_entries(entry) -> Dict[str, Dict]:
```
M6AI, line 113: This function is to generate a "tree" of entries essentially to be used in setting runtime status for a selection of entries only.

M6AI:
Still need to add a few tests to cover the `suites_cases` passed as not None.

testplan/report/testing/base.py
```diff
@@ -913,15 +882,17 @@ def status(self, new_status):
...
         """
-        Used for interactive mode, the runtime status of a testcase may be one
-        of ``RuntimeStatus``.
+        The runtime status is used in interactive mode and stands for
+        whether a particular entry is READY, WAITING, RUNNING,
+        RESETTING, FINISHED, or NOT_RUN.
         """
         return self._runtime_status
 
+    # In principle, this does not have to change
```
zhenyu-ms, line 893: what does this comment mean?

testplan/runnable/interactive/http.py
```diff
@@ -46,6 +54,93 @@ def wrapper(*args, **kwargs):
...
+    """
+    Given a  report entry, extracts all entries.
```
zhenyu-ms, line 117:
```suggestion
    Given a report entry, extracts all entries.
```

testplan/report/testing/base.py
M6AI: Good question, I think I was looking into the runtime status early in development and left some comments for myself as markers. Will remove.

testplan/testing/base.py
```diff
@@ -890,7 +889,9 @@ def dry_run(self):
 
         return result
 
-    def run_testcases_iter(self, testsuite_pattern="*", testcase_pattern="*"):
+    def run_testcases_iter(
```
zhenyu-ms, line 892: just wondering if we can only keep `suites_cases`

M6AI: This is a major point here. I was afraid to get rid of them for the moment as suites_cases is dynamically changing based on the shallow report entry passed down on put requests to the backend. Allowing it to be None let me be on the safe side so this is more of a defensive programming here. But let us discuss, because I think you feeling is correct and I share that only suites_cases (perhaps with some tweaking) could be enough.

M6AI: Actually, I think at this point the only real change would be to refactor our testing to mock the shallow entry in the new sense. suites_cases is never None as long as the filter matches at least a suite level entry (or something below that), but the way we run interactive mode without UI in our tests would lead to ~5-10 failures. I can think of cleaning this up as a hygiene item, but perhaps it is risky to put it at that priority.

testplan/runnable/interactive/http.py
```diff
@@ -46,6 +54,93 @@ def wrapper(*args, **kwargs):
...
+
+def _extract_suites_cases(shallow_entry: Dict) -> Dict[str, List[str]]:
+    """
+    Given a shallow report entry, extracts the testsuites and testcases
+        to be run.
+
+    :param shallow_entry: holds entry name and category for all children
+    :return: mapping of testsuites to testcases to be run
+    """
+    if not hasattr(shallow_entry, "entries"):
```
M6AI, line 95: This should be `if "entries" not in shallow_entry:`.

testplan/runnable/interactive/http.py
```diff
@@ -46,6 +54,93 @@ def wrapper(*args, **kwargs):
...
+
+
+def _extract_entries(entry) -> Dict[str, Dict]:
+    """
+    Given a  report entry, extracts all entries.
+
+    :param entry: report entry
+    :return: nested dictionary of entries mapped to entries recursively
+    """
+    if not hasattr(entry, "entries"):
```
M6AI, line 122: This should be `if "entries" not in entry:`

testplan/runnable/interactive/base.py
```diff
@@ -303,11 +325,19 @@ def run_test_suite(self, test_uid, suite_uid, await_results=True):
             self.logger.debug('Run suite ["%s" / "%s"]', test_uid, suite_uid)
             self._update_reports(
                 self.test(test_uid).run_testcases_iter(
-                    testsuite_pattern=suite_uid
+                    testsuite_pattern=suite_uid,
+                    suites_cases=suites_cases,
```
Pyifan, line 329: I would that not to pass suites_cases down to Test level object, but to expand that here and call run_testcases_iter multiple times with testsuite_pattern & testcase_pattern.

Pyifan (changes requested):

doc/en/interactive.rst
```diff
@@ -33,10 +33,10 @@ interrupt processes started as part of the test environment.
 
 .. note::
 
-    The interactive mode is currently in a beta stage and under active development.
-    There are still a few features yet to be implemented, for example the filter
-    box, tree view navigation and timing view, etc. Please bear with us while we
-    work through these, and feel free to raise an issue for any bugs you may find.
+    The interactive mode is currently in under active development.
```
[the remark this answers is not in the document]
M6AI: Agree, just wanted to be on the safe side, but we can remove it.

doc/newsfragments/2277_new.interactive_mode_filter_run.rst
```diff
@@ -0,0 +1 @@
+Interactive UI now runs only cases visible on the navigation to be in line with filter expression.
```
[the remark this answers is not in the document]
M6AI: Agree, this is misleading on a second read. More like matching patterns at this point. Will update.

testplan/web_ui/testing/src/Report/InteractiveReport.js
```diff
@@ -396,7 +396,6 @@ class InteractiveReport extends BaseReport {
       uids: currReportEntry.uids,
       counter: currReportEntry.counter,
     };
-    delete newEntry.entry_uids;
```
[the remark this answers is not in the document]
M6AI: This is of no use to be honest and it is not generated anymore in shallowreport. I can add back the functionality and keep it, but it does not affect the behavior overall.

testplan/web_ui/testing/src/Report/InteractiveReport.js
```diff
@@ -457,8 +456,28 @@ class InteractiveReport extends BaseReport {
...
+
+    // copy all of the fields from the input object to the new object, except for the "entries" field
+    for (let key in reportEntry) {
+      if (key !== "entries") {
+        shallowEntry[key] = reportEntry[key];
+      }
+    }
+
+    // if the input object has an "entries" field, process it recursively
+    if (reportEntry.entries) {
```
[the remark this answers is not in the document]
M6AI: Sounds like it can affect reruns. I will check and circle back.

testplan/runnable/interactive/base.py
M6AI: I checked this logic. We would repeatedly generate the test context because the property rather dynamic. It has a quadratic complexity so I would be in favor of keeping the suites_cases.

Pyifan (changes requested):
I gave the api interface some thoughts. We have four layers here:

1) http api: from Report/put to SingleTestcase/put, there are 5 of them
2) IRunner level: from run_all_test to run_test_case, there are 5 of them
3) MultiTest level: we have run_testcase_iter
4) MultiTest level: we have _run_testsuite_iter & _run_testcase_iter

My suggestion of what each layer does
1) http api: deserialize the json report, and decide whether a filter is present - if not, we don't even need to pass the json content to next layer right? If yes, then the next layer will receive an optional parameter (currently it's called suites_cases, but I think we need a better name). The runtime status also needs to be taken care here, which is a bit tricky...
2) IRunner level apis needs a new optional parameter (except for run_test_case, it runs only 1 testcase so we cannot further filter). I suggest call this parameter something like ‘filter'.
3) run_testcase_iter shall also take a `filter` parameter, and this is where the filter is decoded, and compared with test_context to determine which testsuites/testcases needs to run. 
4) this layer may not need any change.

PR #960, opened 2023-06-01
Updated metadata collecting in PDFExporter

rnemes:
PDFExporter failed to export metadata.

## Solution description
Fixed metadata collecting in PDFExporter and added new fields and also fixed text wrapping.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/exporters/testing/pdf/renderers/base.py
+++ b/testplan/exporters/testing/pdf/renderers/base.py
@@ -136,10 +136,11 @@ class MetadataMixin:
         Return metadata context to be rendered
         on the PDF with readable labels.
         """
+        metadata = dict(source.information)
         return collections.OrderedDict(
             [
-                (label, dict(source.information)[key])
+                (label, metadata[key])
                 for key, label in self.get_metadata_labels()
-                if dict(source.information).get(key)
+                if key in metadata
             ]
         )
--- a/testplan/exporters/testing/pdf/renderers/base.py
+++ b/testplan/exporters/testing/pdf/renderers/base.py
@@ -138,8 +138,8 @@ class MetadataMixin:
         """
         return collections.OrderedDict(
             [
-                (label, source.meta[key])
+                (label, dict(source.information)[key])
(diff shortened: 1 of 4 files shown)


testplan/exporters/testing/pdf/renderers/reports.py
```diff
@@ -111,31 +115,45 @@ def get_row_data(self, source, depth, row_idx):
...
+                        font=(const.FONT_BOLD, const.FONT_SIZE_SMALL),
+                        start_column=0,
+                        end_column=0,
+                    ),
+                    RowStyle(
+                        font=(const.FONT, const.FONT_SIZE_SMALL),
+                        start_column=1,
+                        end_column=1,
+                    ),
+                    RowStyle(
```
rnemes, line 150: Because of the behaviour of RowData.append() function the RowStyles are getting merged for columns if we pass a list of the rows as content. This is bot acceptable for span style because it would merge all columns. So we either add the columns separately or we manually add the span styles to the RowData instance after inserting the rows.

testplan/exporters/testing/pdf/renderers/base.py
```diff
@@ -138,8 +138,8 @@ def get_metadata_context(self, source):
         """
         return collections.OrderedDict(
             [
-                (label, source.meta[key])
+                (label, dict(source.information)[key])
```
kn-ms, line 141: create this dict once outside of comprehension

testplan/exporters/testing/pdf/renderers/base.py
```diff
@@ -138,8 +138,8 @@ def get_metadata_context(self, source):
...
                 for key, label in self.get_metadata_labels()
-                if source.meta.get(key)
+                if dict(source.information).get(key)
```
kn-ms, line 143: use key in ... instead of the get, it is nicer to read unless you want to ignore all keys that does not exist and the ones which are falsy

kn-ms (approved):

PR #963, opened 2023-06-13
Feature/log rotate

kn-ms:
No Log rotation support in LogMatcher

## Solution description
Added Log rotation support. User can pass a glob instead of a logfile. Logfiles matching the glob ordered by mtime.

Some simpler helper class added as well, during the gradual refactoring. RotatingLogStream can be extended with different log rotation strategies.

## Checklist:
- [X] Test
- [X] Documentation (API)
- [X] MS info leakage check


testplan/common/utils/logfile.py
```diff
@@ -0,0 +1,303 @@
...
+from itertools import dropwhile
+from os import PathLike, SEEK_END
+from typing import Union, Optional, Generic, TypeVar, List
+
+
+class LogPosition:
+    pass
+
+
+T = TypeVar("T", str, bytes)
```
yuxuan-ms, line 15: AnyStr

testplan/common/utils/logfile.py
```diff
@@ -0,0 +1,303 @@
...
+        return self.file.readline()
+
+    @property
+    def position(self) -> FileLogPosition:
+        return self.FileLogPosition(position=self.file.tell())
+
+    def compare(
+        self, position1: FileLogPosition, position2: FileLogPosition
+    ) -> int:
+        return position1.position - position2.position
```
yuxuan-ms, line 89: why not use `__sub__`?

testplan/common/utils/match.py
```diff
@@ -132,14 +219,9 @@ def match(
...
-        # As a convenience, we create the compiled regex if a string was
-        # passed.
-        if not hasattr(regex, "match"):
-            regex = re.compile(regex)
-        if isinstance(regex.pattern, str):
-            read_mode = "r"
+        regex = self._prepare_regexp(regex)
 
-        with open(self.log_path, read_mode) as log:
+        with closing(self.log_stream) as log:
```
Pyifan, line 224: Is it necessary to open-close the log_stream every time we match? I see the __del__ already do close, is that good enough?

testplan/common/utils/match.py
```diff
@@ -155,7 +237,7 @@ def match(
                 else:
                     break
 
-            self.position = log.tell()
+            self.position = self.log_stream.position
```
Pyifan, line 240: Also do we need to maintain the current position at LogMatch level now that the log_stream object also keeps record of current position. Maybe it's good enough to just keep a few marks at LogMatcher level? and delegate to log_stream to start from current position.

testplan/common/utils/match.py
kn-ms: That was the original behaviour. Having it in the del is just for a last resort, unbfortunately there are no guaranties that the logmatcher is already garbage collected when the drivers/runtime clean off logfiles in some of the cases. I wanted to be on the safe side here.

testplan/common/utils/match.py
kn-ms: log matcher keeps the position for continuous read, however if the stream is closed and reopened it is not maintained, taht I want to keep for now, see above.

testplan/common/utils/logfile.py
```diff
@@ -0,0 +1,303 @@
...
+        position: int = 0
+
+    def __init__(self, path: Union[PathLike, str], binary: bool) -> None:
+        self._path = path
+        self._binary = binary
+        self._file: Optional[TextIOWrapper] = None
+
+    @property
+    def file(self):
+        mode = "rb" if self._binary else "rt"
```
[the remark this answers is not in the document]
kn-ms: I went for full inheritence in the last update

testplan/common/utils/logfile.py
kn-ms: ordering in LogPosition could not be interpretted without the actual stream (like for rotated logfiles), I do not want the position have a refference to the stream. In the simple single file case of course they are comparable without the stream, but for the generic api the compare function should be a member of the stream,

Pyifan (approved):

PR #965, opened 2023-06-16
Changed default exporter behaviour

rnemes:
Introducing new ExportContext class for exporters to communicate with each other.

## Solution description
See code

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check


examples/Test Output/Exporters/JSON/json_to_pdf.py
kn-ms, line 1: I think we can get rid of this example, better to use tpr

testplan/cli/commands/writers.py
```diff
@@ -16,6 +17,7 @@
 
 
 writer_commands = CommandList()
+export_context = ExportContext()
```
kn-ms, line 20: module level export context might not be the best idea, as it putting all the exporters into a single context, I'm not sure that is what we always want. I suggest to keep separate unique contexts for now for tpr.

testplan/runnable/base.py
```diff
@@ -228,7 +233,7 @@ class TestRunnerResult(RunnableResult):
     def __init__(self):
         super(TestRunnerResult, self).__init__()
         self.test_results = OrderedDict()
-        self.exporter_results = []
+        self.export_context = ExportContext()
```
kn-ms, line 236: I'm not sure the whole context is needed for the results. The context is for the exporters to communicate, and each generating results, so keep it here

testplan/runnable/base.py
```diff
@@ -1174,28 +1179,65 @@ def _invoke_exporters(self):
...
-                self._result.exporter_results.append(exp_result)
+                self._result.export_context.results.append(exp_result)
             else:
                 raise NotImplementedError(
                     "Exporter logic not implemented for: {}".format(
                         type(exporter)
                     )
                 )
 
+        self.logger.user_info("Yupi")
```
kn-ms, line 1233: Be happy, but be silent :)

tests/functional/testplan/exporters/testing/test_pdf.py
```diff
@@ -17,6 +18,8 @@
 from testplan.report.testing import styles
 from testplan.testing.multitest.entries import assertions
 
+export_context = ExportContext()
```
kn-ms, line 21: make this a fixture, and let every testcase get it's own, or just create it in the testcase

examples/Test Output/Exporters/Custom Exporters/Basic/test_plan.py
```diff
@@ -74,7 +74,19 @@ def __init__(self, file_path):
     def get_text_content(self, source):
         raise NotImplementedError
 
-    def export(self, source: TestReport) -> Optional[str]:
+    def export(
+        self,
+        source: TestReport,
+        export_context: ExportContext,
+    ) -> ExporterResult:
```
M6AI, line 81: The change in signature reminds me that exporters are user facing and they may have their own customizations. Do we know the changes may affect these users? If yes, I suggest we make note of this change and put together some doc.

testplan/runnable/base.py
```diff
@@ -1165,31 +1170,91 @@ def _pre_exporters(self):
...
+
+        :param exporter: exporter to run
+        :param source: Testplan report to export
+        :param export_context: information about other exporters
+        :return: ExporterResult object containing information about the actual exporter object and its possible output
+        """
+
+        traceback_string = None
+        exp_result = None
+        try:
```
zhenyu-ms, line 1190: i think we'd better check the method signature here to make things more explicit

testplan/runnable/base.py
```diff
@@ -1165,31 +1170,91 @@ def _pre_exporters(self):
...
+                ),
+                exporter,
+            )
+            exp_result = exporter.export(
+                source=source,
+            )
+        except Exception:
+            traceback_string = traceback.format_exc()
+        finally:
+            if isinstance(exp_result, ExporterResult):
```
zhenyu-ms, line 1210: will the if branch be invoked?

testplan/runnable/base.py
```diff
@@ -1165,31 +1170,91 @@ def _pre_exporters(self):
...
+                        (
+                            "Exporter '%s' returns with not an ExporterResult object! "
+                            "This will be prohibited in the future, please modify your custom exporter!"
+                        ),
+                        exporter,
+                    )
+                exp_result = ExporterResult(
+                    exporter=exporter,
+                    traceback=traceback_string,
+                    result={"unknown": exp_result},
```
zhenyu-ms, line 1226: since we cannot make any assumptions here, i'd suggest we use `Any` as result type

actually, i think we can just have `export` methods returning `Any` since we need to wrap the exception and actual result here nevertheless. am i missing anything here?

doc/newsfragments/2528_new.export_context.rst
```diff
@@ -0,0 +1 @@
+Introduced new :py:class:`ExportContext <testplan.common.exporters.ExportContext>` class for exporters to communicate with each other.
```
M6AI, line 1: As suggested below, probably breaking change, so needs to be reflected.

M6AI (commented):
Few comments for now.

testplan/runnable/base.py
rnemes: There are opposite opinions, discuss this on the team meeting.

doc/newsfragments/2528_new.export_context.rst
rnemes: It shouldn't be breaking. The case when the user uses a custom exporter which does not implement the new features is handled and will show a warning message for now. (This will be removed in the future). Let me test it more.

examples/Test Output/Exporters/Custom Exporters/Basic/test_plan.py
rnemes: It shouldn't be breaking. The case when the user uses a custom exporter which does not implement the new features is handled and will show a warning message for now. (This will be removed in the future).
Creating a doc describing the changes is a good idea!

testplan/runnable/base.py
rnemes: We would like to move the users to use ExportContest and ExporterResult in their custom exporters. The code will handle if not and will show a warning message.

testplan/runnable/interactive/http.py
```diff
@@ -607,7 +609,9 @@ def post(self):
                         try:
                             report = copy.deepcopy(ihandler.report)
                             report.reset_uid()
-                            export_path = exporter.export(report)
+                            export_path = exporter.export(
```
kn-ms, line 612: Now this export path will be a dictionary, it used to be a string. I think this need a bit oof thought, what to show to interactive users whne they export

kn-ms: Also this seems does not use the helper that handles deprecated exporters

Pyifan (dismissed):

testplan/runnable/interactive/http.py
```diff
@@ -595,27 +595,41 @@ def get(self):
             }
 
         @api.expect(post_export_model)
-        def post(self):
+        def post(self) -> Optional[Dict]:
             save_exports = request.json
```
Pyifan, line 599: e.g save_reports -> request_data

testplan/runnable/interactive/http.py
```diff
@@ -595,27 +595,41 @@ def get(self):
...
-                            export_path = _run_exporter(
+                            exporter_result = _run_exporter(
                                 exporter=exporter,
                                 source=report,
                                 export_context=export_context,
-                            ).result
-                            export_result["success"] = True
-                            export_result["message"] = export_path
+                            )
+                            export_result = {
```
Pyifan, line 612: exporter_result v.s export_result (eye crossed, lol)

consider renaming export_result -> request_result? (i'm open to other better names)

Pyifan (approved):

PR #968, opened 2023-06-26
Fix events cannot be serialized in tpr issue.

yuxuan-ms:

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/common/report/base.py
```diff
@@ -223,6 +223,24 @@ class EventRecorder:
...
     children: List["EventRecorder"] = dataclasses.field(default_factory=list)
 
+    @classmethod
+    def load(
+        cls, event_record: Union["EventRecorder", Dict]
+    ) -> "EventRecorder":
+        if isinstance(event_record, cls):
+            return event_record
+
+        event = cls(
```
kn-ms, line 233: this could be written as event=cls(**event_record). Though this works I do not like this load idea, as it is not the same as we follow with other schema objects, where proper deserialization is enclosed in the schema close to the schema definition.

yuxuan-ms: updated

Pyifan (approved):

kn-ms (commented):
The code looks good. I think it deserves a news fragment, and there is a typo in pr title trp -> tpr

PR #972, opened 2023-07-06
MultiTest fast fail

M6AI:
Added fast fail feature to MultiTest. Added associated unit test and newsfragment.

MultiTest lacks an option to abort upon first failed testcase.

## Solution description
Add a similar logic as for stop on error, however, we want to abort the entire MultiTest not just current suite.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/base.py
+++ b/testplan/testing/multitest/base.py
@@ -29,6 +29,7 @@ from testplan.report import (
 )
 from testplan.testing import base as testing_base
 from testplan.testing import filtering, tagging
+from testplan.testing.environment import DriverDepGraph
 from testplan.testing.multitest import result
 from testplan.testing.multitest import suite as mtest_suite
 from testplan.testing.multitest.entries import base as entries_base
@@ -189,6 +190,7 @@ class MultiTestConfig(testing_base.TestConfig):
             config.ConfigOption("thread_pool_size", default=0): int,
             config.ConfigOption("max_thread_pool_size", default=10): int,
             config.ConfigOption("stop_on_error", default=True): bool,
+            config.ConfigOption("stop_on_failure", default=False): bool,
             config.ConfigOption("part", default=None): Or(
                 None,
                 And(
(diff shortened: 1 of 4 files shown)


testplan/testing/multitest/base.py
```diff
@@ -937,6 +950,13 @@ def _run_serial_testcases(self, testsuite, testcases):
                         testsuite.name,
                     )
                     break
+            elif testcase_report.status == Status.FAILED:
```
Pyifan, line 953: Because we are checking self.active in the for loop, so we can achieve by set self._should_abort instead of explicitly break? Also it might be better to set self._should_abort in _run_testcase method, rather than need to do it in both _run_serial_testcases and _run_parallel_testcases

testplan/testing/multitest/base.py
```diff
@@ -988,17 +1008,22 @@ def _run_parallel_testcases(self, testsuite, execution_groups):
                 # If any testcase errors and we are configured to stop on
                 # errors, we still wait for the rest of the current execution
                 # group to finish before stopping.
-                if (
-                    testcase_report.status == Status.ERROR
-                    and self.cfg.stop_on_error
-                ):
-                    should_stop = True
+                if testcase_report.status == Status.ERROR:
```
Pyifan, line 1011: The original _run_parallel_testcases implementation looks problematic to me. How do we know the futures have all finished running before we do future.result().

testplan/testing/multitest/base.py
```diff
@@ -988,17 +1008,22 @@ def _run_parallel_testcases(self, testsuite, execution_groups):
...
+                    if self.cfg.stop_on_error:
+                        self.logger.debug(
+                            'Stopping execution of testsuite "%s" due to error.',
+                            testsuite.name,
+                        )
+                        should_stop = True
+                elif testcase_report.status == Status.FAILED:
```
Pyifan, line 1018: Also I think the expected behavior is if we are executing testcases in parallel, and when error/failure occurs, testcases that have not started shall cancel.

testplan/testing/multitest/base.py
```diff
@@ -496,11 +495,23 @@ def run_tests(self):
...
 
+                if (
+                    testsuite_report.status == Status.FAILED
+                    and self.cfg.fast_fail
+                ):
+                    self.logger.debug(
+                        'Aborting MultiTest "%s" due to testcase failure.',
+                        self.name,
+                    )
+                    self.abort()
```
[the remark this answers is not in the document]
Pyifan: The logic is if stop_on_error is working fine, then we shall not need explicitly call abort() here.

Pyifan (changes requested):

testplan/testing/multitest/base.py
```diff
@@ -189,6 +190,7 @@ def get_options(cls):
             config.ConfigOption("thread_pool_size", default=0): int,
             config.ConfigOption("max_thread_pool_size", default=10): int,
             config.ConfigOption("stop_on_error", default=True): bool,
+            config.ConfigOption("fast_fail", default=False): bool,
```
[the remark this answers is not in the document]
M6AI: I think the original feature for which we are trying to achieve parity is fast_fail. I am ok with the change on my end, not sure about our users.

testplan/testing/multitest/base.py
```diff
@@ -496,11 +495,23 @@ def run_tests(self):
 
                 testsuite_report = self._run_suite(testsuite, testcases)
                 report.append(testsuite_report)
+                # We should signal here that something is wrong with the suite and
```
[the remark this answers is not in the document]
M6AI: We can do that, although the abort itself looks proper to me.

testplan/testing/multitest/base.py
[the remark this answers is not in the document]
zhenyu-ms: i think parameter name change is ok

testplan/testing/multitest/base.py
M6AI: future.result waits for completion as per my local check. So it is functioning as it is proper and you can see the other cases executing in the background.

Pyifan: I see, i was wrong in my previous comment.

PR #980, opened 2023-08-09
Fix Attribute Errors

M6AI:
There are two attribute errors related to the interactive UI with filter expression feature. One is a silent exception thrown on rerun of single testcases on a missing length attribute. The other is related to running a single testcase where the filtered runtime status was not implemented.

## Solution description
Added proper handling for these scenarios and an additional test for running a single testcase with an API call.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/doc/newsfragments/2649_changed.fix_attribute_error_prune.rst
+++ b/doc/newsfragments/2649_changed.fix_attribute_error_prune.rst
@@ -1,0 +1,1 @@
+Fixed silent exception thrown by the UI when attempting a rerun of a single testcase with an active filter.
--- a/doc/newsfragments/2649_changed.fix_attribute_error_prune.rst
+++ b/doc/newsfragments/2649_changed.fix_attribute_error_prune.rst
@@ -1,1 +1,0 @@
-Fixed silent exception thrown by the UI when attempting a rerun of a single testcase with an active filter.
--- a/testplan/runnable/interactive/http.py
+++ b/testplan/runnable/interactive/http.py
@@ -919,7 +919,7 @@ def _extract_entries(entry: Dict) -> Dict:
     """
     entries = {}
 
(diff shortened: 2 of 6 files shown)


doc/newsfragments/2649_changed.fix_attribute_error_prune.rst
```diff
@@ -0,0 +1 @@
+Fixed silent exception thrown by the UI when attempting a rerun of a single testcase with an active filter.
```
Pyifan, line 1: Merge the 2 news fragment? they read very similar to me.

testplan/runnable/interactive/http.py
```diff
@@ -919,10 +919,10 @@ def _extract_entries(entry: Dict) -> Dict:
     """
     entries = {}
 
-    # NOTE: we assume "entries" is present, see application of the function
-    for child in entry["entries"]:
-        if child["category"] == ReportCategories.TESTCASE:
-            entries[child["name"]] = {}
+    # NOTE: we ignore assertions (DEFAULT category) and testcases
+    for child in entry.get("entries", []):
+        if child["category"] in ("DEFAULT"):
```
Pyifan, line 924: I don't think we shall rely on the category of child. Can we check the category of the entry upfront, as the stop condition for this recursive function?

Pyifan (changes requested):

testplan/runnable/interactive/http.py
M6AI: Let me check I remember, having some weird scenario, but perhaps with the update to frontend that is gone as well.

kn-ms (approved):

Pyifan (approved):

PR #981, opened 2023-08-10
Added a new 'Run all' button to interactive toolbar on the web interface

rnemes:

## Solution description

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -227,11 +227,14 @@ class TestRunnerIHandler(entity.Entity):
             self._update_reports([(self.test(test_uid).dry_run().report, [])])
 
     def run_all_tests(
-        self, await_results: bool = True
+        self,
+        shallow_report: Optional[Dict] = None,
+        await_results: bool = True,
     ) -> Union[TestReport, Awaitable]:
         """
         Runs all tests.
 
+        :param shallow_report: shallow report entry, optional
         :param await_results: Whether to block until tests are finished,
             defaults to True.
         :return: If await_results is True, returns a testplan report.
@@ -239,12 +242,20 @@ class TestRunnerIHandler(entity.Entity):
             ready.
         """
         if not await_results:
-            return self._run_async(self.run_all_tests)
+            return self._run_async(
+                self.run_all_tests, shallow_report=shallow_report
+            )
 
-        self.logger.debug("Interactive mode: Run all tests")
-
-        for test_uid in self.all_tests():
-            self.run_test(test_uid)
+        if shallow_report:
+            self.logger.debug("Interactive mode: Run filtered tests")
+            for multitest in shallow_report["entries"]:
+                self.run_test(
+                    test_uid=multitest["name"], shallow_report=multitest
+                )
+        else:
+            self.logger.debug("Interactive mode: Run all tests")
+            for test_uid in self.all_tests():
+                self.run_test(test_uid=test_uid)
 
     def run_test(
         self,
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -800,25 +800,30 @@ class TestRunnerIHandler(entity.Entity):
     def reload_report(self):
         """Update report with added/removed testcases"""
         new_report = self._initial_report()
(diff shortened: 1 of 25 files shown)


testplan/web_ui/testing/src/Report/InteractiveReport.js
```diff
@@ -72,6 +74,7 @@ class InteractiveReportComponent extends BaseReport {
       navWidth: `${INTERACTIVE_COL_WIDTH}em`,
       resetting: false,
       reloading: false,
+      running: false,
```
M6AI, line 77: So I looked at this new status and wonder two things:
- Can we do without it?
- If not, then we need to block certain actions when it is True. Right now I think we can trigger it twice without problem, send a reload during it, reset, etc... See for instance, how the reloadCode in InteractiveReport is handling it.

M6AI (changes requested):
Minor update but more of an ask for some deep-dive, let us follow up on it offline.

testplan/web_ui/testing/src/Report/InteractiveReport.js
rnemes: > Can we do without it?

We only need that for logging the state of the progress. However, we could use another variable eg.: _runtime_status_ which could replace _resetting_, _reloading_, _aborting_ as well, but in that case need to refactor the related functionalities.

> If not, then we need to block certain actions when it is True. Right now I think we can trigger it twice without problem, send a reload during it, reset, etc... See for instance, how the reloadCode in InteractiveReport is handling it.

That makes sense. I will update the code.

testplan/web_ui/testing/src/Report/InteractiveReport.js
```diff
@@ -497,6 +501,22 @@ class InteractiveReportComponent extends BaseReport {
...
+      return;
+    } else {
+      const updatedReportEntry = {
+        ...this.shallowReportEntry(this.state.report),
+        runtime_status: "running",
+      };
+      this.putUpdatedReportEntry(updatedReportEntry);
+      this.setState({ running: true });
+    }
+  }
```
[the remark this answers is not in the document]
rnemes: No, it doesn't handle filtering, it starts everything.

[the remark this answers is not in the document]
M6AI: As mentioned offline, this will be needed.

testplan/runnable/interactive/base.py
```diff
@@ -227,24 +227,35 @@ def reset_test(self, test_uid, await_results=True):
...
-
-        self.logger.debug("Interactive mode: Run all tests")
+            return self._run_async(
+                self.run_all_tests, shallow_report=shallow_report
+            )
 
-        for test_uid in self.all_tests():
-            self.run_test(test_uid)
+        if shallow_report:
+            self.logger.debug("Interactive mode: Run filtered tests")
```
M6AI, line 250: Loglines are a welcome addition, but I do not see it for the others. Also I wonder if it is INFO level already or perhaps USER_INFO.

M6AI (dismissed):
Tested offline too.

testplan/web_ui/testing/src/Toolbar/InteractiveButtons.js
```diff
@@ -141,6 +143,60 @@ export const AbortButton = (props) => {
...
+            className={css(styles.toolbarButton, styles.toolbarInactive)}
+            icon={faHourglass}
+            title="Running tests..."
+            spin
+          />
+        </div>
+      </NavItem>
+    );
+  } else {
+    if (props.filter) {
```
yuxuan-ms, line 168:
```
if (pops.filter) {
  title="Run filtered tests"
} else {
  title = "Run all tests"
}
return (
  <NavItem key="runall-button">
          <div className={css(styles.buttonsBar)}>
            <FontAwesomeIcon
              key="toolbar-runall"
              className={css(styles.toolbarButton)}
              icon={faPlay}
              title={title}
              onClick={props.runAllCbk}
            />
          </div>
        </NavItem>
)
```

M6AI (approved):

PR #982, opened 2023-08-13
Failed setup runtime status fix

M6AI:
Fixed a defect whereas failure in setup would cause the runtime status to not move from WAITING. Added a unit test to cover the scenario and a newsfragment.

A failed setup results only in the teardown report yielded in the iterative testcase runner and does not set the runtime status back to "ready" or "not run" for any of the cases, leaving the UI hanging on those cases forever.

## Solution description
Added a method to forcefully skip cases and update their status to "not run".

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/base.py
+++ b/testplan/testing/multitest/base.py
@@ -1297,7 +1297,9 @@ class MultiTest(testing_base.Test):
 
             if setup_report.failed:
                 # NOTE: we are going to skip the cases and update the status
-                for status, parent_uids in self._skip_testcases(testsuite, testcases):
+                for status, parent_uids in self._skip_testcases(
+                    testsuite, testcases
+                ):
                     yield status, parent_uids
                 teardown_report = self._teardown_testsuite(testsuite)
                 if teardown_report is not None:
@@ -1322,9 +1324,7 @@ class MultiTest(testing_base.Test):
         :return: list of parent UIDs
         """
 
-        param_template = getattr(
-            testcase, "_parametrization_template", None
-        )
+        param_template = getattr(testcase, "_parametrization_template", None)
         if param_template:
             parent_uids = [
                 self.uid(),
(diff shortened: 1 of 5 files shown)


testplan/testing/multitest/base.py
```diff
@@ -1310,6 +1313,47 @@ def _run_testsuite_iter(self, testsuite, testcases):
...
+            parent_uids = [
+                self.uid(),
+                testsuite.uid(),
+                testcase._parametrization_template,
+            ]
+        else:
+            parent_uids = [self.uid(), testsuite.uid()]
+        return parent_uids
+
+    def _skip_testcases(self, testsuite, testcases):
```
M6AI, line 1338: This may not be ideal, but functions as expected. We can think of a better name, simply add the necessary logic to the testcase runner method itself, or making this even more specialized.

testplan/testing/multitest/base.py
```diff
@@ -1310,6 +1313,47 @@ def _run_testsuite_iter(self, testsuite, testcases):
         if teardown_report is not None:
             yield teardown_report, [self.uid(), testsuite.uid()]
 
+    def _get_parent_uids(self, testsuite, testcase):
```
M6AI, line 1316: Adding a new function I noticed we lack annotations in this module, but a bit puzzled what a testsuite (which is essentially any testsuite decorated class instance) and a testcase (a decorated method of said class) is in terms of a reasonable type.

tests/unit/testplan/runnable/interactive/test_irunner.py
```diff
@@ -161,6 +181,38 @@ def test_run_suite(irunner, sync):
...
+        test_filter=filtering.Filter(),
+        test_sorter=ordering.NoopSorter(),
+        stdout_style=defaults.STDOUT_STYLE,
+    )
+    local_runner.add(test, test.uid())
+    target.resources.add(local_runner)
+
+    with mock.patch("cheroot.wsgi.Server"), mock.patch(
+            "testplan.runnable.interactive.reloader.ModuleReloader"
+    ) as MockReloader:
```
M6AI, line 200: Followed the style of the module here, but I think pytest allows parametrized fixtures, so we could simplify this module at some point quite a lot.

kn-ms (approved):

PR #983, opened 2023-08-14
Fix parametrization reload

M6AI:
Currently, the reload feature of the interactive UI does not capture changes to parametric testcases at parametrization level. In fact, changing any of the parameters will result in the old parameters possibly pointing to non-existent cases and the UI becomes hanging.

## Solution description
Added parametrization level to the report reload so that changes to parametric testcases are captured.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/web_ui/testing/src/Report/InteractiveReport.js
+++ b/testplan/web_ui/testing/src/Report/InteractiveReport.js
@@ -224,7 +224,7 @@ class InteractiveReportComponent extends BaseReport {
                 const existingParametrization =
                   existingSuite &&
                   existingSuite.entries.find(
-                    (entry) => entry.uid === newTestCase
+                    (entry) => entry.uid === newTestCase.uid
                   );
 
                 if (


testplan/runnable/interactive/base.py
```diff
@@ -797,16 +797,29 @@ def reload_report(self):
...
                     try:
-                        # Update new report testcase state
-                        new_report[mt.uid][st.uid].entries[
-                            new_case_index
-                        ] = st[
-                            new_report[mt.uid][st.uid]
-                            .entries[new_case_index]
-                            .uid
-                        ]
+                        tc = new_report[mt.uid][st.uid].entries[new_case_index]
```
kn-ms, line 800: consider to introduce temporary variables instead of those long indexing chains. Speaking names could make this more(euphemism) readable.

M6AI: Moved most of it to enumerate.

kn-ms (dismissed):

testplan/runnable/interactive/base.py
```diff
@@ -789,25 +789,30 @@ def reload(self, rebuild_dependencies=False):
...
-                            new_case_index
-                        ] = st[
-                            new_report[mt.uid][st.uid]
-                            .entries[new_case_index]
-                            .uid
-                        ]
-                    except KeyError:  # new testcase
-                        pass
-                mt.entries[st_index] = new_st
+                        if isinstance(case, TestGroupReport):
```
Pyifan, line 797: Can we add some test coverage for this change? To prevent future regression.

Pyifan (changes requested):

Pyifan (approved):

PR #986, opened 2023-08-24
Fix!/fix certain naming issues, better patterns for multitests with parts

zhenyu-ms:
Spec to be discussed and updated.

## Solution description

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/testing/filtering.py
```diff
@@ -285,10 +293,61 @@ def parse_pattern(self, pattern: str) -> List[str]:
...
             )
 
-        return patterns + ([self.ALL_MATCH] * (self.MAX_LEVEL - len(patterns)))
+        test_level, suite_level, case_level = patterns + (
+            [self.ALL_MATCH] * (self.MAX_LEVEL - len(patterns))
+        )
 
-    def filter_test(self, test):
-        return fnmatch.fnmatch(test.name, self.test_pattern)
+        m = re.match(TEST_PART_REGEX, test_level)
```
kn-ms, line 300: I have two problem here, one is that the pattern match means limitation in the naming of multitests. 

The information currently is not available from theoutput of info:

```
python.exe .\test_plan.py --info pattern-full
MultitestParts
  MultitestParts::Suite1
    MultitestParts::Suite1::test_equal
    MultitestParts::Suite1::test_greater
  MultitestParts::Suite2
    MultitestParts::Suite2::test_bool <val=1>
    MultitestParts::Suite2::test_bool <val=4>
MultitestParts
  MultitestParts::Suite1
    MultitestParts::Suite1::test_not_equal
    MultitestParts::Suite1::test_approximate_equal
  MultitestParts::Suite2
    MultitestParts::Suite2::test_bool <val=2>
    MultitestParts::Suite2::test_bool <val=5>
MultitestParts
  MultitestParts::Suite1
    MultitestParts::Suite1::test_less
  MultitestParts::Suite2
    MultitestParts::Suite2::test_bool <val=0>
    MultitestParts::Suite2::test_bool <val=3>
Exiting code: 0
```

if one want to run just all from the 2nd part they need to count how many parts are there, and hoping the order is just correct in the output. If there is a problem in calculation they got nothing

```
python.exe ./test_plan.py --patterns "MultitestParts - part(1/2):*:*" --info pattern-full
Exiting code: 0
```

This also makes(keeps) multiple entry in the json lister for a single Multitest and anyone try to derive ids could not doo it from there. Also the --info is not in sync with ids in the result json and results xml, so one (a software) could not easily correlate betwing list of testcases from cli , the pattern need to be provided and the results it is generating.

If we would be just listing the uid or full name whatever is the best to call it, then patterns still can nicely works with `"MultitestParts*:Suite1:*` way. That looks simple in testplan (no restriction on naming, no parse name for filtering), simple the user to interpret and work with, fit nice for automation on listing <--> filtering <--> result crossreference.

kn-ms (changes requested):

testplan/common/report/base.py
```diff
@@ -57,11 +57,17 @@ class Report:
...
+        description: Optional[str] = None,
+        instance_name: Optional[str] = None,
+        uid: Optional[str] = None,
+        entries: Optional[list] = None,
+        parent_uids: Optional[List[str]] = None,
     ):
         self.name = name
         self.description = description
-
+        self.instance_name = instance_name or name
```
Pyifan, line 70: suggest rename instance_name to definition_name.
And the filters has a match_definition param, is the behavior per its name??

testplan/testing/multitest/suite.py
```diff
@@ -357,24 +357,14 @@ def _ensure_unique_generated_testcase_names(names, functions):
...
 
     # Functions should have different __name__ attributes after the step above
-    name_counts = collections.Counter(
-        itertools.chain(names, (func.__name__ for func in functions))
-    )
-    dupe_names = {k for k, v in name_counts.items() if v > 1}
-    assert len(dupe_names) == 0
+    names = list(itertools.chain(names, (func.__name__ for func in functions)))
+    if len(set(names)) != len(names):
+        raise RuntimeError("Duplicate testcase __name__ found.")
```
Pyifan, line 367: it will be more helpful if we tell user which testcase names are dup?

doc/newsfragments/2626_changed.pattern_with_parts.rst
```diff
@@ -0,0 +1,5 @@
+Filtering Patterns now are easier to use with MultiTest with parts.
+
+    * Both patterns with part specified or not could be used to filter MultiTest with parts.
```
Pyifan, line 3: Support filtering with MultiTest name pattern both with and without parts

doc/newsfragments/2626_changed.pattern_with_parts.rst
```diff
@@ -0,0 +1,5 @@
...
+    * Applying patterns won't cause certain testcase appear in some different part now.
```
Pyifan, line 4: Applying filtering pattern will not change parts shuffling - the same testcase always end up in the same part of a MultiTest.

Pyifan (changes requested):
When we change the part logic & add offset, do we need any change on merge end?

testplan/testing/filtering.py
```diff
@@ -275,7 +283,7 @@ def __repr__(self):
         return '{}(pattern="{}")'.format(self.__class__.__name__, self.pattern)
 
     def parse_pattern(self, pattern: str) -> List[str]:
-        # ":" or "::" can be used as delimiter
+        # ":" would be used as delimiter
         patterns = [s for s in pattern.split(self.DELIMITER) if s]
```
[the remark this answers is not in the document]
zhenyu-ms: oh didn't notice that, thanks Krisz, will revert it back

testplan/testing/filtering.py
```diff
@@ -285,10 +293,61 @@ def parse_pattern(self, pattern: str) -> List[str]:
...
+            test_name_p = m.group(1)
+            test_cur_part_p = m.group(2)
+            test_ttl_part_p = m.group(3)
+
+            try:
+                test_cur_part_p, test_ttl_part_p = int(test_cur_part_p), int(
+                    test_ttl_part_p
+                )
+            except ValueError:
+                pass
```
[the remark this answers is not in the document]
zhenyu-ms: well I need the else clause here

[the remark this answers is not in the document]
zhenyu-ms: the ValueError should only come out from converting to integer call, if both of them are integers have a simple sanity check, else just pass it to fnmatch calls

testplan/testing/multitest/suite.py
zhenyu-ms: sure thing

testplan/testing/multitest/base.py
```diff
@@ -379,8 +376,19 @@ def get_test_context(self):
         ctx = []
         sorted_suites = self.cfg.test_sorter.sorted_testsuites(self.cfg.suites)
 
+        g_offset = 0
         for suite in sorted_suites:
             testcases = suite.get_testcases()
+
+            if self.cfg.part and self.cfg.part[1] > 1:
```
zhenyu-ms, line 383: actually I may put parting below sorting

testplan/testing/filtering.py
zhenyu-ms: Sure I will definitely make the `--list` output conform with input patterns, will also check `--patterns` using together with `--info`

Tbh I don't like the single asterisk way. It can easily include other tests happening to have the same name prefix. It cannot express whether the user only wants to select cases from a specific part or from all the parts.

zhenyu-ms:
> When we change the part logic & add offset, do we need any change on merge end?

Update on dividing and merging cases could worth another PR, will revert related changes in this PR.

testplan/testing/multitest/parametrization.py
```diff
@@ -271,18 +274,18 @@ def _parametrization_name_func_wrapper(func_name, kwargs):
...
-        # Generated method name is not a valid Python attribute name
-        # e.g. "{func_name}_1", "{func_name}_2" will be used.
+        # Generated method name is not a valid Python attribute name,
+        # i.e. "{func_name}__1", "{func_name}__2" will be used.
         return func_name
 
     if len(generated_name) > MAX_METHOD_NAME_LENGTH:
         # Generated method name is a bit too long. Simple index suffixes
-        # e.g. "{func_name}_1", "{func_name}_2" will be used.
+        # e.g. "{func_name}__1", "{func_name}__2" will be used.
```
zhenyu-ms, line 288: 👀

testplan/testing/filtering.py
kn-ms: I see your point on the single asterix case, however if that is a problem for a project they can, add the part as well to the pattern just use aterix in place of a the part numbers, I do nto expect a collide there, but if tehre are they still can change their naming, or go with multiple patterns to capture all. Figuring out a the right pattern if you do not see it for a part is not good.

testplan/testing/filtering.py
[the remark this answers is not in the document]
kn-ms: I see, than maybe not.

Though if I read right the only use of parts as integers are for the sanity check. So the most probbable path you convert the str to int do sanity check and the convert the int back to str, you can use a temp variabnle to hold the integer forms so you do not need to convert it back to str. Also not sure on the added value of this check, as the numbers my stil be invalid afterall, or is this just try to target the case when someone swap the index and length?

testplan/common/report/base.py
zhenyu-ms: tbh `match_definition` is more like `match_uid` to me

testplan/testing/filtering.py
zhenyu-ms: I want to support the use case that user want to select some specific testcases with patterns, regarding less of parts. While with the auto parts feature, there could be cases that parted multitest no longer getting parted in another run.

testplan/testing/filtering.py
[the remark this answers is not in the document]
zhenyu-ms: yeah this check is not that useful

testplan/testing/multitest/suite.py
zhenyu-ms: actually we may not need these lines

tests/functional/testplan/testing/multitest/test_multitest_parts.py
zhenyu-ms, line 1: to defer

testplan/testing/multitest/base.py
```diff
@@ -379,8 +376,19 @@ def get_test_context(self):
         ctx = []
         sorted_suites = self.cfg.test_sorter.sorted_testsuites(self.cfg.suites)
 
+        g_offset = 0
```
zhenyu-ms, line 379: to defer

doc/newsfragments/2626_changed.pattern_with_parts.rst
```diff
@@ -0,0 +1,5 @@
...
+    * Part feature has been tuned to generate more even parts in term of number of testcases.
```
zhenyu-ms, line 5: to defer

tests/functional/testplan/testing/test_listing.py
```diff
@@ -191,39 +230,53 @@ def test_programmatic_listing(
...
+            else:
+                for i in range(prim_parts):
+                    plan.add(
+                        MultiTest(
+                            name="Primary",
+                            suites=[Beta(), Alpha()],
+                            part=(i, prim_parts),
+                        )
+                    )
+            plan.add(MultiTest(name="Secondary", suites=[Gamma()]))
```
kn-ms, line 275: Could this be factored out to a factory function, which will be used to create the multitests for both testcase?

testplan/testing/multitest/suite.py
```diff
@@ -357,24 +357,14 @@ def _ensure_unique_generated_testcase_names(names, functions):
...
-                    func.__name__ not in dupe_names
-                    and func.__name__ not in valid_names
-                ):
-                    valid_names.add(func.__name__)
-                    break
-        else:
-            valid_names.add(name)
+            func.__name__ = "{}__{}".format(name, count)
+            dupe_counter[name] += 1
+        valid_names.add(func.__name__)
```
zhenyu-ms, line 362: whoops...

Pyifan (dismissed):

kn-ms (approved):

Pyifan (approved):

PR #989, opened 2023-09-06
TCPClient supports binding interface.

yuxuan-ms:

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/common/utils/sockets/client.py
+++ b/testplan/common/utils/sockets/client.py
@@ -2,6 +2,7 @@
 
 import time
 import socket
+from typing import Union, Tuple
 
 
 class Client:
@@ -16,19 +17,21 @@ class Client:
         4. close
     """
 
-    def __init__(self, host, port, interface=None):
+    def __init__(
+        self,
+        host: str,
+        port: Union[str, int],
+        interface: Union[Tuple[str, int], None] = None,
+    ) -> None:
         """
         Create a new TCP client.
         This constructor takes parameters that specify the address (host, port)
         to connect to and an optional logging callback method.
 
         :param host: hostname or IP address to connect to
-        :type host: ``str``
         :param port: port to connect to
-        :type port: ``str`` or ``int``
         :param interface: Local interface to bind to. Defaults to None, in
             which case the socket does not bind before connecting.
-        :type interface: (``str``, ``str`` or ``int``) tuple
         """
         self._input_host = host
         self._input_port = port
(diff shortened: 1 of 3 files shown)


testplan/testing/multitest/driver/tcp/client.py
```diff
@@ -53,7 +53,7 @@ class TCPClient(Driver):
         and will be expanded on runtime.
     :type port: ``int``
     :param interface: Interface to bind to.
-    :type interface: ``NoneType`` or ``tuple``(``str, ``int``)
```
M6AI, line 56: Not to be picky, but please refactor to type annotations as per usual.

M6AI: Oh I see the annotations being in constructor already. Just remove these then.

testplan/common/utils/sockets/client.py
```diff
@@ -28,7 +28,7 @@ def __init__(self, host, port, interface=None):
         :type port: ``str`` or ``int``
         :param interface: Local interface to bind to. Defaults to None, in
             which case the socket does not bind before connecting.
-        :type interface: (``str``, ``str`` or ``int``) tuple
```
M6AI, line 31: Type annotations here too please.

M6AI (approved):
Neat, thanks for the gardening on typing. Approved.

testplan/common/utils/sockets/client.py
```diff
@@ -37,49 +40,48 @@ def __init__(self, host, port, interface=None):
...
-    def port(self):
+    def port(self) -> Union[str, int]:
         return self._input_port
 
-    def connect(self):
+    def connect(self) -> None:
         """
         Connect client to socket.
         """
         self._client = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
```
Pyifan, line 57: The document says:

socket.bind(address)
Bind the socket to address. The socket must not already be bound. (The format of address depends on the address family — see above.)

A pair (host, port) is used for the [AF_INET](https://docs.python.org/3/library/socket.html#socket.AF_INET) address family, where host is a string representing either a hostname in internet domain notation like 'daring.cwi.nl' or an IPv4 address like '100.50.200.5', and port is an integer.

Because we are always create to AF_INET type of socket, i wonder if we really need to support bytes type of input.

Pyifan (changes requested):

testplan/common/utils/sockets/client.py
yuxuan-ms: fixed

Pyifan (approved):

PR #994, opened 2023-09-28
Feature/interactive display hostname

M6AI:
There is a requirement that instead of localhost or IP address we log the exact hostname on the console during interactive UI startup.

## Solution description
Simply use platform module and its node method.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -855,7 +855,8 @@ class TestRunnerIHandler(entity.Entity):
 
         self.logger.user_info(
             "\nInteractive Testplan web UI is running. Access it at:\n%s:%s/interactive",
-            platform.node(), str(port)
+            platform.node(),
+            str(port),
         )
 
     def _initial_report(self):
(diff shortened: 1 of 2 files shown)


M6AI:
Before:

Interactive Testplan web UI is running. Access it at:
    Local: http://localhost:8888/interactive/
    On Your Network: http://xx.yyy.zzz.www:8888/interactive/

After:

Interactive Testplan web UI is running. Access it at: myhost.domain.stg.com:8888/interactive

testplan/runnable/interactive/base.py
```diff
@@ -861,8 +859,9 @@ def _display_connection_info(self):
             )
 
         self.logger.user_info(
-            "\nInteractive Testplan web UI is running. Access it at:\n%s",
-            networking.format_access_urls(host, port, "/interactive/"),
+            "\nInteractive Testplan web UI is running. Access it at: %s:%s/interactive",
+            platform.node(),
```
kn-ms, line 863: platform.node might be not fully qualified:https://docs.python.org/3/library/platform.html#platform.node

M6AI: Shall I use `getfqdn` instead? https://docs.python.org/3/library/socket.html#socket.getfqdn

kn-ms (approved):

PR #995, opened 2023-09-28
Suppressed flask/flask_restx warnings

M6AI:
A couple of verbose logs of UserWarnings are emitted during the HTTP handler setup step. We would like to suppress those.

## Solution description
Locally filtering the warnings.

## Checklist:
- [ ] Test
- [x] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/runnable/interactive/base.py
+++ b/testplan/runnable/interactive/base.py
@@ -4,6 +4,7 @@ Interactive handler for TestRunner runnable class.
 import numbers
 import re
 import threading
+import warnings
 from concurrent import futures
 from typing import Union, Awaitable, Dict, Optional
 
@@ -839,7 +840,11 @@ class TestRunnerIHandler(entity.Entity):
             ihandler=self, port=self.cfg.http_port
         )
         http_handler.cfg.parent = self.cfg
-        http_handler.setup()
+
+        # Mute flask_restx warning.
+        with warnings.catch_warnings():
+            warnings.filterwarnings("ignore")
+            http_handler.setup()
 
         return http_handler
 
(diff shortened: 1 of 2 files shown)


M6AI:
Warnings that are now suppressed:

UserWarning: The setup method 'add_url_rule' can no longer be called on the blueprint 'api'. It has already been registered at least once, any changes will not be applied consistently.
Make sure all imports, decorators, functions, etc. needed to set up the blueprint are done before registering it.
This warning will become an exception in Flask 2.3.
  app_or_blueprint.add_url_rule(self._doc, "doc", self.render_doc)
  
UserWarning: The setup method 'record' can no longer be called on the blueprint 'api'. It has already been registered at least once, any changes will not be applied consistently.
Make sure all imports, decorators, functions, etc. needed to set up the blueprint are done before registering it.
This warning will become an exception in Flask 2.3.
  lambda s: s.add_url_rule(

UserWarning: The setup method 'add_url_rule' can no longer be called on the blueprint 'api'. It has already been registered at least once, any changes will not be applied consistently.
Make sure all imports, decorators, functions, etc. needed to set up the blueprint are done before registering it.
This warning will become an exception in Flask 2.3.
  app_or_blueprint.add_url_rule(self.prefix or "/", "root", self.render_root)

testplan/runnable/interactive/base.py
```diff
@@ -839,7 +840,11 @@ def _setup_http_handler(self):
             ihandler=self, port=self.cfg.http_port
         )
         http_handler.cfg.parent = self.cfg
-        http_handler.setup()
+
+        # Mute flask_restx warning.
+        with warnings.catch_warnings():
```
kn-ms, line 845: How about to put a time limit for suppression. If we would not upgrade restx in let say 6 month this should come back even with an extra annoying warning :)

kn-ms (approved):

testplan/runnable/interactive/base.py
M6AI: Let me reply internally.

PR #998, opened 2023-09-29
fix flaky interactive tests

kn-ms:
Interactive tests are flaky as some times the interactive handler did not start when the test start run. This is because of a wait with an always true predicate

## Solution description
- the wait for interactive is now uniformized
- instead of a bool(tuple) which is always true check the first item of tupple for Noneness
- plan.i => plan.interactive to make some sense

## Checklist:
- [X] Test
- [ ] Documentation (API)
- [X] MS info leakage check

### DIFF
--- a/tests/functional/testplan/runnable/interactive/__init__.py
+++ b/tests/functional/testplan/runnable/interactive/__init__.py
@@ -1,0 +1,13 @@
+from testplan.common.utils.timing import wait
+
+
+def interactive_started(plan):
+    return plan.interactive.http_handler_info[0] is not None
+
+
+def wait_for_interactive_start(plan):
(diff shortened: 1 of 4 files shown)


M6AI (approved):
LGTM. Is there a reason for removing the "i" in favor of interactive? I understand we discussed the "i" as bad practice, but not sure if we are getting rid of completely.

kn-ms:
> LGTM. Is there a reason for removing the "i" in favor of interactive? I understand we discussed the "i" as bad practice, but not sure if we are getting rid of completely.

I just removed it from tests where I was seeing it, for this quick fix, as the files were at hand anyway.

PR #1003, opened 2023-10-10
Disable parts in interactive mode, and add http to displayed interactive url

kn-ms:
A recent change make parts not disabled in interactive mode, which make it hard to work with discovered multipart multitests.
Another recent change make the interactive url to use fqdn but it is missing the http schema prefix.

## Solution description
- In interactive mode discover does not split up according parts
- the interactive url now looks a real url, so one can click on it in most editor.

## Checklist:
- [X] Test
- [ ] Documentation (API)
- [X] MS info leakage check


doc/newsfragments/2751_change.no_parts_in_interactive.rst
```diff
@@ -0,0 +1 @@
+Parts are disabled in interactive mode. This fixing a recent regression on this.
```
M6AI, line 1:
```suggestion
Parts are disabled in interactive mode to fix a recent regression on this.
```

M6AI (dismissed):

M6AI (approved):

PR #1004, opened 2023-10-13
Feature/interactive fast stop

kn-ms:
Proper startup can take minutes for a complex environment. During development when using the interactive mode devs would like to be fast on stop and restart. 

## Solution description
Now if ctrl-c is used in interactive mode the testplan and all of its child process is forecfully killed

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [X] MS info leakage check

### DIFF
--- a/doc/newsfragments/2278_new.fast_kill.rst
+++ b/doc/newsfragments/2278_new.fast_kill.rst
@@ -1,0 +1,4 @@
+In interactive mode when testplan is stopped with ctr-c, instead of properly shutting down the tests, which can take
+even minutes for a complex environment, now a fast kill approach is killing all started test processes forcefully.
+The same happens when a testplan script is interrupted by ctr-c, or with the SIGINT signal. SIGTERM will shutdown the
+test gracefully.
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -2,7 +2,7 @@
 Module containing base classes that represent object entities that can accept
 configuration, start/stop/run/abort, create results and have some state.
 """
-
+import multiprocessing
 import os
 import signal
 import sys
@@ -1705,8 +1705,14 @@ class RunnableManager(Entity):
             signum,
             threading.current_thread(),
         )
-
-        self.abort()
+        if signum == signal.SIGINT:
+            multiprocessing.Process(
+                target=fast_kill,
+                args=[multiprocessing.current_process().pid],
+                daemon=True,
+            ).start()
+        else:
+            self.abort()
 
     def pausing(self):
         """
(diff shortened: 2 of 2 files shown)


doc/newsfragments/2278_new.fast_kill.rst
```diff
@@ -0,0 +1,4 @@
+In interactive mode when testplan is stopped with ctr-c, instead of properly shutting down the tests, which can take
```
Pyifan, line 1: ctrl-c

doc/newsfragments/2278_new.fast_kill.rst
```diff
@@ -0,0 +1,4 @@
...
+even minutes for a complex environment, now a fast kill approach is killing all started test processes forcefully.
+The same happens when a testplan script is interrupted by ctr-c, or with the SIGINT signal. SIGTERM will shutdown the
```
Pyifan, line 3: This is against our common sense though, ctrl-c is generating sigint, and:

SIGINT is a gentle nudge asking the process to stop. If that fails, SIGTERM is a more forceful, but still polite request for the process to ...

Pyifan (changes requested):
I think testplan shall have 2 ways to exit:
- stop: do proper clean up and exit gracefully
- abort: no proper clean up, exit asap <-- we are implementing this here right?

Under batch mode, 
execution finish / early stop / timeout --> stop
^c --> abort 

Under interactive mode,
i don't know what shall trigger stop, maybe a new button on the ui
^c --> abort

That means instead of adding a 3rd way of fast_kill, shall we make our abort really fast and brute force instead?

doc/newsfragments/2278_new.fast_kill.rst
kn-ms: SIGINT is asking the process to interrupt on what it is doing (and most of the time wait for new inputs) SIGTERM is asking the process to stop, both of that can be handled even ignored by the app.  in python SIGINT is interpreted as keyboard interrupt, because ctrl-c fire that signal. In this approach the user action of pressing ctrl-c is interpreted as there is an inpatient someone wan to stop this interactive or batch run so just do it fast. I found it reasonable. The user story we want to enable here is mostly manual runs where the user want to stop the interactive session and restart as fast as possible, the same apply for the batch run if it is interrupted by ctrl-c the user just want to stop it as he was seeing some problem in the logs, or realized something why it need to restart it.

kn-ms:
> I think testplan shall have 2 ways to exit:
> 
> * stop: do proper clean up and exit gracefully
> * abort: no proper clean up, exit asap <-- we are implementing this here right?
> 
> Under batch mode, execution finish / early stop / timeout --> stop ^c --> abort
> 
> Under interactive mode, i don't know what shall trigger stop, maybe a new button on the ui ^c --> abort
> 
> That means instead of adding a 3rd way of fast_kill, shall we make our abort really fast and brute force instead?

there is the case when testplan is terminated with SIGTERM to do a proper abort with all cleanup is appropriate for that. (I do not like the extra button to ui, as user need to interact with two thing to stop and restart)

kn-ms:
We decided to keep the old behaviour

PR #1007, opened 2023-10-19
Changed assertion time information logic on the web UI

rnemes:
Current implementation provides misleading data about execution time of assertions leading to perhaps wrong conclusions when improving runtime. We measure assertion time as difference between entries of a testcase. In reality, it should be the execution time of the assertion step itself.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/web_ui/testing/src/Report/reportUtils.js
```diff
@@ -263,33 +264,34 @@ const getAssertions = (selectedEntries, displayTime) => {
...
-          const nextEntryTime = new Date(links[i + 1].utc_time).getTime();
+        if (links[i].utc_time && links[i - 1].utc_time) {
+          const previousEntryTime = new Date(links[i - 1].utc_time).getTime();
           const currentEntryTime = new Date(links[i].utc_time).getTime();
-          const durationMilliseconds = nextEntryTime - currentEntryTime;
+          const durationMilliseconds = currentEntryTime - previousEntryTime;
           duration = formatMilliseconds(durationMilliseconds);
         }
-        duration = "(" + duration + ")";
+        duration = "(+" + duration + ")";
```
M6AI, line 276: Is the `+` needed? Duration is always positive by definition (non-negative if precision matters and 0 is accepted after rounding/representation)

testplan/web_ui/testing/src/Report/reportUtils.js
```diff
@@ -263,33 +264,34 @@ const getAssertions = (selectedEntries, displayTime) => {
...
+            selectedEntry.timer.run.start
           ).getTime();
-          const durationInMilliseconds = nextEntryTime - currentEntryTime;
+          const currentEntryTime = new Date(links[0].utc_time).getTime();
+          const durationInMilliseconds = currentEntryTime - previousEntryTime;
           duration = formatMilliseconds(durationInMilliseconds);
         }
-        duration = "(" + duration + ")";
-        links[links.length - 1].timeInfoArray.push(duration);
+        duration = "(+" + duration + ")";
```
M6AI, line 293: Same as above.

M6AI (changes requested):
I think we can remove the +, otherwise I am good with the changes after the demo of the build.

M6AI (approved):
As discussed, we want to visualize the direction of the duration anyways.

testplan/web_ui/testing/src/Report/reportUtils.js
rnemes: As we discussed offline, "+" was used in the previous version of testplan and also helps the user understand that it is not the duration of the actual assertion block, but the time elapsed between them.

PR #1009, opened 2023-10-27
result.fix.match() assertion will use the tags present in the expected message when include_only_expected parameter set to True

rnemes:
## Solution description
result.fix.match() assertion will use the tags present in the expected message when include_only_expected parameter set to True

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/doc/newsfragments/2691_changed.fix_match_assertion.rst
+++ b/doc/newsfragments/2691_changed.fix_match_assertion.rst
@@ -1,1 +1,1 @@
-py:meth:`result.fix.match() <testplan.testing.multitest.result.FixNamespace.match>` assertion now supports repeating groups when a FIX message is specified in the ``include_tags`` or ``exclude_tags`` parameters.
+py:meth:`result.fix.match() <testplan.testing.multitest.result.FixNamespace.match>` assertion now only compares the tags present in the expected message when ``include_tags`` parameter set to `True`.
--- a/doc/newsfragments/2691_changed.fix_match_assertion.rst
+++ b/doc/newsfragments/2691_changed.fix_match_assertion.rst
@@ -1,0 +1,1 @@
+py:meth:`result.fix.match() <testplan.testing.multitest.result.FixNamespace.match>` assertion now supports repeating groups when a FixMessage object specified in the ``include_tags`` or ``exclude_tags`` parameters.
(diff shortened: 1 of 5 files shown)


testplan/testing/multitest/entries/assertions.py
```diff
@@ -17,9 +17,10 @@
 import subprocess
 import sys
 import tempfile
-
 import lxml
 
+from ms.fix import FixMessage
```
M6AI, line 22: Let us catch up offline as this can be removed I think with similar conditionality that already exists.

Pyifan (commented):
dictmatch works in a recursive way, I think the expected behavior is to only compare the tags that are present in expected msg at each level. Flatten into a list upfront will behave differently.

testplan/common/utils/comparison.py
```diff
@@ -507,7 +507,10 @@ def should_ignore_key(key: Hashable) -> bool:
         if key in ignore:
             should_ignore = True
         elif only is not None:
-            should_ignore = key not in only
+            if isinstance(only, Iterable):
```
Pyifan, line 510: I don't think we should override this existing param, that makes the code full of
 if isinstance(only, Iterable)

Pyifan (changes requested):

testplan/testing/multitest/result.py
```diff
@@ -1212,40 +1207,31 @@ def match(
...
         :return: Assertion pass status
-        :rtype: ``bool``
         """
         entry = assertions.FixMatch(
             value=actual,
             expected=expected,
             description=description,
             category=category,
-            include_tags=expected if include_tags is True else include_tags,
+            include_only_tags_of_expected=include_only_tags_of_expected,
```
M6AI, line 1233: `expected_keys_only`, `rhs_keys_only`. These are my only natural simplifications for the name at this point. Perhaps something starting with `is` due to the boolean, so say `is_rhs_keys_only`.

doc/newsfragments/2691_changed.fix_match_assertion.rst
```diff
@@ -0,0 +1 @@
+Introduced a new parameter ``include_only_tags_of_expected`` for py:meth:`result.fix.match() <testplan.testing.multitest.result.FixNamespace.match>` assertion, It will only compares the tags present in the expected message when the parameter is set to `True`.
```
Pyifan, line 1: this is a bit long...suggest a few see if there's anything you like:

include_as_expected
include_per_expected
include_only_expected
include_tags_of_expected

testplan/common/utils/comparison.py
```diff
@@ -559,11 +562,12 @@ def _rec_compare(
     lhs,
     rhs,
     ignore,
-    only,
+    include,
     key,
     report_mode,
     value_cmp_func,
     _regex_adapter=RegexAdapter,
+    include_only_keys_of_rhs=False,
```
Pyifan, line 570: include_per_rhs?

Pyifan (changes requested):

doc/newsfragments/2691_changed.fix_match_assertion.rst
M6AI: `inculude_only_expected` wins.

testplan/common/utils/comparison.py
rnemes: I am going with include_only_rhs to use the same format as in the assertion

kn-ms (approved):

Pyifan (approved):

PR #1024, opened 2023-11-15
Test Environment delayed evaluation

Pyifan:
- Make environment/dependencies/initial_context take callable
- Interactive mode start/stop_env to use defined step
- before/after_start/stop hook will append entry to report separately

## Solution description

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/examples/Driver/EnvBuilder/env_builder.py
+++ b/examples/Driver/EnvBuilder/env_builder.py
@@ -1,0 +1,64 @@
+"""
+This demonstrates one possible way of implementing of EnvBuilder class.
+"""
+import functools
+
+from testplan.common.utils.context import context
+from testplan.testing.multitest.driver.tcp import TCPClient, TCPServer
+
+
+class EnvBuilder:
+    def __init__(self, env_name: str, drivers: list):
+        """
+        :param env_name: name of this env builder
+        :param drivers: list of drivers to be created
+        """
+        self.env_name = env_name
+        self.drivers = drivers
+        self.client_auto_connect = False if len(self.drivers) == 3 else True
+        self._client1 = None
+        self._client2 = None
+        self._server1 = None
+
+    def build_env(self):
+        return [getattr(self, driver_name) for driver_name in self.drivers]
+
+    def init_ctx(self):
+        return {"env_name": self.env_name}
+
+    def build_deps(self):
+        if len(self.drivers) == 2:
+            return {self.server1: self.client1}
+        elif len(self.drivers) == 3:
+            return {self.server1: (self.client1, self.client2)}
+
+    @property
+    # @functools.cached_property
+    def client1(self):
+        if not self._client1:
+            self._client1 = TCPClient(
+                name="client1",
+                host=context("server1", "{{host}}"),
+                port=context("server1", "{{port}}"),
+                connect_at_start=self.client_auto_connect,
+            )
+        return self._client1
+
+    @property
+    # @functools.cached_property
+    def client2(self):
+        if not self._client2:
+            self._client2 = TCPClient(
+                name="client2",
+                host=context("server1", "{{host}}"),
+                port=context("server1", "{{port}}"),
+                connect_at_start=self.client_auto_connect,
+            )
+        return self._client2
+
+    @property
+    # @functools.cached_property
+    def server1(self):
+        if not self._server1:
+            self._server1 = TCPServer(name="server1")
+        return self._server1
--- a/examples/Driver/EnvBuilder/env_builder.py
+++ b/examples/Driver/EnvBuilder/env_builder.py
@@ -1,7 +1,6 @@
 """
 This demonstrates one possible way of implementing of EnvBuilder class.
 """
-import functools
 
 from testplan.common.utils.context import context
 from testplan.testing.multitest.driver.tcp import TCPClient, TCPServer
@@ -33,7 +32,6 @@ class EnvBuilder:
             return {self.server1: (self.client1, self.client2)}
 
     @property
-    # @functools.cached_property
     def client1(self):
         if not self._client1:
             self._client1 = TCPClient(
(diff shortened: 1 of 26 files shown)


testplan/runnable/interactive/base.py
```diff
@@ -489,7 +489,8 @@ def start_test_resources(self, test_uid, await_results=True):
 
         exceptions = self.test(test_uid).resources.start_exceptions
         if exceptions:
-            self._log_env_errors(test_uid, exceptions.values())
```
Pyifan, line 492: this is no longer needed as already taken care in post_step_call

testplan/testing/base.py
```diff
@@ -412,24 +511,115 @@ def start_test_resources(self):
...
+            # category=ReportCategories.SYNTHESIZED,
+        )
+        case_report = TestCaseReport(
+            name=hook.__name__,
+            suite_related=True,
+            # TODO: use synthesized instead of suite_related
+            # category=ReportCategories.SYNTHESIZED,
+        )
+        suite_report.append(case_report)
+        self.result.report.append(suite_report)
```
kn-ms, line 619: Is not the duplicated label case ned to be handled here, or thta is just needed in the real run?

If the duplicate label case is needed here as well, can this be refactored to share this logic?

testplan/testing/base.py
```diff
@@ -412,24 +511,115 @@ def start_test_resources(self):
         The base implementation is very simple but may be overridden in sub-
         classes to run additional setup pre- and post-environment start.
         """
-        self.make_runpath_dirs()
-        self.resources.start()
+        # in case this is called more than once from interactive
+        self.report.timer.clear()
+
+        self._add_step(self.setup)
```
kn-ms, line 517: These steps are nicely listed in this order in the _run_batch_steps, consider to refactor,

same for the stop_test_resources.

This would share more logic between interactive and non interactive

testplan/testing/base.py
```diff
@@ -366,10 +389,86 @@ def _record_teardown_start(self):
...
+
     def pre_resource_steps(self):
         """Runnable steps to be executed before environment starts."""
         self._add_step(self._record_setup_start)
 
+        self._add_step(self._init_context)
+        self._add_step(self._build_environment, delay=False)
+        self._add_step(self._set_dependencies)
+
+    def start_resource_steps(self):
```
kn-ms, line 442: I know this naming conforms to the rest, but it always tricks me thinking that it actually execute those steps. This is a nice to have, but what if we rename these to something enqueue_..._steps ?

testplan/testing/base.py
```diff
@@ -412,24 +511,115 @@ def start_test_resources(self):
...
         """
+
+        if not hook:
+            return
+
+        suite_report = TestGroupReport(
+            name=label,
+            category=ReportCategories.TESTSUITE,
+            suite_related=True,
+            # TODO: use synthesized instead of suite_related
```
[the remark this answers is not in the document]
Pyifan: no, this is a bit too much, I will create a ticket for this.

testplan/testing/base.py
Pyifan: but _run_batch_steps goes all the way to test execution, here we just need the steps to start/stop_test_resources.

testplan/testing/base.py
Pyifan: agree

testplan/testing/base.py
Pyifan: the dry_run always start with a         

self.result.report = self._new_test_report()

thus we won't have dup label issue - which only happens in interactive mode.

testplan/testing/base.py
kn-ms: yes but the batch steps is really compsed of start resources then run batch step than stop resources, so _run_batch_steps can reuse these guys, instead of trying to keep them in sync

kn-ms:
Please do not ammend - force push (at least not do that after review started) it is really hard to see what is changing.

testplan/testing/base.py
```diff
@@ -366,19 +389,95 @@ def _record_teardown_start(self):
...
-    def pre_resource_steps(self):
+    def _init_context(self):
+        if callable(self.cfg.initial_context):
+            self.resources._initial_context = self.cfg.initial_context()
+        else:
+            self.resources._initial_context = self.cfg.initial_context
+
+    def _build_environment(self, delay: bool):
+        """
+        This method will execute at most twice, once with delay=True during __init__,
```
kn-ms, line 400: I do not like this two phased approach, where we prepare the function that it will be run twice. Why do we need to do the driver line up in init in case we do not have a callable?

I would turn the list during params validation into a callable and just handle the callable case in teh business logic.

yuxuan-ms (approved):

testplan/testing/base.py
Pyifan: 1) the 2 phase approach... is because I don't want to lose any existing test coverage for interactive mode that asserts environment status before test starts

2) I know this turn the list into callable approach, just i would prefer not to alter user input, but save it as-is in self.cfg.xxx

kn-ms: I can buy your second point. I do not like the excuse of leaving complication in code to not break existing tests. Would it be to tedious to update those tests?

examples/Driver/EnvBuilder/env_builder.py
```diff
@@ -0,0 +1,64 @@
...
+        return {"env_name": self.env_name}
+
+    def build_deps(self):
+        if len(self.drivers) == 2:
+            return {self.server1: self.client1}
+        elif len(self.drivers) == 3:
+            return {self.server1: (self.client1, self.client2)}
+
+    @property
+    # @functools.cached_property
```
zhenyu-ms, line 36: also here

yuxuan-ms (dismissed):

kn-ms (dismissed):

zhenyu-ms (approved):

PR #1027, opened 2023-11-22
Skip exporters invocation when report is empty.

yuxuan-ms:

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/runnable/base.py
```diff
@@ -1250,6 +1250,11 @@ def _invoke_exporters(self) -> None:
         # Add this logic into a ReportExporter(Runnable)
         # that will return a result containing errors
 
+        if isinstance(self._result.test_report, TestReport) and (
```
Pyifan, line 1253: Add an is_empty method to some report base class? And check both entries and logs (maybe something else too but I cannot think of right now)

Pyifan (changes requested):

Pyifan (approved):

PR #1029, opened 2023-12-05
Feature/skip remaining tests being executed 

zhenyu-ms:

## Solution description

## Checklist:
- [x] Test
- [x] Documentation (API)
- [x] MS info leakage check
- [x] For new cmdline arg: documentation


testplan/runners/local.py
```diff
@@ -75,9 +94,20 @@ def _loop(self) -> None:
...
-                        self._results[next_uid] = result
                     finally:
-                        self.ongoing.pop(0)
+                        with self._curr_runnable_lock:
+                            if not self._to_skip_remaining:
+                                # otherwise result from aborted test included
+                                self._results[next_uid] = result
+                                self.ongoing.pop(0)
+
+                    if thres and result.report.status <= thres:
```
Pyifan, line 104: I think we can just call a method of TestRunner here to set the flag, something like

if self.cfg.stop_strategy.should_stop_plan(result):
self.parent.stop_plan()

Then TestRunner shall distribute the should_stop flag to all executors (including this one).

Executors just discard all pending_work and pretend we have finished, and triggers normal stop workflow.

testplan/runners/pools/base.py
Pyifan, line 1: I think pool size code change shall be in task result handler, every time pool receives a task result, check if we shall stop plan. Then the handling logic shall be exactly as testplan timeout handler.

The nested child pool on the remote side needs some special handing, need to disable this feature.

Pyifan (changes requested):

testplan/runners/base.py
```diff
@@ -83,7 +97,34 @@ def get(self, uid: str) -> object:
...
+            except Empty:
+                pass
+            else:
+                self.logger.debug(
+                    "Executor[%s] receiving %s.", self.uid(), msg.mark
+                )
+                self._msg_handlers[msg.mark](msg)
+            time.sleep(interval)
+
+    def _handle_expected_abort(self, _):
```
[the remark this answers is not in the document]
Pyifan: 2nd thought... stop_plan is going to discard all on-going tests, so maybe it's good enough to just reuse the timeout workflow, which is essentially abort().

testplan/common/utils/logger.py
```diff
@@ -178,6 +185,10 @@ def configure_file_logger(level, runpath):
         return file_handler
 
 
+def get_config_logger(config_class: Type) -> TestplanLogger:
```
zhenyu-ms, line 188: to remove

testplan/runners/pools/base.py
```diff
@@ -95,6 +98,7 @@ def __init__(self, **options) -> None:
         self.event_recorder = EventRecorder(
             name=str(self), event_type="Worker"
         )
+        self._discard_running = threading.Event()
```
Pyifan, line 101: What is the difference of using threading.Event() v.s boolean.

And _discard_running v.s _discard_pending?

testplan/runners/pools/base.py
```diff
@@ -213,11 +227,22 @@ def _loop(self, transport: QueueClient) -> None:
             elif received.cmd == Message.TaskSending:
                 results = []
                 for item in received.data:
-                    results.append(self.execute(item))
-                transport.send_and_receive(
+                    if self._discard_running.is_set():
+                        break
+
+                    task_result = self.execute(item)
+                    if not self._discard_running.is_set():
```
Pyifan, line 234: I kind of don't like this exception. Can we have execute always return the result?

testplan/runners/pools/base.py
```diff
@@ -322,10 +356,10 @@ def __init__(
         self._task_retries_cnt = {}  # uid: times_reassigned_without_result
         self._task_retries_limit = 2
         self._workers = entity.Environment(parent=self)
-        self._workers_last_result = {}
         self._conn = self.CONN_MANAGER()
         self._conn.parent = self
         self._pool_lock = threading.Lock()
+        self._a_pool_lock = threading.RLock()
```
Pyifan, line 362: this needs a better name

Pyifan (changes requested):

testplan/runnable/base.py
```diff
@@ -890,10 +900,11 @@ def _register_task(self, resource, target, uid, metadata):
         self.resources[resource].add(target, uid)
 
     def _collect_task_info(self, target: TestTask) -> TaskInformation:
+        if callable(target):
+            target = Task(target)
```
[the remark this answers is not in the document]
zhenyu-ms: inconsistent "add" method behaviour between pools & local runner, will leave comments there and fix in the future

testplan/runnable/base.py
```diff
@@ -912,6 +915,7 @@ def _collect_task_info(self, target: TestTask) -> TaskInformation:
                 "Unrecognized test target of type {}".format(type(target))
             )
 
+        # TODO: include executor in ancestor chain?
```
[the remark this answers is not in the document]
zhenyu-ms: currently parents of Test is TestRunner instead of Executor

actually a "NOTE" might be more appropriate

testplan/runners/pools/base.py
zhenyu-ms: in this simple use case i think they could be equivalent even in the GIL-free world, `threading.Event()` might only exist as a reminder

i tend to choose different variable names for the ease of grep

testplan/runners/pools/connection.py
```diff
@@ -13,6 +13,9 @@
 from testplan.common.utils import logger
 from testplan.runners.pools.communication import Message
 
+if TYPE_CHECKING:
+    from testplan.runners.pools.base import Worker
```
zhenyu-ms, line 17: previously missing, see L291

testplan/runners/pools/base.py
zhenyu-ms: maybe we can make a better usage of TaskResult.status, currently it's just a boolean flag

testplan/common/entity/base.py
```diff
@@ -1000,6 +1008,7 @@ def _post_run_checks(self, start_threads, start_procs):
         :param start_procs: processes before run
         :type start_procs: ``list`` of ``Process``
         """
+        # XXX: do we want to suppress process/thread check for tests?
```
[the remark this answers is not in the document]
zhenyu-ms: suppression is from intuition, use XXX instead of FIXME here

doc/newsfragments/2540_new.skip_remaining.rst
```diff
@@ -0,0 +1,3 @@
+Added a new command line argument ``--skip-remaining``, a new argument ``skip_strategy`` to MultiTest, allowing certain test execution being skipped for a better user experience.
```
Pyifan, line 1: allow early return of test execution and skip remaining Testcases/Testsuites/Multitests being executed when a Testcase has failed or raised exception.

Pyifan (dismissed):

Pyifan (approved):

PR #1033, opened 2023-12-13
Disable interactive action buttons when a request is pending

rnemes:
UI controls depend on the runtime status and the user is exposed to both UX impact and possibly undesired functionality when the backend is delayed in updating the report.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


Pyifan (dismissed):

testplan/web_ui/testing/src/Nav/InteractiveNavEntry.js
```diff
@@ -250,14 +261,20 @@ const getEnvStatusIcon = (entryStatus, envStatus, envCtrlCallback) => {
...
+            disabled ? ignoreClickEvent : (e) => {
+              setPendingEnvRequest("STARTING");
+              envCtrlCallback(e, "start");
+            }
           }
         />
       );
 
     case "STOPPING":
+      if (pendingEnvRequest === "STOPPING") {
```
yuxuan-ms, line 275: How about define the value in a object as an enum in Common/defaults.js?

yuxuan-ms (approved):

kn-ms (approved):

PR #1034, opened 2023-12-19
Separating report_target application in interactive mode

M6AI:
The report_target decorator is applied in test_context every time the interactive mode accesses it. This results in an overhead during startup, and also runtime.

## Solution description
For now we are moving to the runner method with report_target in interactive mode. Longer term we need to make reapplication dependent on the reload.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/testing/multitest/base.py
```diff
@@ -397,7 +397,11 @@ def get_test_context(self):
             ) < len(sorted_testcases):
                 testcases_to_run = sorted_testcases
 
-            if self.cfg.testcase_report_target:
+            if self.cfg.testcase_report_target and not (
+                getattr(self, "parent", None)
+                and hasattr(self.parent, "cfg")
+                and self.parent.cfg.interactive_port is not None
+            ):
```
M6AI, line 404: Copied over the condition so that possible refactorization catches both occurrences, there should be better ways to test interactivity.

Pyifan:
I suggest that we address the performance issues more thoroughly, there are 3 of them:

1) we are calling get_test_context again and again under interactive mode --> suggest that we invalidate self._test_context when reload() is called, then next time test_context is needed, it will re-calculate. I think this only needs a few lines of change.
2) testcase_report_target is default on -- this feature is quite heavy on runtime and report size, consider make it default off?
3) report_target is calculated in get_test_context so in interactive mode, we will do that for all testcases no matter if they are run or not. You have addressed this one by move it to _run_testcases_iter, why not move it to _run_testcase so it is shared by batch & interactive mode.

M6AI:
> I suggest that we address the performance issues more thoroughly, there are 3 of them:
> 
> 1. we are calling get_test_context again and again under interactive mode --> suggest that we invalidate self._test_context when reload() is called, then next time test_context is needed, it will re-calculate. I think this only needs a few lines of change (set test._test_context to None in reload()).
> 2. testcase_report_target is default on -- this feature is quite heavy on runtime and report size, consider make it default off?
> 3. report_target is calculated in get_test_context so in interactive mode, we will do that for all testcases no matter if they are run or not. You have addressed this one by move it to _run_testcases_iter, why not move it to _run_testcase so it is shared by batch & interactive mode?

Except for 2 I agree. The default on came from business requirement, let me reach you offline with that.

testplan/runnable/interactive/base.py
```diff
@@ -726,71 +726,6 @@ def all_tests_operation(self, operation, await_results=True):
...
-    ):
-        # Placeholder for function to delele an existing and registering a new
-        # environment resource with probably altered source code.
-        # This should access the already added Environment to plan.
-        pass
-
-    def add_created_environment(self, env_uid):
-        """Add an environment from the created environment maker instance."""
-        self.target.add_environment(self._created_environments[env_uid])
-
```
M6AI, line 793: All kinds of unused functions, we can restore from history if needed at some point.

testplan/runnable/interactive/base.py
```diff
@@ -816,13 +751,13 @@ def reload_report(self):
...
                                     ]
-                                except KeyError:
+                                except (KeyError, IndexError):
                                     continue
                         else:
                             new_report[multitest.uid][suite.uid].entries[
                                 case_index
                             ] = suite[case.uid]
-                    except KeyError:
+                    except (KeyError, IndexError):
```
M6AI, line 760: We checked and this was breaking testcase removal, the reloader would not handle the IndexError that is thrown by lists, we only continued here originally upon KeyError.

testplan/testing/multitest/base.py
```diff
@@ -467,7 +454,7 @@ def get_test_context(self):
                             testcase_instance, None
                         )
                         if data is not None:
-                            testcase.__xfail__ = {
+                            testcase.__func__.__xfail__ = {
```
M6AI, line 457: There was a side effect from report_target, it was turning the bound method into a free function, but here we need to access now the __func__ attribute.

PR #1039, opened 2023-12-30
Information metadata on Interactive UI

M6AI:
Interactive UI does not receive information metadata during serialization.

## Solution description
Added information metadata to the interactive UI.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/runnable/interactive/base.py
```diff
@@ -122,6 +122,7 @@ def http_handler_info(self):
     def setup(self):
         """Set up the task pool and HTTP handler."""
         self.target.make_runpath_dirs()
+        self.report.information.append(("runpath", self.target.runpath))
```
Pyifan, line 125: Suggest move this to TestRunner define_runpath()

M6AI: Tested moving it, it is not sufficient. The trick is that the interactive handler derives the runpath to its own report (created with _initial_report call) from the target. It does not really know any other way. So even if we move to TestRunner that just means that those objects will know about their runpath in information field, but the interactive handler works with a different report at top level.

Pyifan: I see. ihandler has its own top level report rather than using the report of the target (TestRunner).

btw, the report will be regenerated when we reload, I doubt we will lost the runpath info.

M6AI: Info is persisted upon a check.

Pyifan (approved):

PR #1046, opened 2024-01-15
Feature/multipart listing

kn-ms:
Multipart multitests are generated runtime, not import time. 

## Solution description
Now the json lister using test context to collect metadata.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [X] MS info leakage check


doc/newsfragments/2617_change.multipart_json_listing.rst
```diff
@@ -0,0 +1 @@
+`--info json` now working well with multipart multitests.
```
M6AI, line 1: I know it is me always on the writing, but I think it is better:
```suggestion
`--info json` is now working well with multitest multiparts.
```

testplan/testing/multitest/base.py
```diff
@@ -603,10 +608,19 @@ def aborting(self):
         """Suppressing not implemented debug log from parent class."""
 
     def get_metadata(self) -> TestMetadata:
+
+        suites = []
+        for suite, testcases in self.get_test_context():
```
M6AI, line 613: `test_context` is property, inherited through `Test`, I would call it here because that would populate `_test_context` and no need to recalculate it again.

M6AI (changes requested):

M6AI (approved):

PR #1050, opened 2024-01-28
Add error_handler hook to MultiTest

rnemes:
Add an error handler hook to Multitest which will be executed upon an exception.

## Checklist:
- [x] Test
- [x] Documentation (API)
- [x] MS info leakage check


Pyifan:
I would consider implementing this by adding a error handling step to Runnable.
Check the _run method of Runnable and skip_step method of Multitest.

Pyifan:
The error_handler, is something similar to before_start - it's a ConfigOption that we provide at Test level that allows user to define what to do if anything goes wrong. We don't need to know what goes on in it, all up to user. On the other hand, testplan code controls the execution flow - that is when this error_handler shall be invoked.  The _run() and skip_step() things are the centralized place to control the execution flow, thus my suggestion. We can take this discussion offline.

M6AI:
> The error_handler, is something similar to before_start - it's a ConfigOption that we provide at Test level that allows user to define what to do if anything goes wrong. We don't need to know what goes on in it, all up to user. On the other hand, testplan code controls the execution flow - that is when this error_handler shall be invoked. The _run() and skip_step() things are the centralized place to control the execution flow, thus my suggestion. We can take this discussion offline.

That makes sense. We have already reached agreement offline with @kn-ms and @rnemes, my concern was the Runnable level, but here you mentione Test which is fine by me as that was the second level in my suggestion too.

testplan/testing/base.py
```diff
@@ -518,31 +519,18 @@ def _get_hook_context(self, case_report):
             case_report.logged_exceptions(),
         )
 
-    def _run_resource_hook(self, hook, label):
```
Pyifan, line 521: _run_resource_hook is step that we add to runnable execution flow, please don't rename it to _prepare_suite_report, which is not as clear about what we are doing.

    def add_start_resource_steps(self):
        self._add_step(
            self._run_resource_hook,
            hook=self.cfg.before_start,
            label="Before Start",
        )

        self._add_step(self.resources.start)

        self._add_step(
            self._run_resource_hook,
            hook=self.cfg.after_start,
            label="After Start",
        )


You can add a separate error_handler step to steps, which will be skips if no error happened in previous steps.

rnemes: I haven't renamed it, it's there, please see Line 600. Github is not so intelligent when one moves sections of code...

rnemes: I only needed a few lines of codes from it and I reorganized them to another function to prevent code duplication.

testplan/testing/multitest/base.py
```diff
@@ -802,9 +806,25 @@ def _execute_step(self, step, *args, **kwargs):
...
+                if (
+                    self.resources.start_exceptions
+                    or self.resources.stop_exceptions
+                ):
+                    self.logger.critical(
+                        'Found exceptions during "%s" step', step.__name__
+                    )
+                    if self.cfg.error_handler:
+                        self.logger.critical("Running error handler!")
+                        self._run_error_handler(self.cfg.error_handler, step)
```
kn-ms, line 818: _run_error_handler is always running the self.cfg.error_handler, no need to get it as a parameter. You might be move the test of a valid error handfler and the logging into the function.

testplan/testing/multitest/base.py
```diff
@@ -802,9 +806,25 @@ def _execute_step(self, step, *args, **kwargs):
         with self.report.logged_exceptions():
```
Pyifan, line 806: For multitest, all exceptions happened during executing steps are captured in self.report right?
Even for environment start/stop, the exception first goes to self.resources.start_exceptions/stop_exceptions, are collected info self.report in post_step_call?
IMO, we can look at self.report.status == error as a simpler indicator of whether we shall invoke error_handler.

Also when to invoke the error_handler is something we shall consider - right after the exceptioned step or as a final safe net after other major steps. For example, if after_start exceptioned, shall we call error_handler immediately or after stop? 

The existing code is controlling the runnable execution flow with step definition, thus a if-else branch inside _execution_step is breaking style. The steps mechanism meant to avoid such spaghetti code. You can move this to post_step_call.

Another thing is, I feel this new config option and related logic shall be promoted to Test class.

examples/Best Practice/Common Utilities/test_plan.py
```diff
@@ -83,6 +110,7 @@ def main(plan):
             environment=[App("echo", binary="/bin/echo", args=["testplan"])],
             before_start=before_start_fn,
             after_stop=after_stop_fn,
+            error_handler=error_handler_fn,
```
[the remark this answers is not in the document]
rnemes: No, it doesn't. I left a comment in the function, that one should raise in one of the other hooks to get invoked.

testplan/testing/base.py
```diff
@@ -111,6 +111,9 @@ def get_options(cls):
             ConfigOption("after_start", default=None): start_stop_signature,
             ConfigOption("before_stop", default=None): start_stop_signature,
             ConfigOption("after_stop", default=None): start_stop_signature,
+            ConfigOption("error_handler", default=None): Or(
```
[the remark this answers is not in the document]
rnemes: ohh, nice catch!

examples/Best Practice/Common Utilities/test_plan.py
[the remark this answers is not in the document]
rnemes: I added a separate example for error_handler. See below.

testplan/testing/multitest/base.py
```diff
@@ -572,9 +572,19 @@ def get_tags_index(self):
...
-        if step in (
+        if step == self._run_error_handler:
+            if (
+                self.resources.start_exceptions
+                or self.resources.stop_exceptions
+                or self._get_error_logs()
+            ):
+                return False
+            else:
+                self.logger.critical("Skipping error handler")
```
zhenyu-ms, line 585: actually will this line being triggered in most cases? (testing passed or error handler not present)

rnemes: You are right, log has been removed!

zhenyu-ms (approved):

PR #1055, opened 2024-02-16
New timer field for resource monitor and multitest parts merge

Pyifan:
This change is to consolidate timer and event_recorder and get ready for resource monitor.
It is also useful for multitest parts merge.

## Solution description
1) change timer field from Dict[str, Interval] to Dict[str, List[Interval]]
2) make timer available to (almost) all entities via a recursive link to parent's report
3) remove EventRecorder
4) re-org Report class hierarchy
5) re-org ReportSchema class hierarchy
6) ui change and test
7) news frag

## Checklist:
- [X] Test
- [ ] Documentation (API)
- [X] MS info leakage check


testplan/common/utils/timing.py
```diff
@@ -320,30 +327,35 @@ def record(self, key):
             >>> with timer.record('my-key'):
             >>>  ... custom code ...
             >>>  ... custom code ...
-            >>> timer['my-key'].elapsed
+            >>> timer['my-key'][-1].elapsed
```
kn-ms, line 330: What is the usecase to have more than one timer for a key? Most of the code elsewhere I see either the first item, or (most cases) the last item is used only.

If the last and the first is so significan, can those guys have some dedicated accessors

like Timer.last("key"), Timer.first("key")

kn-ms (commented):
given that timer part get a redesign here, I suggest to reconsider a less deep coupling instead of build a reverse coupling on the parent hierarchy.

Timers nicely translate to events in my thinking, but not only timers are events that might be interesting. What if we replace the timer with events emitted to an even queue, and listeners on event queue can translate events into timing information. And later these listeners can contribute data to the report. Events can be used to track progress as as well, and the nice thing is that the logic would be in listeners concentrate on one task (progress indication, timing) not buried into the entity hierarchies. Events would be also something  that can be streamed from child processes easily, so maybe not even merging of timing would be needed , but there can be a single listener on master testplan side that could collect information. And of course if someone do not need this kind of information, it can be just easily turned off not creating the listener at all.

testplan/common/report/schemas.py
```diff
@@ -6,14 +6,58 @@
...
+            {
+                k: [IntervalSchema().load(v) for v in l]
+                for k, l in value.items()
+            }
+        )
+
+
+class ChildSchema(Schema):
+    """
+    Field for serializing ``timer.Timer`` objects, which is a ``dict``
```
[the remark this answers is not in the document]
Pyifan: Yes, with `timer` I really meant `event`, I'm just re-using the name for backward compat, but we can change everything to event as well.

Currently the timer class is still a bit rough in terms of API, I will need to give it a bit more thoughts and design.

However, I kind of liked it that the timing data is embedded in the report hierarchy, so that we don't need to re-build the hierarchy somewhere else.

We can introduce logging level concept to timer recording later so that we can turn on/off with some config.

We can also have pub-sub for realtime update by defining a super-shallow schema of the report and only serialize e.g name & timer.

I see a potential of having this reverse coupling on the parent hierarchy, e.g for including driver information into multitest report.

testplan/runnable/interactive/base.py
```diff
@@ -96,6 +96,10 @@ def __call__(self, *args, **kwargs):
         finally:
             self.teardown()
 
+    @property
+    def report(self):
```
zhenyu-ms, line 100: well it's conflicting with #1062

zhenyu-ms (approved):
LGTM

PR #1062, opened 2024-03-12
Fix/reload report

zhenyu-ms:
Report appeared broken after reloading test code with newly added testcases.

## Solution description
Instead of grafting, copying certain information from the previous report to the new report here.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/common/report/base.py
```diff
@@ -348,9 +348,14 @@ def set_by_uid(self, uid, item):
...
             self.append(item)
 
-    def __setitem__(self, uid, item):
-        """Shortcut to `set_by_uid()` method via [] operator."""
-        self.set_by_uid(uid, item)
+    __setitem__ = set_by_uid
+
+    def remove_by_uid(self, uid):
+        self.entries.pop(self._index[uid])
+        del self._index[uid]
```
kn-ms, line 355: if you rebuild the whole index, this is pointless

tests/functional/testplan/runnable/interactive/test_reloader.py
kn-ms, line 1: I guess the idea to having the before after state in a single file is based on the need to easily see the differences when one working with them.

I think clubbing them together in a single file does not serve that as if they would be in two different I could easily diff them, or just open side by side and it would be visual. Also the assertions are anyway in a different file so you need to open more than one when working on this testcase.

Having them two separate .py file has some added value:
  - we can diff or see the diff between before after easily
  - you do not need the split logic, which is ugly and do not document that well the secret that you need exactly 80 # as separator
  - you will get black formatting so the code in it conforms the rest.
  - if we refactoring something used in these examples they will be automatically refactored as well

tests/functional/testplan/runnable/interactive/test_reloader.py
```diff
@@ -614,3 +621,61 @@ def test_reload():
...
+                path=Path(f.name).parent, name_pattern=Path(f.name).name
+            )
+            irunner = base.TestRunnerIHandler(runner)
+            irunner.run_all_tests()
+
+            prev_assertions(irunner.report)
+
+            f = open(f.name, "w+")
+            f.seek(0)
+            f.writelines(curr_lines)
```
kn-ms, line 673: This may leak a file descriptor if writelines throwing an exception. See my other comment and just use simple copy over.

kn-ms (changes requested):
Can you fill the template. It is not clear what this pr is fixing, I feel that this also can have an newwfragment

tests/functional/testplan/runnable/interactive/test_reloader.py
```diff
@@ -623,18 +624,9 @@ def test_reload():
...
-    with open(
-        Path(__file__).parent / "reload_cases" / f"case_{case_num}", "r"
-    ) as f:
-        for l in f.readlines():
-            if l.strip() == "#" * 80:
-                o += 1
-                continue
-            r[o] += l
-    return r
+def load_case_paths(case_num):
```
kn-ms, line 627: this actually not loading anything, how about  `get_testcase_paths` ?

kn-ms (dismissed):

testplan/runnable/interactive/base.py
```diff
@@ -736,30 +737,37 @@ def reload(self, rebuild_dependencies=False):
...
-                                        case.uid
-                                    ].entries[param_index] = case[
-                                        param_case.uid
-                                    ]
-                                except (KeyError, IndexError):
-                                    continue
-                        else:
-                            new_report[multitest.uid][suite.uid].entries[
-                                case_index
-                            ] = suite[case.uid]
```
Pyifan, line 759: I think this is where the original problem is right? that we assume case of a particular uid is still at the same index?

Pyifan: Can we just fix this bit?

testplan/runnable/interactive/base.py
```diff
@@ -736,30 +737,37 @@ def reload(self, rebuild_dependencies=False):
...
+                    _preserve_partial_prev(prev[uid], curr[uid])
+
+            elif isinstance(prev, TestCaseReport) and isinstance(
+                curr, TestCaseReport
+            ):
+                curr.timer = prev.timer
+                curr.runtime_status = prev.runtime_status
+                curr.status = prev.status
+                curr.attachments = prev.attachments
+                curr.status_override = prev.status_override
```
Pyifan, line 760: Copying selected entries from prev to curr isn't very bulletproof, how do we know these are everything we need to copy? or what if we add new fields to report, we will forget to come back here to update.

Pyifan (changes requested):

Pyifan (approved):

PR #1064, opened 2024-03-13
Feature/logfile assertions

zhenyu-ms:
new testplan assertions, under namespace `logfile`, providing easy-to-use combinations of current `LogMatcher` apis; also new api `expect` to `LogMatcher`

## Checklist:
- [x] Test
- [x] Documentation (API)
- [x] MS info leakage check


testplan/common/utils/match.py
```diff
@@ -278,13 +297,13 @@ def not_match(
...
-        regex: Union[str, bytes, Pattern],
+        regex: Regex,
         timeout: float = LOG_MATCHER_DEFAULT_TIMEOUT,
         raise_on_timeout: bool = True,
     ) -> List[Match]:
         """
-        Similar to match, but returns all occurrences of regex. Can be
-        configured to raise an exception if no match is found.
+        Similar to match, but returns all occurrences of regex. By default an
+        an exception is raised if no match is found.
```
kn-ms, line 306: type: remove one of the an

testplan/common/utils/match.py
```diff
@@ -400,3 +413,63 @@ def get_between(self, mark1=None, mark2=None):
                 lines_between.append(line)
             separator = b"" if self.binary else ""
             return separator.join(lines_between)
+
+    @contextmanager
+    def expect(
```
kn-ms, line 418: I would do this the other way around, not the logmatcher know about the existence of a result, but maybe the result knows something about the logmatcher and just use it's api

insteadl of 

width log_matcher expect(resutl,...)

I would do 

result.exepct_logs(logmatcher,...)

here the intent is more apparent, that you do an assertion. Not passing a result object to a supposed to be simple utility class that does log matching.

testplan/common/utils/match.py
```diff
@@ -400,3 +413,63 @@ def get_between(self, mark1=None, mark2=None):
...
+        elif len(timeout) != len(regex):
+            raise ValueError(
+                "``timeout`` length doesn't match ``regex`` length."
+            )
+        pairs = zip(regex, timeout)
+
+        yield
+
+        for r, t in pairs:
+            m = self.match(r, t, raise_on_timeout=False)
```
kn-ms, line 458: the first match might not run from s_pos if the log matcher was used within the context manager

kn-ms (changes requested):
can you fill in or just get rid of the description template. it looks ugly in this form.

can we have test/example or both?

zhenyu-ms:
> can you fill in or just get rid of the description template. it looks ugly in this form.
> 
> can we have test/example or both?

sure thing, will definitely add them in the next commit

testplan/common/utils/match.py
zhenyu-ms: right... i need to think it over 👍

testplan/common/utils/match.py
Pyifan: To say the least, I don't want our user to have 2 options doing the same thing, lol.

They used to do driver.log_matcher.match(), it's easier adapt the mind to just add the new functionality as driver.log_matcher.expect_logs() - rather than if you do vanilla log match, do it from driver; else if you need a context manager, do it with result, this complicates things.

I think result can probably passed to the context manager as an optional parameter. Not very mind clean I know...

kn-ms: You are right they using driver.log_matcher.match() and then they are doing an assertion with the result if they need it, and in the form and with description as they see fit. So in my view having it in assertion is much closer to current usecase.

Pyifan (dismissed):

doc/en/download/Assertions.rst
```diff
@@ -18,16 +18,16 @@ test_plan_basic.py
...
   - :download:`test_plan_group.py <../../../examples/Assertions/Basic/test_plan_group.py>`
 
-test_plan_skip.py
+test_plan_group.py
 ++++++++++++++++++
-.. literalinclude:: ../../../examples/Assertions/Basic/test_plan_skip.py
+.. literalinclude:: ../../../examples/Assertions/Basic/test_plan_group.py
 
 Required files:
   - :download:`test_plan_exception.py <../../../examples/Assertions/Basic/test_plan_skip.py>`
```
Pyifan, line 26: mismatch

kn-ms (commented):
One more thing, I'm not sure that the log position in display has any added value. By python docs it is not really defined what it is, it should be treated as an opaque value. It  is different on windows on linux, and I guess it might change in the future. Even the current impl is not useful as it is kind of bytes from the beginning, it is not easy to find the position at the logfile.

The inode in case of rotated log is again something that you could not use from the report unless you have the exact runtime logs on the system it was created. As soon as it is attached to the report, run on remote host that is basically useless.

zhenyu-ms:
> One more thing, I'm not sure that the log position in display has any added value. By python docs it is not really defined what it is, it should be treated as an opaque value. It is different on windows on linux, and I guess it might change in the future. Even the current impl is not useful as it is kind of bytes from the beginning, it is not easy to find the position at the logfile.
> 
> The inode in case of rotated log is again something that you could not use from the report unless you have the exact runtime logs on the system it was created. As soon as it is attached to the report, run on remote host that is basically useless.

yeah the inode is only useful when tests executed on the same host, I can get rid of it. the log position info is for debugging purpose such that when certain testcases failed user can have more information to reconstruct the situation (and maybe ensure that the testing framework is working properly)

testplan/common/utils/match.py
```diff
@@ -70,6 +87,58 @@ def match_regexps_in_file(
...
+            if m is not None:
+                self.match_results.append((m, r, t, s_pos, e_pos))
+            else:
+                self.match_failure = (None, r, t, s_pos, e_pos)
+                break
+
+            if self.strict_order:
+                s_pos = e_pos
+            else:
+                self.log_matcher.position = s_pos
```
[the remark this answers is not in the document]
zhenyu-ms: ah yes... i will have position set to the e_pos of the last match (whether succeeded or failed). mark & seek needs a proper name for the mark, i think direct usage of position won't cause much trouble, after all it's our own code

testplan/testing/result.py
```diff
@@ -1341,6 +1348,188 @@ def log(self, msg, description=None):
...
+
+
+class LogfileNamespace(AssertionNamespace):
+    """
+    Contains assertion methods that operates on log files equipped with
+    :py:class:`~testplan.common.utils.match.LogMatcher`.
+    """
+
+    @assertion
+    def seek_eof(
```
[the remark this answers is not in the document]
zhenyu-ms: yeah this assertion isn't quite useful, i think it's okay to leave it there though

Pyifan (approved):

kn-ms (approved):

kn-ms (approved):

PR #1087, opened 2024-05-23
do not propagate to root logger from report.logger

kn-ms:
The report logger is a child of the root logger currently. Normally the root logger does not have any handler in tp runs, so any log meant to the report is not logged to the logfile/stdout. 

It could happen that a user or other libraries sets up root logger to emmmit logs to stdout/stderr so report logs appear there as well, causing noise.

## Solution description
Do not propagate from report logger to the root logger

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [X] MS info leakage check


yuxuan-ms (dismissed):

doc/newsfragments/2882_change.no_report_log_propagation.rst
zhenyu-ms, line 1: file name should be 2882_**changed**.no_report_log_propagation.rst

Pyifan (approved):

PR #1099, opened 2024-06-06
Make install_files invoked automatically for Driver object startup

rnemes:

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/driver/app.py
+++ b/testplan/testing/multitest/driver/app.py
@@ -310,20 +310,6 @@ class App(Driver):
         """
         return socket.gethostname()
 
-    def pre_start(self) -> None:
-        """
-        Create mandatory directories and install files from given templates
-        using the drivers context before starting the application binary.
-        """
-        super(App, self).pre_start()
-
-        self._make_dirs()
-        makedirs(self.app_path)
-        self.std = StdFiles(self.app_path)
-
-        if self.cfg.install_files:
-            self.install_files()
-
     def starting(self) -> None:
         """Starts the application binary."""
         super(App, self).starting()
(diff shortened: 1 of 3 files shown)


testplan/testing/multitest/driver/app.py
```diff
@@ -411,13 +397,20 @@ def stopping(self) -> None:
             )
             raise RuntimeError(err_msg)
 
-    def _make_dirs(self) -> None:
```
kn-ms, line 414: JavaApp  override _make_dirs, you loose that functionality here.

testplan/testing/multitest/driver/app.py
```diff
@@ -411,13 +397,20 @@ def stopping(self) -> None:
...
+        super(App, self).make_runpath_dirs()
+
         bin_dir = os.path.join(self.runpath, "bin")
         etc_dir = os.path.join(self.runpath, "etc")
-        for directory in (bin_dir, etc_dir):
+        for directory in (bin_dir, etc_dir, self.app_path):
             makedirs(directory)
         self._binpath = bin_dir
         self._etcpath = etc_dir
+        self.std = StdFiles(self.app_path)
```
kn-ms, line 413: Std files are not dirs,  it might could go back to pre_start

kn-ms (changes requested):

testplan/testing/multitest/driver/app.py
rnemes: That is true, but we have a circular dependency here, because we need self.std to be defined for install_files() to work properly and also cannot be executed before make_runpath_dirs() as it needs the directories being created here.
So another solution would be to rename make_runpath_dirs function to a better describing name, but does it worth the hassle?

kn-ms (approved):

PR #1104, opened 2024-06-20
Display multitest setup and teardown times on webUI

rnemes:
Users would like to see multitest setup and teardown duration on the Nav bar.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


yuxuan-ms (approved):

testplan/web_ui/testing/src/Common/utils.js
```diff
@@ -27,6 +27,21 @@ function calcExecutionTime(entry) {
...
+  if (timerField && Array.isArray(timerField) && !_.isEmpty(timerField)) {
+    elapsed = 0;
+    timerField.forEach((interval) => {
+      elapsed +=
+      timeToTimestamp(interval.end) - timeToTimestamp(interval.start);
+    });
+  }
+  return elapsed < 0 ? null : elapsed;
+}
+
```
Pyifan, line 44: under interactive mode, if we start/stop multitest multiple times, do we get multiple timer entries (or maybe we wipe out the timer when we restart)?

rnemes: during my testing, it always had one element

PR #1110, opened 2024-07-12
Feature/more debug logging to LogMatcher

zhenyu-ms:
to add debug logging in logmatcher public apis

## Solution description
having method `_match` as internal impl, debug logging info attached to `self`, set to `None` to allow info logging

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/common/utils/match.py
+++ b/testplan/common/utils/match.py
@@ -411,7 +411,6 @@ class LogMatcher(logger.Loggable):
         end_time = time.time() + timeout
 
         self._debug_info_s = None
-        self._debug_info_e = None
         regex = self._prepare_regexp(regex)
 
         while True:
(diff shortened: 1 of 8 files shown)


testplan/common/utils/match.py
```diff
@@ -285,19 +298,58 @@ def match(
                     break
 
             self.position = self.log_stream.position
+            if self._debug_info_e is None:
+                self._debug_info_e = (
+                    str(self.position),
+                    time.time(),
+                    _truncate_line(line),
```
Pyifan, line 305: if we are reaching eof, more useful info would be the last line seen, rather than <eof>

Pyifan (approved):

yuxuan-ms (approved):

PR #1127, opened 2024-08-23
Doc/update interactive & assertions doc

zhenyu-ms:

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


yuxuan-ms (dismissed):

doc/en/interactive.rst
```diff
@@ -73,6 +73,12 @@ circumstances.
 * In interactive mode, all modules containing test targets will be loaded at
   the same time, if the modules from which tasks are scheduled have the same
   name, they won't be reloaded properly.
+* `modulefinder.ModuleFinder <https://docs.python.org/3/library/modulefinder.html#modulefinder.ModuleFinder>`_
+  works purely on static analysis, which means it is completely ignorant of any
```
Pyifan, line 77: let's explain in a bit more detail how module finder find the dependency of the module from import statement:
- absolute import if the module path is accessible from the root (pwd)? (e.g import a.b.c, and if a is under root syspath)
- relative import if the module path is reachable from current module (e.g import ..x.y, if we can traverse to x from current module)

run this through co-pilot for refinement

Pyifan (changes requested):

Pyifan (approved):

PR #1133, opened 2024-09-12
Refactor/app stopping

zhenyu-ms:
lift proc poll at app stopping to parent level (make use of stopped check)

## Solution description
have new timeout attributes & overridden wait methods for drivers

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/common/serialization/fields.py
+++ b/testplan/common/serialization/fields.py
@@ -283,7 +283,7 @@ class GenericNested(fields.Field):
         self.schema_context = schema_context
         self.type_field = type_field
         self.many = kwargs.get("many", False)
-        super(GenericNested, self).__init__(default=default, **kwargs)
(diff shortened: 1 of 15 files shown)


testplan/common/serialization/fields.py
```diff
@@ -283,7 +283,7 @@ def __init__(
         self.schema_context = schema_context
         self.type_field = type_field
         self.many = kwargs.get("many", False)
-        super(GenericNested, self).__init__(default=default, **kwargs)
+        super(GenericNested, self).__init__(default=default, metadata=kwargs)
```
zhenyu-ms, line 286: this change is to address `RemovedInMarshmallow4Warning: Passing field metadata as keyword arguments is deprecated. Use the explicit `metadata=...` argument instead. Additional metadata: `

Pyifan (approved):

PR #1148, opened 2024-11-19
Use lazy import for Matplotlib and move cache to runpath

zsambokiandras:
When one runs Testplan scripts that include matplotlib you get a warning about matplotlib not being able to write to a disted path.

`Matplotlib created a temporary config/cache directory at /tmp/matplotlib-xxxxxxxx because the default path (xxx/.config/matplotlib) is not a writable directory; it is highly recommended to set the MPLCONFIGDIR environment variable to a writable directory, in particular to speed up the import of Matplotlib and to better support multiprocessing.`

## Solution description
Set matplotlib default config/cache path. We can use MPLCONFIGDIR env var to define where matplotlib can write. Use testplan runtime path.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/runnable/base.py
```diff
@@ -1054,12 +1054,21 @@ def make_runpath_dirs(self):
...
         if self.cfg.path_cleanup is False:
             makedirs(self._runpath)
             makedirs(self._scratch)
+            if not env_mpl_cache:
+                makedirs(self._mplcache)
         else:
             makeemptydirs(self._runpath)
             makeemptydirs(self._scratch)
+            if not env_mpl_cache:
+                makeemptydirs(self._mplcache)
```
Pyifan, line 1071: i don't think we shall always create this dir for matplotlib, 99% of time we don't need this dir. Now that the warning msg is only issued when matplotlib is actually used by pdf exporter, it's OK to leave it to user to set it.

Pyifan (changes requested):

testplan/runnable/base.py
zsambokiandras: Removed the creation, now matplotlib automatically creates the dir. Should I also remove setting the env var to the runpath, if not set by the user?

testplan/runnable/base.py
```diff
@@ -1072,6 +1072,11 @@ def make_runpath_dirs(self):
             )
             makedirs(self.resource_monitor_server_file_path)
 
+        if not os.environ.get("MPLCONFIGDIR"):
+            os.environ["MPLCONFIGDIR"] = os.path.join(
+                self._runpath, "matplotlib"
```
Pyifan, line 1077: does matplotlib complain if the dir pointed by MPLCONFIGDIR doesn't exist?

zsambokiandras: No, it tries to create that.

This sets the path to the runpath, if the user didn't set the env var. This way there will be no warning, and the dir is only created when the matplotlib import executes. If I remove this default setting, there will be a warning when matplotlib is imported (and the default location is read-only).

testplan/exporters/testing/pdf/renderers/entries/baseUtils.py
```diff
@@ -81,6 +77,11 @@ def get_matlib_plot(source):
     Call the appropriate plotting function based on
     whether a graph or chart is being plotted.
     """
+    import matplotlib
+
+    matplotlib.use("Agg")
+    import matplotlib.pyplot as plot
```
zhenyu-ms, line 83: maybe this line can be removed?

Pyifan (approved):

Pyifan (approved):

PR #1154, opened 2024-12-04
Add --code flag for collecting code context.

rnemes:
Include the 1 liner of the testcase assertion (top-level from testcase) into the report.

## Solution description
Added ``--code`` flag to collect code context for the assertions. Code context one-liner will be displayed on the web UI if enabled.
Note that file path information is no longer collected by default. To collect file path information, enable code context.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check
- [x] For new cmdline arg: documentation


testplan/web_ui/testing/src/Common/utils.js
```diff
@@ -447,3 +447,17 @@ export const truncateString = (
   
   return `${startPortion}...${endPortion}`;
 };
+
+/**
+ * Extracts the "workspace" part of the path
+ * @param {String} path
+ */
+export const getWorkspacePath = (path) => {
+  const srcIndex = path.indexOf('/src/');
```
rnemes, line 456: This only works in linux

tests/unit/testplan/testing/multitest/test_result.py
```diff
@@ -43,17 +43,29 @@ def intermediary(result, description=None):
...
+
+    # Functions to test marking, with relative position of the assertions
+    functions = [
+        (test_group_marking, 6),
+        (helper, 1),
+        (intermediary, 2),
+    ]
+    current_function = inspect.currentframe().f_code.co_name
+    for function, relative_position in functions:
+        # To prevent infinite recursion
```
Pyifan, line 62: The modification to this testcase doesn't look right to me.

Let's discuss offline.

Pyifan (changes requested):

testplan/testing/multitest/base.py
```diff
@@ -1158,6 +1158,9 @@ def _run_testcase(
                 ),
```
Pyifan, line 1158: when --code is not specified, we don't need to call report_target on testcase either. Every report_target call evokes inspect and is expensive.

testplan/testing/multitest/base.py
```diff
@@ -1158,6 +1158,9 @@ def _run_testcase(
...
             )
 
+        if getattr(self.cfg, "collect_code_context", False):
+            testcase = collect_code_context(func=testcase)
```
Pyifan, line 1162: This doesn't make sense, the decorator sets a threading.local variable, why are do calling it on every testcase?

Pyifan (changes requested):

testplan/testing/multitest/base.py
```diff
@@ -1158,6 +1158,9 @@ def _run_testcase(
                 ),
             )
 
+        if getattr(self.cfg, "collect_code_context", False):
```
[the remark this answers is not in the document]
rnemes: Several tests under "unit/testplan/testing/multitest/test_multitest" fail if I don't use getattr

tests/unit/testplan/testing/multitest/test_result.py
rnemes: Reverted back to the previous style as we discussed

testplan/testing/multitest/base.py
rnemes: modified: report_target is now no-op when --code is not set

testplan/testing/base.py
```diff
@@ -901,6 +901,14 @@ def driver_info(self) -> bool:
...
         return self.cfg.driver_info
 
+    @property
+    def collect_code_context(self) -> bool:
+        """
+        Collecting the file path, line number and code context of the assertions
+        if enabled.
+        """
+        return getattr(self.cfg, "collect_code_context", False)
+
```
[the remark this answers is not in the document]
zhenyu-ms: might need it for hooks & maybe result call under pytest wrapper

testplan/testing/multitest/base.py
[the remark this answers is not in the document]
zhenyu-ms: the getattr call should be inside that `collect_code_context` property in `testplan/testing/base.py`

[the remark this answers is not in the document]
rnemes: > the getattr call should be inside that `collect_code_context` property in `testplan/testing/base.py`

Good catch, makes sense!

testplan/testing/multitest/base.py
rnemes: I have refactored the solution and added using Result for storing the flag as we discussed it online.

testplan/testing/result.py
```diff
@@ -1383,8 +1392,11 @@ def __exit__(self, exc_type, exc_value, traceback):
             self.description,
             self.category,
         )
-        assertion.file_path = os.path.abspath(caller_frame[1])
-        assertion.line_no = caller_frame[2]
+
+        if caller_frame:
+            assertion.file_path = os.path.abspath(caller_frame[1])
+            assertion.line_no = caller_frame[2]
+            # assertion.code_context = caller_frame.code_context[0].strip()
```
Pyifan, line 1399: why is this commented out?

Pyifan (changes requested):

testplan/testing/result.py
rnemes: very good question, not intentionally

testplan/web_ui/testing/src/Common/utils.js
zhenyu-ms: 🤔

rnemes: (Discussed this offline.) This is a request from management, the full path will be visible in the popup.

Pyifan (approved):

PR #1162, opened 2024-12-18
Testplan result importer to best effort fix attachment path

Pyifan:
When we copy report.json and _attachment around, we are not able to locate the attachments after import json.

## Solution description
Best effort fix attachment path by checking local _attachment dir.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [X] MS info leakage check


zhenyu-ms (approved):
LGTM

PR #1174, opened 2025-01-14
Implemented Worker-Specific Task Assignment Capabilities.

tejaschauhan373:

* Implemented [Worker](https://testplan.readthedocs.io/en/24.9.2/_modules/testplan/runners/pools/base.html#Worker)-Specific [Task](https://testplan.readthedocs.io/en/24.9.2/_modules/testplan/runners/pools/tasks/base.html#Task) Assignment Capabilities. 

## Solution description

* Added default parameter `workers_name` to store `uid` of workers in [Task](https://testplan.readthedocs.io/en/24.9.2/_modules/testplan/runners/pools/tasks/base.html#Task).
* Added variable `is_picked_up` to keep track of task assignment in [Task](https://testplan.readthedocs.io/en/24.9.2/_modules/testplan/runners/pools/tasks/base.html#Task).
* Added dictionary `queue_map` in [TaskQueue](https://testplan.readthedocs.io/en/24.9.2/_modules/testplan/runners/pools/base.html#TaskQueue) to keep separate task queue for all workers in case of need, or else it will be stored in `default` queue. 
* If there is non-empty `workers_name` given in [Task](https://testplan.readthedocs.io/en/24.9.2/_modules/testplan/runners/pools/tasks/base.html#Task) object, means task needs to run by given `uid` workers only, else it can be run by anyone.
* Create separate queue for each `uid` of `workers_name` & manage it through `queue_map`.
* While assigning [Task](https://testplan.readthedocs.io/en/24.9.2/_modules/testplan/runners/pools/tasks/base.html#Task) to [Worker](https://testplan.readthedocs.io/en/24.9.2/_modules/testplan/runners/pools/base.html#Worker) and popping from queue, check if task is picked up by another worker already or not through `Task.is_picked_up` variable.
* If `Task.is_picked_up` is True then don't assign anything for that pull request, else assign it & mark `Task.is_picked_up` True.
* While `_handle_taskresults`, `_decommission_worker`, `_discard_task`, `discard_pending_tasks`, mark `Task.is_picked_up` False.

## Checklist:
-  Test


Pyifan:
Generally speaking, doing worker-specific task assignment is against pool design principle. With pool execution, we expect all workers are equivalent, and all tasks isolated.

Pyifan:
I'm open to have traits on both worker & task as assignment constraints (like jenkins selector), if we have a valid use case.

tejaschauhan373:
- Is there any other way to start remote workers on servers in different regions, such as Asia, Europe, and America?
- Some internet sites are blocked in specific regions/countries, so need to fetch those sites from allowed countries only, and so need to deploy workers in different regions, and give them doable tasks for that geo-location only.

PR #1182, opened 2025-02-12
Fix/some rework on remote service & remote resource

zhenyu-ms:

TODO

## Solution description

- turn remote service start into steps, thus delay remote driver construction on remote host
- fix certain remote cmd gen utils
- explicitly disallow remote resource usage if remote hosts being windows
- fix empty report check in certain exporters
- add cleanup for dangling symlink/dir due to workspace imitation on remote host, rearrange some source in remote resource as well
- fix a bug in driver dep graph type check

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/common/remote/remote_resource.py
+++ b/testplan/common/remote/remote_resource.py
@@ -167,10 +167,10 @@ class RemoteResource(Entity):
             "host": self.cfg.remote_host,
             "port": self.cfg.ssh_port,
         }
-        if 0 != self._execute_cmd_remote("uname"):
-            raise NotImplementedError(
-                "RemoteResource not supported on Windows remote hosts."
-            )
+        # if 0 != self._execute_cmd_remote("uname"):
+        #     raise NotImplementedError(
+        #         "RemoteResource not supported on Windows remote hosts."
+        #     )
 
         self._remote_plan_runpath = None
         self._remote_resource_runpath = None
--- a/testplan/common/remote/remote_resource.py
+++ b/testplan/common/remote/remote_resource.py
@@ -70,6 +70,8 @@ class UnboundRemoteResourceConfig(EntityConfig):
             ConfigOption("remote_runpath", default=None): str,
             ConfigOption("testplan_path", default=None): str,
             ConfigOption("remote_workspace", default=None): str,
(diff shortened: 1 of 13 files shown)


testplan/common/remote/remote_resource.py
```diff
@@ -400,31 +423,64 @@ def _prepare_workspace(self, exist_on_remote: bool) -> None:
...
+                        )
+                    )
+                    + '; do e+="/${i}"; if [ ! -e "$e" ]; then echo "$e"; break; fi; done'
+                ),
+                label="imitate local workspace path on remote - detect non-existing",
+                stdout=os.fdopen(w),
+                check=False,  # XXX: bash might not be there
+            ):
+                rmt_non_existing = os.fdopen(r).read().strip() or None
+            if 0 == self._execute_cmd_remote(
```
Pyifan, line 461: If we are only logging self._dangling_remote_fs_obj, let's comment out this part for now too. i don't think we shall add any extra ssh call, or try to do complicated things with shell. I think more strategically we can create python scripts that can be used in remote to take care of such, e.g remote_setup.py and remote_cleanup.py

testplan/common/remote/remote_resource.py
```diff
@@ -400,31 +423,64 @@ def _prepare_workspace(self, exist_on_remote: bool) -> None:
...
                 cmd=link_cmd(
                     path=self._workspace_paths.remote,
                     link=self._workspace_paths.local,
                 ),
                 label="imitate local workspace path on remote - ln",
                 check=False,  # just best effort
-            )
+            ):
+                # NOTE: we shall always remove created symlink
+                self._dangling_remote_fs_obj = (
```
Pyifan, line 474: we can though, just remove the symlink we created, at least to avoid circular symlink problem for user. Take care of more complicated use case later.

Pyifan (changes requested):

testplan/common/remote/remote_resource.py
zhenyu-ms: sure thing, will convert some lines into comments

Pyifan (approved):

PR #1184, opened 2025-02-20
Make testplan timeout obvious for users in report

rnemes:
When testplan level timeout happens, the error message is only visible if we click the 'home' button on UI, which is very well hidden from the users.

## Solution description
- Add a test level synthesize entry named "Testplan Timeout"
- Let's see if test type information is available, and add the test type icon
- Fix the '-> _Passed_' issue in stdout, make it _Incomplete_

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check

### DIFF
--- a/testplan/common/entity/base.py
+++ b/testplan/common/entity/base.py
@@ -27,6 +27,7 @@ import psutil
 from schema import Or
 
 from testplan.common.config import Config, ConfigOption
+from testplan.common.report import Status
 from testplan.common.utils import logger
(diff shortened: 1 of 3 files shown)


tests/unit/testplan/runners/test_local.py
rnemes, line 1: Need better tests here.

rnemes:
Still need to implement collecting the states of tests when we hit timeout.

Pyifan (approved):

PR #1203, opened 2025-05-15
Create dict.match_all and fix.match_all UI components

zsambokiandras:
dict.match_all and fix.match_all frontend components are missing

## Solution description
created missing components

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/doc/newsfragments/3172_new.dictmatchall_fixmatchall.rst
+++ b/doc/newsfragments/3172_new.dictmatchall_fixmatchall.rst
@@ -1,0 +1,1 @@
+Create new frontend components ``DictMatchAll`` and ``FixMatchAll``.


testplan/web_ui/testing/src/AssertionPane/Assertion.js
```diff
@@ -46,6 +48,7 @@ function Assertion({
   uid,
   reportUid,
   toggleExpand,
+  hideType,
```
zsambokiandras, line 51: Writing "DictMatch" for all entries of DictMatchAll is a bit redundant, hence the new prop to hide the assertion type.

testplan/web_ui/testing/src/Report/reportUtils.js
```diff
@@ -274,8 +274,9 @@ const getAssertions = (selectedEntries, displayTime, UTCTime) => {
     // new entry structure
     if (entry.timestamp) {
       let d = new TZDate(entry.timestamp * 1000, IANAtz);
-      let rep = UTCTime ? formatISO(d.withTimeZone("UTC")) : formatISO(d);
-      return rep.split("T")[1];
+      return UTCTime
+        ? format(d.withTimeZone("UTC"), "HH:mm:ss.SSSXXX")
+        : format(d, "HH:mm:ss.SSSxxx");
```
zsambokiandras, line 279: The difference between the ...xxx and ...XXX formats is appending "+00:00" or "Z" as the time zone. This is to keep the current behavior (reports without timestamp).

PR #1217, opened 2025-08-07
Add paramiko ssh_client to RemoteResouce

Pyifan:
Use paramiko ssh_client instead of ssh cmd for remote features

## Solution description

## Checklist:
- [X] Test
- [ ] Documentation (API)
- [X] MS info leakage check


testplan/common/remote/remote_resource.py
```diff
@@ -784,3 +811,15 @@ def _transfer_data(
                 stdout=devnull,
                 logger=self.logger,
             )
+
+    def _check_remote_os(self):
```
zhenyu-ms, line 815: we may store `/etc/os-release` output in `self._remote_os_release` instead, thus no need to make any assumption

testplan/common/remote/ssh_client.py
```diff
@@ -0,0 +1,230 @@
...
+            "ssh_client executing command [%s]: '%s'",
+            label,
+            cmd_string,
+        )
+        start_time = time.time()
+
+        _, stdout, stderr = self._ssh_client.exec_command(
+            command=cmd_string,
+            timeout=timeout,
+            environment=env,
```
zhenyu-ms, line 137: `env` usually won't work if passed in this way, need to prepend them to command line instead

PR #1225, opened 2025-08-19
Print search-friendly anchor in testplan log for test/suite/case start/stop

lialuga:
The benefit is for user to quickly navigate to start/stop of a particular testcase during investigation.

## Solution description
Added logging to testplan.testing.multitest.base

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/testing/multitest/base.py
+++ b/testplan/testing/multitest/base.py
@@ -476,6 +476,10 @@ class MultiTest(testing_base.Test):
         testsuites = self.test_context
         report = self.report
 
+        self.logger.debug(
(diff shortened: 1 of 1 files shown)


Pyifan:
hi Luca, please find/create an example with >1 multitests, each multitest shall have before_start/after_start/before_stop/after_stop hooks defined; and with >1 testsuites, with setup / teardown; and >1 testcases, with pre_testcase / post_testcase.

Run this example and let's take a look at it's testplan.log together. Let's see where most appropriate to insert these logging lines, so that we can clearly define the boundary between entries.

PR #1228, opened 2025-08-29
Render DictMatch ignored keys and different types correctly in the HTML report

zsambokiandras:
- Ignored key-value pairs in the HTML report's DictMatch table are rendered in red (as failing) in some cases (when _value_ and _expected_ have certain differing types).
- Comparing a non-iterable and an _iterable_ type (at the same key) is bugged, this renders a table with empty rows.

## Solution description
Modified JSON comparison format and fixed comparison logic.

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [x] MS info leakage check


testplan/common/utils/convert.py
```diff
@@ -279,8 +279,8 @@ def flatten_dict_comparison(comparison: List[Tuple]) -> List[List]:
             # key = '(group)'
 
         status = lpart[3] if lpart else rpart[3]
-        lval = lpart[4] if lpart else None
-        rval = rpart[4] if rpart else None
+        lval = lpart[4] if lpart else (None, "")
```
zsambokiandras, line 282: This change makes the frontend render missing rows in the table, without modifying the frontend. I can still look for a frontend solution though

Pyifan (approved):

PR #1300, opened 2026-03-27
Feature/add condition param to xfail decorator

zhenyu-ms:
impl of conditional xfail

## Solution description
see diff

## Checklist:
- [x] Test
- [x] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/doc/en/introduction.rst
+++ b/doc/en/introduction.rst
@@ -361,7 +361,7 @@ Command line
                                 "Fatal MultiTest:Environment Start:Starting": {
                                     "reason": "env does not start",
                                     "strict": true,
-                                    "condition": {"error": "While starting driver MyApp[app1]"}
+                                    "condition": {"error": "While starting driver MyApp\\[app1\\]"}
                                 },
                                 "Flaky MultiTest:Suite Name:\*": {
                                     "reason": "everything under that suite flaky",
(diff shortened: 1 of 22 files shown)


doc/en/introduction.rst
```diff
@@ -340,7 +340,36 @@ Command line
...
+                                        "assertions": {
+                                            "type": "Equal",
+                                            "description": "some part of description of that failed assertion"
+                                        }
+                                    }
+                                },
+                                "Fatal MultiTest:\*:\*": {
+                                    "reason": "env does not start",
+                                    "strict": true
+                                },
```
Pyifan, line 365: Can we add an example "when" for multitest env does not start?

doc/en/multitest.rst
```diff
@@ -1437,6 +1434,49 @@ If a test is expect to fail all the time, you can also use the `strict=True` the
     def fail_testcase(self, env, result):
         ...
 
+Each xfail entry can include an optional ``when`` parameter to indicate which
+failure should be expected. This is useful when a test may fail for different
+reasons and you only want to suppress a specific known failure. The ``when``
+parameter must contain exactly **one** of the following keys:
+
+**Matching on log messages** - applies xfail only when a matching log message
+is found in the report entry's logs:
```
Pyifan: report entry's log is for capturing exceptions, right? so from user perspective, testcase is in error status and we would see exception message in the report.

Pyifan (changes requested):

doc/newsfragments/3584_new.xfail_condition.rst
```diff
@@ -0,0 +1 @@
+:ref:`Xfail` now supports a ``condition`` parameter for conditional xfail based on log messages or assertion entries.
```
Pyifan: log -> error or exception

Pyifan (dismissed):

Pyifan (approved):

PR #1316, opened 2026-05-12
Refactor/drivers w dependency defined start & stop in threadpool

zhenyu-ms:
^

## Solution description
see change

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/testing/environment/base.py
+++ b/testplan/testing/environment/base.py
@@ -1,5 +1,6 @@
 import copy
 import time
+from concurrent.futures import Future, ThreadPoolExecutor
 from dataclasses import dataclass
 from typing import Any, Dict, TYPE_CHECKING, Optional
 
@@ -10,8 +11,10 @@ from testplan.testing.environment.graph import DriverDepGraph
 
 if TYPE_CHECKING:
     from testplan.testing.base import Test
+    from testplan.testing.environment.graph import D
 
 MINIMUM_CHECK_INTERVAL = 0.1
+MAX_WORKER_THREADS = 32
 
 
 @dataclass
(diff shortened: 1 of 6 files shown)


testplan/testing/environment/base.py
```diff
@@ -95,9 +107,115 @@ def stop_in_pool(self, *_: Any) -> None:
...
+        timeout occurs. Records exceptions into ``self.start_exceptions``.
+
+        Raises no exception; failure is signalled by recording into the
+        exception store and returning.
+        """
+        try:
+            watch.record_start()
+            driver.start()
+        except Exception:
+            self._record_resource_exception(
```
Pyifan, line 126: is this thread safe?

testplan/testing/environment/base.py
```diff
@@ -95,9 +107,115 @@ def stop_in_pool(self, *_: Any) -> None:
...
+                            "Timeout when starting %s despite"
+                            " it's probably started now. %s",
+                            driver,
+                            TimeoutExceptionInfo(watch.start_time).msg(),
+                        )
+                    driver._after_started()
+                    driver.logger.info("%s started", driver)
+                    return
+                if time.time() >= watch.start_time + watch.total_wait:
+                    raise TimeoutException(
```
Pyifan, line 152: Timing accuracy on timeout: not properly recorded. The driver's setup timer (and lifespan timer) is left open with no end, so the elapsed setup time for a timed-out driver is meaningless. This is a pre-existing bug, not a regression — but given that the PR's stated goal is accurate per-driver timing, it would be worth fixing here (a try: driver.timer.end("setup") in the exception handler is a one-line fix and aligns with the spirit of the refactor).

testplan/testing/environment/base.py
zhenyu-ms: it's just writing to a shared dict, according to python doc this is thread safe (it should use its internal lock)

testplan/testing/environment/base.py
zhenyu-ms: (discussed offline)

Pyifan (approved):

Pyifan (approved):

PR #1318, opened 2026-05-13
Increase precedence of XFAIL status in report

zsambokiandras:
XFAIL has lower precedence than PASS. this means that XFAIL and PASS will aggregate to PASS. This is not desirable, as it hides potential errors.

## Solution description
Moved XFAIL up in the Status enum.

## Checklist:
- [x] Test
- [ ] Documentation (API)
- [x] MS info leakage check

### DIFF
--- a/testplan/web_ui/testing/src/Common/defaults.js
+++ b/testplan/web_ui/testing/src/Common/defaults.js
@@ -138,7 +138,6 @@ const STATUS_CATEGORY = {
   skipped: "unstable",
   xfail: "unstable",
   xpass: "unstable",
-  "xpass-strict": "failed",
   unstable: "unstable",
   unknown: "unknown",
 };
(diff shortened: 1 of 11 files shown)


testplan/web_ui/testing/src/Common/defaults.js
dteo00, line 136: theres already a "xpass-strict": "failed" here

testplan/web_ui/testing/src/Common/defaults.js
```diff
@@ -138,6 +138,7 @@ const STATUS_CATEGORY = {
   skipped: "unstable",
   xfail: "unstable",
```
zhenyu-ms: do we need a new category in the front-end side?

testplan/web_ui/testing/src/Common/defaults.js
zsambokiandras: oops, wrong rebase...

testplan/web_ui/testing/src/Common/defaults.js
zsambokiandras: xfail is orange, so it's not completely wrong. but I'm thinking about giving different icons to xfail, xpass and skipped, and for that, we need to create more categories

zsambokiandras: nvm, I added the new categories and and icons in this PR

Pyifan (approved):

PR #1356, opened 2026-09-18
TESTPLAN-3960: Improve worker readiness and pooled teardown

Pyifan:
Expose a non-blocking process-worker readiness check while retaining the blocking startup lifecycle. Complete pooled asynchronous teardown without changing resource configuration, transfer remote sys.path over the existing SSH connection, and allow 60 seconds for child-worker responses.

Include regression coverage and release notes; omit investigation-only payload diagnostics. Validated with 176 entity, pool and downstream Treadmill tests, plus Ruff formatting and lint checks.

User prompt: "can you clean up oss under ets/testplan and make ready for creating pr?"

Amp-Thread-ID: https://ampcode.com/threads/T-01a00e39-12e6-76be-abfb-9f5efafa6a99

## Solution description

## Checklist:
- [ ] Test
- [ ] Documentation (API)
- [ ] MS info leakage check


testplan/runners/pools/process.py
```diff
@@ -120,24 +120,29 @@ def starting(self) -> None:
...
             interval=(0.04, 0.5),
             timeout=timeout,  # type: ignore[arg-type]
             raise_timeout_with_msg=f"Worker start timeout, logfile = {self.outfile}",
         )
         while next(sleeper):
-            if match_regexps_in_file(
-                self.outfile, [re.compile("Starting child process worker on")]
-            )[0]:
+            if self.started_check():
                 super(ProcessWorker, self)._wait_started(timeout=timeout)
```
Pyifan, line 131: is it still right to call super._wait_started here? does it block?

yuxuan-ms (approved):

testplan/runners/pools/process.py
zhenyu-ms: yes, in practice it shouldn't really block, more like a lifecycle hook here

zhenyu-ms (approved):

